PR #39, opened 2025-02-27
adds JSDoc to public symbols.

elithrar:
Used Claude 3.7 with the following prompt:

  Generate valid JSDoc comments for the public types in the `agents` package.
  Comments should be generated for all public types, including classes,
  interfaces, and functions. Comments should include a description of the type,
  its parameters, and its return value. Be concise, and do not include any
  unnecessary details.

  Return the entirety of each file separately, and include all other code, even
  if there are no JSDOc comments for those symbols.

### DIFF
--- a/.changeset/healthy-bananas-roll.md
+++ b/.changeset/healthy-bananas-roll.md
@@ -1,0 +1,5 @@
+---
+"agents-sdk": patch
+---
+
+adds JSDoc to public symbols.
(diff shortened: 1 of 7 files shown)


threepointone (approved):
this is dope, thank you!

craigsdennis:
This is the good part of the timeline

PR #43, opened 2025-03-01
Fix agents prefix in example

ozanmakes:
From what I could tell from grepping the codebase and what worked for me, the leading slash is not supposed to be there.


threepointone:
Good catch thank you!

PR #53, opened 2025-03-06
fix onMessage not getting called

deathbyknowledge:
The commit updating the `cf_agent_state` logic added a bug where the user's `onMessage` would never get called for string messages.

https://github.com/cloudflare/agents/commit/b244068c7266f048493b3796393cfa74bbbd9ec1#diff-93607242176a99b6eb20508d9bb1707085a97ec316e8835d194d97fa6b44a04cL35-R40


threepointone:
good catch thank you!

Issue #60, opened 2025-03-08
useAgentChat caches request, leading to sync errors


sam-goodwin:
Clearing messages and then unmounting/remounting the agent leads to sync errors because the `/get-messages` call is cached globally.

Is this necessary?

https://github.com/cloudflare/agents/blob/1a4fd6cd39cbdad608fc8f6cf37df71ea02b5ca6/packages/agents/src/ai-react.tsx#L18-L41

sam-goodwin:
Tried fixing with `useState` and `useEffect`. Probably not ideal, but did fix my sync problem: https://github.com/cloudflare/agents/pull/61

PR #61, opened 2025-03-08
fix: avoid global cache of /get-messages request

sam-goodwin:
I am not sure if the team had a better idea here, but this is how I ended up fixing the problem on my end.

Closes: https://github.com/cloudflare/agents/issues/60


threepointone:
I'm going to rewrite this anyway this weekend 
might end up using a websocket message for it

threepointone:
fixed in https://github.com/cloudflare/agents/pull/79

PR #67, opened 2025-03-09
fix: prevent data race with useAgentHook on slow websocket connection

zebp:
In the event the WebSocket connection to the agent is slow the `useAgentChat` hook will try to send `cf_agent_chat_init` before the connection has opened, dropping the message and breaking streaming since `isChatConnection` is never set.

### DIFF
--- a/packages/agents/src/ai-react.tsx
+++ b/packages/agents/src/ai-react.tsx
@@ -130,6 +130,10 @@ export function useAgentChat(options: UseAgentChatOptions) {
   });
 
   useEffect(() => {
+    if (agent.readyState !== agent.OPEN) {
+      return;
+    }
+
     agent.send(
       JSON.stringify({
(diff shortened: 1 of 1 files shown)


threepointone:
can you try with the latest version? I might have fixed this lately 


(partysocket is meant to buffer messages until the connection opens up.)

zebp:
TIL! I updated to latest agent-sdk and still was experiencing broken behavior where the queue in partysocket's `_handleOpen` event handler is empty so the message is never flushed. Going to close this issue and investigate further, I'll re-open it in the appropriate repo once I get a better understanding on whats happening

threepointone:
I removed the need for cf_chat_agent_init here https://github.com/cloudflare/agents/commit/39197ab65a08784b4d5851d5844cb5287c43040e

PR #74, opened 2025-03-11
Replace discriminatedUnion with simple object for Gemini models

gingerhendrix:
It's not perfect some more prompt engineering is required.
  
I can't run the full eval on gemini-2.0-pro-exp because of rate limiting.  It scores 95% on gemini-1.5-pro, and 90% on gemini-2.0-flash


threepointone:
This is actually pretty great, so I'm going to land it. Thank you very much!

PR #100, opened 2025-03-19
Pass state generic through `useAgentChat`

danieljvdm:
In current version I get an error 

`
Types of parameters state and state are incompatible.
`

When trying to type the `useAgent` hook with my state object. 

E.g.

```ts
// No errors here
const agent = useAgent<{ currentRecipe: Recipe | null }>({
  host: 'http://localhost:8787',
  agent: 'chat',
  onStateUpdate: (state) => {
    setCurrentRecipe(state.currentRecipe);
  },
});

const {
  messages: agentMessages,
  input: agentInput,
  handleInputChange: handleAgentInputChange,
  handleSubmit: handleAgentSubmit,
  addToolResult,
  clearHistory,
} = useAgentChat({
  // Error here, my `currentRecipe` shape does not match `unknown` since the generic isn't being passed through
  agent,
  maxSteps: 5,
});
```

### DIFF
--- a/.changeset/thin-foxes-hang.md
+++ b/.changeset/thin-foxes-hang.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
+Pass state generic through `useAgentChat`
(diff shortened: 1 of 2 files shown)


threepointone:
great catch, thank you!

threepointone:
this will be released in 0.0.40 once CI does its thing  https://github.com/cloudflare/agents/actions/runs/13951993889/job/39053669107

PR #105, opened 2025-03-20
WIP: Implement chat agent stop generation

### LINKED ISSUE #68: `stop` works client side, but backend continue to stream
`useAgentChat` proxies the underlying `useChat`'s `stop` function, but does not reflect on the server.

I know this, because if i refresh after stopping the stream, the message appears.

```ts
const {
  stop,
  // ..
} = useAgentChat({ .. });

const handleCancel = () => {
  stop();
}
```

To workaround, I had to create an `AbortController` and pass the `AbortSignal` through to `streamText`:

```ts
const result = streamText({
  // ..
  abortSignal: controller.signal,
});
```

Then, I call my own API on the `Agent` to trigger the Abort Signal. 

The final barrier I haven't yet gotten over is how to make sure the final message parts are persisted. 

Not sure if I am approaching this in the right way.

brettimus:
Fixes #68 (kind of)

***

Will write a bit more of a stable description when I get better internet 

However, broad strokes:

- the ChatAgent keeps a map of request ids to AbortControllers
- the ChatAgent passes an abortSignal that can be used by implementers of `onChatMessage`
- when you call `stop` on the client, it emits a new type of message (`"cf_agent_chat_request_cancel"`)
- the ChatAgent handles this new message by looking up if it has an assocaited abortSignal that it can invoke

### DIFF
--- a/packages/agents/src/ai-chat-agent.ts
+++ b/packages/agents/src/ai-chat-agent.ts
@@ -82,10 +82,10 @@ export class AIChatAgent<Env = unknown, State = unknown> extends Agent<
           [connection.id]
         );
         await this.#persistMessages(messages, [connection.id]);
-        console.log("HEYYY MESAGES: ", JSON.stringify(messages, null, 2))
-        const { id: requestId, abortSignal } = this.#getLatestMessageAbortSignal(messages) ?? {};
-        console.debug("id:", requestId)
-        console.debug("abortSignal:", abortSignal);
+
+        const chatMessageId = data.id;
+        const abortSignal = this.#getAbortSignalForMessage(chatMessageId);
+
         return this.#tryCatch(async () => {
           const response = await this.onChatMessage(async ({ response }) => {
             const finalMessages = appendResponseMessages({
@@ -94,10 +94,9 @@ export class AIChatAgent<Env = unknown, State = unknown> extends Agent<
             });
 
             await this.#persistMessages(finalMessages, [connection.id]);
-            if (requestId) {
-              this.#removeAbortController(requestId)
-            }
-          }, abortSignal ? { signal: abortSignal } : undefined);
+            this.#removeAbortController(chatMessageId)
+          }, abortSignal ? { abortSignal } : undefined);
+
           if (response) {
             await this.#reply(data.id, response);
           }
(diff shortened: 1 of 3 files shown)


packages/agents/src/ai-chat-agent.ts
```diff
@@ -74,21 +82,31 @@ export class AIChatAgent<Env = unknown, State = unknown> extends Agent<
...
+            abortSignal ? { abortSignal } : undefined
+          );
+
           if (response) {
             await this.#reply(data.id, response);
           }
         });
       }
       if (data.type === "cf_agent_chat_clear") {
+        this.#destroyAbortControllers();
```
brettimus, line 109: Note: Here, when you clear the chat, you'd also end up cancelling pending generations. Not sure if that's expected behavior

threepointone:
sorry I haven't reviewd this yet, on my dock for next week

brettimus:
> sorry I haven't reviewd this yet, on my dock for next week

Niiiice thanks. Appreciate the comment 

figured things were hectic with lead up to dev week

threepointone:
I'm going to try and push a merge commit now, but don't freak out if it's broken, I'll try to patch it

threepointone (approved):
Ok it looks good! I'm going ot land this, and we can iterate on feedbakc as it comes. Thanks so much, and apologies for the delay!

moesmufti:
hello, is it true that the stop does not propogate or change the isLoading to false? it stays on true at this point still

packages/agents/src/ai-react.tsx
```diff
@@ -135,6 +135,24 @@ export function useAgentChat<State = unknown>(
...
+
+      // NOTE - If we wanted to, we could preserve the "interrupted" message here, with the code below
+      //        However, I think it might be the responsibility of the library user to implement that behavior manually?
+      //        Reasoning: This code could be subject to collisions, as it "force saves" the messages we have locally
+      //
+      // agent.send(JSON.stringify({
+      //   type: "cf_agent_chat_messages",
+      //   messages: ... /* some way of getting current messages ref? */
+      // }))
+
```
Gyurmatag, line 155: As @moesmufti said in the comment, it is not propogating the isLoading state to change. Currently I am working on a feature for the [Agents starter](https://github.com/cloudflare/agents-starter/issues/69) and I found out that the isLoading is still on when we stop the request. I am not sue 100% about the solution but I think here we need also the 
```controller.close();```
to be called. What do you think @brettimus ?

brettimus: Hey! On vacation but have time to look today. May the Fourth be with you

brettimus: yeah good catch, i was able to repro locally and then your fix resolved it

PR #106, opened 2025-03-21
First pass at adding an `McpAgent`

geelen:
Slight conflict with how Agents run with how MCP servers work, which I think we can unify eventually.

But for this initial release, I've made McpAgent extend DurableObject directly, then include an internal #agent instance to provide:

- state
- initialState
- setState
- onStateUpdate
- sql

An McpAgent provides a static mount(path), which does its own session-based routing. This again could be unified with routeAgentRequest, but isn't yet.

Example usage:

```ts
import { McpAgent } from 'agents/mcp'

export class MyMCP extends McpAgent<Env, { counter: number }, {}> {
  server = new McpServer({
    name: 'Demo',
    version: '1.0.0',
  })
  initialState = {
    counter: 1,
  }

  async init() {
    this.server.tool('getCounter', 'Get the current value of the counter', {}, async ({}) => {
      return {
        content: [{ type: 'text', text: String(this.state.counter) }],
      }
    })
    this.server.tool('add', 'Add to the counter, stored in the MCP', { a: z.number() }, async ({ a }) => {
      this.setState({ ...this.state, counter: this.state.counter + a })
      return {
        content: [{ type: 'text', text: String(`Added ${a}, total is now ${this.state.counter}`) }],
      }
    })
  }
}

export default MyMCP.mount('/sse')
```

Or, with auth:

```
export default new OAuthProvider({
  apiRoute: '/sse',
  apiHandler: MyMCP.mount('/sse'),
  defaultHandler: app,
  authorizeEndpoint: '/authorize',
  tokenEndpoint: '/token',
  clientRegistrationEndpoint: '/register',
})
```

### DIFF
--- a/.changeset/purple-vans-end.md
+++ b/.changeset/purple-vans-end.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
+Adding the first iteration of McpAgent
--- a/package-lock.json
+++ b/package-lock.json
@@ -2589,6 +2589,27 @@
         "node": ">=6 <7 || >=8"
       }
     },
+    "node_modules/@modelcontextprotocol/sdk": {
+      "version": "1.7.0",
+      "resolved": "https://registry.npmjs.org/@modelcontextprotocol/sdk/-/sdk-1.7.0.tgz",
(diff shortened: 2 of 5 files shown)


irvinebroque (commented):
👍 trying it out now

threepointone:
this looks good, reviewing asap. can you also add a short example under examples/mcp? happy for it to be in a new PR if it's not simple

PR #114, opened 2025-03-24
Added example McpAgent

geelen:


threepointone:
I'm going to land this so we can iterate on it

Issue #71, opened 2025-03-11
First Class MCP Agents and Tools


Disturbing:
## Context
Agent SDK is leveraging AI SDK’s [Tools](https://sdk.vercel.ai/docs/foundations/tools) to enable LLMs to do custom tasks or operations. The Model Context Protocol (MCPs) is becoming a standard for how LLMs can communicate with each other for the same use cases that tools enable LLMs to discover functionality that it can invoke. We should consider enabling Agents created with the Agent SDK to both:

- Expose itself as an MCP so that others can utilize the agent through this protocol.
- Define a standard for bootstrapping MCPs as Tools for the agent to connect to and leverage in its operations.

**Definition of a tool in MCP**
```
{
  name: string;          // Unique identifier for the tool
  description?: string;  // Human-readable description
  inputSchema: {         // JSON Schema for the tool's parameters
    type: "object",
    properties: { ... }  // Tool-specific parameters
  }
}
```
**Definition of an Tools using AI SDK**
```
tools: {
 name: tool({
      description: 'Get the weather in a location',
      parameters: // Zod or JSON Schema of tool's parameters 
      execute: // Args from params above + History of messages
  }),
  ...
}
```

**Standard Server Sent Events MCP Example**
We already have a solution from cloudflare with [Workers MCP](https://github.com/cloudflare/workers-mcp) which allows servers to serve API endpoints as standard MCP Servers via [JSDdoc standard](https://jsdoc.app/) which makes it intuitive for users to not have to handle bootstrapping MCP Servers.

```
export class ExampleWorkerMCP extends WorkerEntrypoint<Env> {
  /**
   * Generates a random number. This is extra random because it had to travel all the way to
   * your nearest Cloudflare PoP to be calculated which... something something lava lamps?
   *
   * @return {string} A message containing a super duper random number
   * */
  async getRandomNumber() {
    return `Your random number is ${Math.random()}`
  }
  // ...etc
}
```

## Problem
Agent SDK needs a way to bootstrap itself as an MCP Server and easily depend on existing MCP Servers as Tools to perform its desired functionality. This will enable Cloudflare Agents to participate in the growing MCP standard, decouple reusable Agent Tools that can be reused across multiple agents, help support multi-tenancy and the separation/security of state.

[...]

PR #116, opened 2025-03-25
MCP-enabled Agents using HTTP transport

felixnorden:
Tying into the contribution chain from what @geelen added with #114 and adding what we've been tinkering with in the last couple of weeks to make Agents first-class citizens of MCP. 

This follows from the discussion over in #71, opened by @Disturbing, who found that 114 recently got merged while we have been talking about what can be done to allow more MCP adoption in CF.

The base implementation hooks into the main Agent lifecycle to intercept any request that falls under `/mcp` and feed it to the defined MCP server held in DO state. Any updates to the MCP server config are intended to update on a DO being reconstructed (unless the SQLite storage prevents this — if so, I'll have to update the lifecycle logic).

The HTTP transport mainly stems from our own needs and that I don't like the current leaky abstraction of the SSE transport from the MCP SDK. Instead, here we define all lifecycle-dependent calls that aren't in the `Transport` interface to be handled internally by providing callbacks in the `constructor` (although the event-loop timeout to await the Response feels like a hack).

Any and all feedback is appreciated!

### Example implementation
```ts
export class DemoMCPAgent extends MCPAgent<Env> {
  async onRequest(request: Request): Promise<Response> {
    if (request.method !== "GET") {
      return new Response("Unsupported method", { status: 400 });
    }
    return new Response(JSON.stringify({ data: await this.random() }));
  }

  createServerParams(): [Implementation, ServerOptions] {
    return [
      { name: "Demo", version: "1.0.0" },
      {
        capabilities: {
          tools: { listChanged: true, random: true },
        },
      },
    ];
  }

  configureServer(server: Server): void {
    server.setRequestHandler(ListToolsRequestSchema, () => {
      return {
        tools: [{ name: "example", description: "An example tool" }],
      };
    });
    server.setRequestHandler(CallToolRequestSchema, async ({ params }) => {
      if (params.name !== "example") throw new Error("Unknown tool");

      return {
        content: [{ type: "text", text: "Hello, world!" }],
      };
    });
  }
}

export default class Worker extends WorkerEntrypoint<Env> {
  /**
   * @ignore
   **/
  async fetch(request: Request): Promise<Response> {
    const agent = await getAgentByName(this.env.RANDOM, "random");
    return await agent.fetch(request);
  }
}
```


irvinebroque:
Seems like the transport piece gets simpler with https://github.com/modelcontextprotocol/specification/pull/206?

felixnorden:
> Seems like the transport piece gets simpler with [modelcontextprotocol/specification#206](https://github.com/modelcontextprotocol/specification/pull/206)?

Correct, and the way that this is currently implemented is to allow us to swap out the current `HTTPTransport` to instead use the new `Streamable HTTP Transport` once that is released or this could be tailored to facilitate it (hence why I was so adamant to not add any non-interface methods to the transport)

threepointone:
Love that you did this, we'll have a look this week

threepointone:
waiting on @geelen, we'll probably look at this on monday!

felixnorden:
Roger that, thanks for keeping me in the loop @threepointone 🫡

felixnorden:
I'm circling back here just so it doesn't get forgotten—anything you'd want/need me to look at further. One thing I've been looking at is starting to implement the new requirements for SSE promotions (and baking in WS options, too, since they're better in CF) for the Transport.

felixnorden:
Saw that the `StreamableHTTPTransport` just got added in the repo today, so a new release including it is likely very close!

I'll close this PR for now and update examples etc with that transport instead when it comes.

PR #118, opened 2025-03-26
fix: Pass Env param thru to DurableObject definition

max-stytch:
Fixes an issue where classes extending `McpAgent` cannot access typesafe `this.env`, since the DurableObject type is declared as `DurableObject<Env = unknown>`


threepointone:
Thank you!

PR #125, opened 2025-03-27
[Draft] MCP Client x Agents Implementation

cmsparks:
This implements a basic MCP Client in the Agents DO.

This client setup supports:
- Capability discovery for tools, prompts, resources
- ListChangedNotification handlers for tools, prompts, and resources
- Multiple client connections and tool/prompt/resource namespacing
- SSEEdgeClientTransport (to support running in workers)

Coming in a later PR:
- More ergonomic support for: `tools/call`, `resources/read`, `resources/templates/list`, `resources/subscribe`, `prompts/get`
- OAuth (currently just supports header/bearer auth)
- Support for the new MCP transport spec
- Client capabilities (`roots`, `sampling`)

I included a straightforward example which you can test out with whatever MCP servers.


packages/agents/src/index.ts
```diff
@@ -759,6 +773,64 @@ export class Agent<Env, State = unknown> extends Server<Env> {
...
+  MCPConnections: Record<string, MCPClientConnection> = {};
+
+  /**
+   * Connect to and register an MCP server
+   *
+   * @param transportConfig Transport config
+   * @param clientConfig Client config
+   * @param capabilities Client capabilities (i.e. if the client supports roots/sampling)
+   */
+  async connectToMCPServer(
```
cmsparks: Important part is the API here

packages/agents/src/lib/mcpClient.ts
```diff
@@ -0,0 +1,184 @@
...
+  async registerTools(): Promise<Tool[]> {
+    if (!this.serverCapabilities || !this.serverCapabilities.tools) {
+      return [];
+    }
+
+    if (this.serverCapabilities.tools.listChanged) {
+      this.client.setNotificationHandler(
+        ToolListChangedNotificationSchema,
+        async (_notification) => {
+          this.tools = await this.fetchTools();
```
cmsparks: Maybe we allow people to hook into these notification handlers in addition to us handling the updates because I _think_ the MCP SDK only supports adding one handler per notification type. But that's probably out of scope of this PR

examples/mcp-client/src/server.ts
```diff
@@ -0,0 +1,85 @@
...
+  constructor(
+    public ctx: DurableObjectStorage,
+    public env: Env
+  ) {
+    super(ctx, env);
+  }
+
+  async onStart(): Promise<void> {
+    await Promise.all(
+      this.env.MCP_SERVERS.map((server) => {
```
irvinebroque: Interesting that in this example the agent goes and opens connections to MCP servers when it starts, rather than on-demand. I see what you mean about need for tools caching eventually. Because at inference time need to know what tools are even available.

examples/mcp-client/src/server.ts
```diff
@@ -0,0 +1,85 @@
+import {
+  Agent,
+  routeAgentRequest,
+  type AgentNamespace,
+  type Connection,
+} from "agents";
+
+type Env = {
+  MCP_SERVERS: { host: string; info: { name: string; version: string } }[];
```
irvinebroque: listChanged as vNext cc @dinasaur404 since we have been talking to people asking if we'd support

irvinebroque: https://spec.modelcontextprotocol.io/specification/2024-11-05/basic/lifecycle/

packages/agents/src/index.ts
```diff
@@ -759,6 +773,64 @@ export class Agent<Env, State = unknown> extends Server<Env> {
...
+  /**
+   * Utility functions to get namespaced data for all MCP connections
+   */
+  listTools(): Tool[] {
+    return getNamespacedData(this.MCPConnections, "tools");
+  }
+
+  listPrompts(): Prompt[] {
+    return getNamespacedData(this.MCPConnections, "prompts");
+  }
```
irvinebroque: Should we include this yet? If we don't yet totally support it? Or do we and I don't see it?

packages/agents/src/index.ts
```diff
@@ -759,6 +773,64 @@ export class Agent<Env, State = unknown> extends Server<Env> {
...
+    this.MCPConnections[info.name] = new MCPClientConnection(url, info, opts);
+    await this.MCPConnections[info.name].init();
+  }
+
+  /**
+   * Utility functions to get namespaced data for all MCP connections
+   */
+  listTools(): Tool[] {
+    return getNamespacedData(this.MCPConnections, "tools");
+  }
```
irvinebroque: Effectively does this mean listTools === MCP servers tools? Or can there be other types?

packages/agents/src/lib/mcpClient.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  /**
+   * Notification handler registration
+   */
+  async registerTools(): Promise<Tool[]> {
+    if (!this.serverCapabilities || !this.serverCapabilities.tools) {
+      return [];
+    }
+
+    if (this.serverCapabilities.tools.listChanged) {
```
irvinebroque: Oh I see, you have support for list changed - in that case does the client need to tell the server about this in initialization?

packages/agents/src/lib/sseEdge.ts
```diff
@@ -139,3 +144,32 @@ export class SSEEdgeTransport implements Transport {
...
+    private url: URL,
+    options: SSEClientTransportOptions
+  ) {
+    const fetchOverride = (url: any, options = {}) => {
+      const workerOptions = {
+        ...options,
+      };
+      // Remove unsupported properties
+      // @ts-ignore
+      delete workerOptions.mode;
```
irvinebroque, line 164: We have to do this because of lack of support for request.mode property in runtime?

https://developer.mozilla.org/en-US/docs/Web/API/Request/mode

irvinebroque (commented):
This is awesome. 👏👏👏

Example makes me think a lot about when in the lifecycle of the agent it should connect to MCP servers, and what this will be like to work with when the MCP server that gets used is defined by the end user.

Excited for this, finding some external people to share it with for feedback

examples/mcp-client/src/server.ts
cmsparks: Ideally if the MCP server is well behaved and supports `listChanged` we never need to do caching and can assume our tools/resources/prompts list are always up to date. But for cases where `listChanged` isn't supported, we'd need to do some caching

packages/agents/src/index.ts
cmsparks: Yes, when you call `agent.listTools()` it will enumerate tools for every MCP server connected with name=`${serverName}.${toolName}`

packages/agents/src/index.ts
cmsparks: Yeah, no reason not to IMO. I'm going to go ahead and add the other prompt/resource functionality tomorrow, so it should be mostly supported if anyone wants to use it.

packages/agents/src/lib/sseEdge.ts
cmsparks: Yup, Dhravya figured out this workaround when we were working on connecting to the CF MCP server

packages/agents/src/lib/mcpClient.ts
cmsparks: Yeah, mentioned a bit more here: https://github.com/cloudflare/agents/pull/125#discussion_r2015369764

irvinebroque:
@threepointone @elithrar @rita3ko FYI

elithrar:
@threepointone to review before merging.

examples/mcp-client/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@cloudflare/agents-mcp-client-example",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "start": "concurrently \"vite dev\" \"wrangler dev\" --kill-others"
```
threepointone: why not use vite for the whole thing?

examples/mcp-client/wrangler.jsonc
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "mcp-client",
+  "main": "src/server.ts",
+  "compatibility_date": "2025-03-14",
+  "compatibility_flags": ["nodejs_compat"],
+  "vars": {
```
threepointone: probably not necessary to have this in wrangler.jsonc, just make it a regular var in the source code?

packages/agents/src/index.ts
```diff
@@ -759,6 +780,94 @@ export class Agent<Env, State = unknown> extends Server<Env> {
     // biome-ignore lint/complexity/noBannedTypes: <explanation>
     return callableMetadata.has(this[method as keyof this] as Function);
   }
+
```
threepointone: I'm hesitant to add this to root Agent class, tho I might be able to be convinced. Ping me and let's chat about it?

packages/agents/src/index.ts
```diff
@@ -759,6 +780,94 @@ export class Agent<Env, State = unknown> extends Server<Env> {
...
+    }
+
+    this.MCPConnections[info.name] = new MCPClientConnection(url, info, opts);
+    await this.MCPConnections[info.name].init();
+  }
+
+  /**
+   * Utility functions to get namespaced data for all MCP connections
+   */
+  listTools(): NamespacedData["tools"] {
```
threepointone: probably should be listMCPTools()

packages/agents/src/index.ts
```diff
@@ -759,6 +780,94 @@ export class Agent<Env, State = unknown> extends Server<Env> {
...
+    return getNamespacedData(this.MCPConnections, "prompts");
+  }
+
+  listResources(): NamespacedData["resources"] {
+    return getNamespacedData(this.MCPConnections, "resources");
+  }
+
+  listResourceTemplates(): NamespacedData["resourceTemplates"] {
+    return getNamespacedData(this.MCPConnections, "resourceTemplates");
+  }
```
threepointone: listMCP*

tsconfig.base.json
```diff
@@ -11,7 +11,7 @@
     // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
 
     /* Language and Environment */
-    "target": "ES2021" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
+    "target": "ES2022" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
```
threepointone, line 14: how come? (it's fine, just checking why)

cmsparks: Ran into an `top level await is not available` issue when building on my machine. The source of the issue was some package (don't remember exactly which) calling `await import(node:crypto)`

packages/agents/src/index.ts
cmsparks: Moving the client management/aggregation into a different class, so probably don't need to rename these anymore, right?

packages/agents/src/index.ts
cmsparks: Decided to move this into a separate MCPClientManager class, which handles managing multiple MCP servers at once.

packages/agents/src/index.ts
threepointone: yup, fine now that it's in an inner client

threepointone:
Moved some files around, but this is looking good. I'll do a review with @geelen early next week and land this!

threepointone (approved):
Let's land this and get some feedback on usage.

wesbos:
FYI this demo isn't working because of this issue: https://github.com/cloudflare/workerd/issues/2240



```bash
  VITE v6.2.5  ready in 448 ms

  ➜  Local:   http://localhost:5173/
  ➜  Network: use --host to expose
  ➜  Debug:   http://localhost:5173/__debug
  ➜  press h + enter to show help
node:internal/process/promises:394
    triggerUncaughtException(err, true /* fromPromise */);
    ^

Error: Attempting to read .name on MyAgent before it was set. The name can be set by explicitly calling .setName(name) on the stub, or by using routePartyKitRequest(). This is a known issue and will be fixed soon. Follow https://github.com/cloudflare/workerd/issues/2240 for more updates.
    at async ProxyServer.fetch (file:///Users/wesbos/Sites/reporting-bot/mcp-client/node_modules/miniflare/src/workers/core/proxy.worker.ts:173:11) {
  [cause]: undefined
}

Node.js v23.7.0
```

threepointone:
Dammit, ok. I can have a look tomorrow morning, unless @cmsparks beta to it before me

cmsparks:
Ah, I think that's because the new version of `MCPClientManager` uses the `this.name` in the initialization in the example, so we need to set it in onStart. I think we may have moved it around in the example in #140. In the meantime, the example here works: https://github.com/cloudflare/ai/tree/main/demos/mcp-client 

Let me fix + roll into the client changes I was going to make today anyways

wesbos:
@cmsparks thanks - that one looks to run! I had another issue with the demo when using an MPC server that doesn't implement Resource Templates (Playwright MCP sever) - the entire demo breaks instead of just not showing them. Not sure if this is an error in the library or the demo, but I thought I'd let you know

https://share.cleanshot.com/LDhYw0xL

cmsparks:
Ah I think I know what it is, we assume that resource templates are supported if the server supports resources during capability negotiation (there isn't an explicit resource template capability). I'll add a fix to handle this edge case in `agents`

wesbos:
awesome thanks a ton!

PR #126, opened 2025-03-27
Add ai-types to esm exports

nickfujita:
Would like to use the `IncomingMessage` and `OutgoingMessage` types when building off of AIChatAgent. This just adds the `ai-types` to the accessible list of exported modules.

Upon merging, we would then be able to import these types as:
`import type { OutgoingMessage, IncomingMessage } from 'agents/ai-types';`


threepointone:
can you tell me what your usecase for these types are? The handlers are hidden and have their own hooks, so I'm trying to understand what you'd use them for

nickfujita:
@threepointone thanks for checking out the pr!

When sending a message, there are cases where we may want to add additional metadata into the body in `useAgentChat({ body: { something: 1 }})`. So to accommodate this I extended the class with an `onMessage` override, where I applied the similar parsing logic as in `AIChatAgent`, where I'm using the `IncomingMessage` type. For now I just copied the type over, but would be great if I could just use the actual exported type from the module.

Open to alternative approaches if you have any for this type of scenario.

threepointone:
How about if we actually passed `body` as the second arg into onChatMessage?

nickfujita:
Sure, that would be great, and save us a bit of boilerplate code parsing the incoming message for our specific case. However, since anyone can still extend the `AIChatAgent` class, and override the `onMessage` method, I still think it would make sense to export these types as they would also most likely be using them to parse the incoming message.

xcaptain:
Will this PR be merged?

PR #128, opened 2025-03-27
Hibernate-able transport

jmorrell-cloudflare:
Current problem: When a user adds an McpAgent to a session with an LLM, the DO is always active during the whole session, even if they never call it.

This modifies the MCP Transport to allow the Durable Object to hibernate. In order to facilitate that, we need a much simpler Transport class that is not based around HTTP+SSEs. https://developers.cloudflare.com/durable-objects/best-practices/websockets/#websocket-hibernation-api

When a request comes in on `/sse` the worker initializes a websocket connection to the appropriate DO for the session, and returns an SSE stream back to the browser. All responses flow over this stream.

Incoming messages on `/sse/message` pass the message to the DO, which allows the MCP server to handle it, sending the response back of WS -> SSE -> Client.

Despite the relative complexity, this seems to work pretty well!


packages/agents/src/mcp.ts
```diff
@@ -101,17 +114,164 @@ export abstract class McpAgent<
...
+  async fetch(request: Request): Promise<Response> {
+    // Only handle WebSocket upgrade requests
+    if (request.headers.get("Upgrade") !== "websocket") {
+      return new Response("Expected WebSocket Upgrade request", {
+        status: 400,
+      });
+    }
+
+    const url = new URL(request.url);
+    const sessionId = url.searchParams.get("sessionId");
```
jmorrell-cloudflare: One other issue is that we always have the DO id === the session id. Ideally we would separate these two, so we could resume a previous session if desired.

packages/agents/src/mcp.ts
```diff
@@ -101,17 +114,164 @@ export abstract class McpAgent<
...
+      this.onerror?.(error as Error);
+      throw error;
+    }
+
+    // ISSUE: After hibernation, this.onmessage is undefined and this.server is a
+    // brand new instance of McpServer without our tools since init is not re-run
+    // on wake up.
+    console.log("onmessage", this.onmessage);
+    console.log("this.server", this.server);
+    this.onmessage?.(parsedMessage);
```
jmorrell-cloudflare: This is where things run into trouble. The DO receives a message after resuming, but these properties are no longer what we expected

jmorrell-cloudflare:
After some discussion I'm going to try to move the server definition into Agent's `onStart` and see where that gets us

packages/agents/src/mcp.ts
```diff
@@ -33,10 +37,13 @@ interface CORSOptions {
...
-  Env = unknown,
-  State = unknown,
-  Props extends Record<string, unknown> = Record<string, unknown>,
-> extends DurableObject<Env> {
+    Env = unknown,
+    State = unknown,
+    Props extends Record<string, unknown> = Record<string, unknown>,
+  >
+  extends DurableObject<Env>
+  implements Transport
```
geelen: Feels weird if the Transport methods are visible within the McpAgent class, unless there's some reason why they should be?

Your other option is a `#transport` internal object, where you wire up just the methods that the hibernation API requires...

packages/agents/src/mcp.ts
geelen: We can actually just accept an arbitrary session ID param on the `/sse` endpoint, it was just that the MCP inspector didn't send one initially.

packages/agents/src/mcp.ts
```diff
@@ -140,28 +300,91 @@ export abstract class McpAgent<
         const url = new URL(request.url);
         const namespace = env[binding];
 
+        // Handle SSE connections
         if (request.method === "GET" && basePattern.test(url)) {
-          const object = namespace.get(namespace.newUniqueId());
+          // Create a unique session ID for this connection
+          const sessionId = namespace.newUniqueId().toString();
```
geelen: If a `sessionID` query param comes in here, you could use that with `idFromString()`. We should maybe make the playground/`useMcp` send one tbh

packages/agents/src/mcp.ts
```diff
@@ -170,20 +393,27 @@ export abstract class McpAgent<
               { status: 400 }
             );
           }
+
+          // Get the Durable Object
           const object = namespace.get(namespace.idFromString(sessionId));
```
geelen: You could also accept client-driven sessionIds here and pass them to `idFromName`, by testing whether it's precisely 64 hex digits...

packages/agents/src/mcp.ts
```diff
@@ -170,20 +393,27 @@ export abstract class McpAgent<
...
+            headers.set(key, value);
           });
-          headerObj["Access-Control-Allow-Origin"] = corsOptions?.origin || "*";
+          headers.set(
+            "Access-Control-Allow-Origin",
+            corsOptions?.origin || "*"
+          );
 
-          return new Response(response.body as unknown as BodyInit, {
+          return new Response(await response.text(), {
```
geelen: Probably should just pass `response.body` here, since that'll stream. Not that it'll matter for this, just a habit to get into I guess

packages/agents/src/mcp.ts
```diff
@@ -101,17 +114,164 @@ export abstract class McpAgent<
...
 
-  async onSSE(path: string): Promise<Response> {
-    this.transport = new SSEEdgeTransport(
-      `${path}/message`,
-      this.ctx.id.toString()
-    );
-    await this.server.connect(this.transport);
-    return this.transport.sseResponse;
+  // Allow the worker to fetch a websocket connection to the agent
+  async fetch(request: Request): Promise<Response> {
```
geelen: Prefer not to use `fetch` at all once you've got RPC methods, personally. You can still use a function that takes Request -> Promise<Response>, but there's no chance that it's being called for some other route/purpose

packages/agents/src/mcp.ts
```diff
@@ -140,28 +300,91 @@ export abstract class McpAgent<
...
+          }
+
+          // Accept the WebSocket
+          ws.accept();
+
+          // Handle messages from the Durable Object
+          ws.addEventListener("message", async (event) => {
+            try {
+              // Send the message as an SSE event
+              const messageText = `event: message\ndata: ${event.data}\n\n`;
```
geelen: I'd _maybe_ consider first decoding the event data here, making sure it's JSON and has the format `{ok: true, data: <mcp-response-here> }`, rather than passing it through unchecked? I suppose it's not like a Response so you wouldn't get error messages or status codes here so maybe it's not necessary, but it always feels weird to sit in the middle of things and not inspect them?

packages/agents/src/mcp.ts
```diff
@@ -101,17 +114,164 @@ export abstract class McpAgent<
...
+        event.wasClean || false
+      );
+    });
+
+    server.addEventListener("error", async (event) => {
+      await this.webSocketError(server, new Error("WebSocket error"));
+    });
+
+    // Connect to the MCP server
+    await this.server.connect(this);
```
geelen: Is this something that needs to happen `onStart`, not here? I don't understand how it can persist across hibernation without it... 🤔

geelen (commented):
This is actually way better than I thought it was gonna look like. So we have no need for `SSEEdgeTransport` at all? That's pretty great tbqh.

If you make `init` an alias for `onStart` this should start working, right?

threepointone:
A lot of this boilerplate is already handled by the PartyServer class, can we just use that?

packages/agents/src/mcp.ts
jmorrell-cloudflare: It cannot. This is one of the things that's currently broken by hibernation

packages/agents/src/mcp.ts
jmorrell-cloudflare: Additionally, we also lose `this.props`

jmorrell-cloudflare:
> If you make init an alias for onStart this should start working, right?

That's part of what's required. We also need to persist `props` and make sure that `onStart` gets called everywhere it's needed since we aren't subclassing Agent.

> A lot of this boilerplate is already handled by the PartyServer class, can we just use that?

I think @geelen had some concerns about directly subclassing Agent / PartyServer, but I don't remember exactly what they were.

I've started trying to sketch out what that would look like in `packages/agents/src/mcp-next.ts`. Guidance / suggestions welcome!

packages/agents/src/mcp.ts
jmorrell-cloudflare: I am not sure you can return a websocket connection if you don't name it fetch and my initial attempts didn't seem to work. I could always have been doing something wrong though. Notably PartyServer also does this: https://github.com/threepointone/partyserver/blob/main/packages/partyserver/src/index.ts#L283

packages/agents/src/mcp.ts
jmorrell-cloudflare: Agreed. This is all really messy. I'll try factoring this out.

packages/agents/src/mcp.ts
```diff
@@ -32,17 +36,56 @@ interface CORSOptions {
...
+  }
+}
+
 export abstract class McpAgent<
   Env = unknown,
   State = unknown,
   Props extends Record<string, unknown> = Record<string, unknown>,
 > extends DurableObject<Env> {
+  #status: "zero" | "starting" | "started" = "zero";
+  #transport?: McpTransport;
```
jmorrell-cloudflare: This assumes there is only one connection / McpAgent. That might work for now, but we should likely make this more flexible in the future.

packages/agents/src/mcp.ts
```diff
@@ -82,36 +125,196 @@ export abstract class McpAgent<
...
   abstract server: McpServer;
-  private transport!: SSEEdgeTransport;
   props!: Props;
   initRun = false;
 
   abstract init(): Promise<void>;
 
   async _init(props: Props) {
+    // @ts-ignore
+    this.setState({ props });
```
jmorrell-cloudflare: Persisting props into the state and re-initializing them works, but this is going to interfere with any user setting their own state. Open to suggestions for better ways of doing this.

packages/agents/src/mcp.ts
```diff
@@ -82,36 +125,196 @@ export abstract class McpAgent<
...
-    );
-    await this.server.connect(this.transport);
-    return this.transport.sseResponse;
+  async #initialize(): Promise<void> {
+    await this.ctx.blockConcurrencyWhile(async () => {
+      this.#status = "starting";
+      await this.onStart();
+      this.#status = "started";
+    });
+  }
```
jmorrell-cloudflare: Shamelessly stolen from PartyServer: https://github.com/threepointone/partyserver/blob/main/packages/partyserver/src/index.ts#L397-L403

packages/agents/src/mcp.ts
jmorrell-cloudflare: Updated this to use `this.ctx.storage` instead

jmorrell-cloudflare:
Marking as ready for review. Testing locally this seems to work well!

@threepointone I'd love your eyes here. I suspect that there's stuff we can further simplify, but guidance on where and how would be very welcome 🙏 I think we want to keep from inheriting Agent / PartyServer directly for now just to keep the scope on this manageable. cc @geelen for verification on that

packages/agents/src/mcp.ts
```diff
@@ -82,36 +133,192 @@ export abstract class McpAgent<
   onStateUpdate(state: State | undefined, source: Connection | "server") {
     // override this to handle state updates
   }
+  async onStart() {
+    this.props = (await this.ctx.storage.get("props")) as Props;
+    this.init();
```
geelen: Can you make this optional? I.e. if they don't define it, don't call it?

geelen (approved):
A few comments still left outstanding but otherwise lgtm, keen to get this out there.

threepointone:
ugh, I merged https://github.com/cloudflare/agents/pull/125 that moved some files around, and now I'm struggling to do a rebase or merge to fix the conflicts. Can you help please? Sorry! Happy to pair.

jmorrell-cloudflare:
@threepointone I brought back the transport file and squished all my commits and that made the rebase a lot easier. This should be ready to review now

threepointone:
awesome I'll review asap

packages/agents/src/mcp/index.ts
```diff
@@ -82,36 +133,192 @@ export abstract class McpAgent<
...
+        server,
+        event.code || 1000,
+        event.reason || "",
+        event.wasClean || false
+      );
+    });
+
+    server.addEventListener("error", async (event) => {
+      await this.webSocketError(server, new Error("WebSocket error"));
+    });
```
```suggestion
```

These aren't necessary when hibernation is on (and won't even work), the handlers are called anyway

packages/agents/src/mcp/index.ts
```diff
@@ -82,36 +133,192 @@ export abstract class McpAgent<
...
    */
   abstract server: McpServer;
-  private transport!: SSEEdgeServerTransport;
   props!: Props;
   initRun = false;
 
   abstract init(): Promise<void>;
 
   async _init(props: Props) {
+    this.ctx.storage.put("props", props);
```
```suggestion
    await this.ctx.storage.put("props", props);
```

packages/agents/src/mcp/index.ts
```diff
@@ -82,36 +133,192 @@ export abstract class McpAgent<
...
+    if (!sessionId) {
+      return new Response("Missing sessionId", { status: 400 });
+    }
+
+    // Create a WebSocket pair
+    const webSocketPair = new WebSocketPair();
+    const [client, server] = Object.values(webSocketPair);
+
+    // Accept the WebSocket with hibernation support
+    this.ctx.acceptWebSocket(server);
```
threepointone, line 202: if we only care about the first websocket that's connected, do you want to throw an error if this is the second one?

threepointone (approved):
just a couple of comments, but tentative approval anyway. happy to land whenever you say.

packages/agents/src/mcp/index.ts
```diff
@@ -140,28 +347,117 @@ export abstract class McpAgent<
...
+              if (message.jsonrpc !== "2.0") {
+                throw new Error("Invalid jsonrpc version");
+              }
+
+              // must have either result or error field
+              if (
+                (message.hasOwnProperty("result") ? 1 : 0) +
+                  (message.hasOwnProperty("error") ? 1 : 0) !==
+                1
+              ) {
```
```suggestion
              if (!message.result && !message.error) {
```

threepointone: this will pass lint

threepointone:
I'm going to land and ship this today so we can play with it! as soon as CI does its thing

examples/playground/tsconfig.json
```diff
@@ -1,3 +1,6 @@
 {
-  "extends": "../../tsconfig.base.json"
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "target": "ES2021"
```
jmorrell-cloudflare, line 4: Note that I didn't change the target in the base config, it was already ES2022, so this should be a change in target for the playground.

I changed the lib though I'm not 100% on the implications of that

PR #132, opened 2025-03-30
add description migration and update signatures

deathbyknowledge:
The criteria object in `getSchedules` [supports the field description](https://github.com/cloudflare/agents/blob/main/packages/agents/src/index.ts#L629-L632) but neither the table schema nor the `schedule` method do.

This is just to kick off a discussion on whether we want to remove `description` or add support for it. My changes are a bit rough, as they introduce a breaking change to the `schedule` method.

### DIFF
--- a/packages/agents/src/index.ts
+++ b/packages/agents/src/index.ts
@@ -135,6 +135,8 @@ export type Schedule<T = string> = {
   callback: string;
   /** Data to be passed to the callback */
   payload: T;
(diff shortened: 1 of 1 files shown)


threepointone:
Thanks for starting this! I suspect the actual thing to do is to actually remove description, since it practice it looks like the payload satisfies that role

threepointone:
following up here https://github.com/cloudflare/agents/pull/133

Issue #135, opened 2025-04-01
Default values for clientInfo in MCP client API?


irvinebroque:
refs https://github.com/cloudflare/agents/pull/125

Today you write this:

```ts
    await this.mcp.connectToServer(new URL("http://localhost:5174/sse"), {
      name: "mcp-server-1",
      version: "1.0.0",
    });
```

In accordance with the [spec](https://spec.modelcontextprotocol.io/specification/2025-03-26/basic/lifecycle/), since the MCP client must share its own name and version.

But:

1. It's a bit awkward to have an API where in the call to connect to something else, you set options about the identity/details of the caller
2. The name and version of the client will be the same when connecting to multiple MCP servers, but you have to declare this multiple times
3. There are likely defaults we can set — the name from the name of the Agent class, and protocol version is something that we should be able to know

Seems like these properties could be set on the `MCPClientManager` class? So that you don't have to set them here?

threepointone:
Yeah I like that idea

irvinebroque:
@cmsparks @geelen @dinasaur404 @jmorrell-cloudflare curious what you think (just thinking about shape that this API ultimately takes and if there's path for it to feel super simple)

irvinebroque:
In same vein — curious why the input type is `URL` rather than a string that then gets validated?

Wonder if there's a way to get it down to this:

```ts
await this.mcp.connect("http://localhost:5174/sse");
```

Or this:

```ts
await this.mcp.connect(["http://localhost:5174/sse"]);
```

...thinking towards a world where there's a registry, an agent connects to many MCP servers, etc.

cmsparks:
Yeah, I sorta mirrored the Client API for the URL/client info portion just for the initial API here (that's why we're explicitly passing it in and using a URL rather than a string). I think I like this though. I think there's inherently going to be complexity though, because the user will eventually need to provide per server transport/client/capability options upon connecting.

threepointone:
can we close this now @cmsparks ?

bermudo01:
I think I found a solution to your problem. Check it out here [click](https://javierlopez.eu/captha/)

cmsparks:
Oh missed this, but yeah we can close this!

PR #140, opened 2025-04-03
Remote MCP Client with auth support

cmsparks:
This PR adds:
* Support for authentication for MCP Clients (Via a DO based auth provider)
* Some improvements to the client API per #135 
* A more in depth example of MCP Client, which allows you to add any number of remote MCP servers with or without auth

This should be good for a review. I still think there are significant improvements we can make to the Client API, but this is just to get something working out there. Might try to flesh the example out a little more tomorrow though (i.e. tool calling?).

One concern is that the user has to handle a lot of application level logic here (auth callback, MCP server+agent lifecycle, frontend flow). That makes the API feel a bit less ergonomic than it should be. I think this might be one case where integrating the MCP server lifecycle/auth handling into the Agent class would be beneficial, but I think that's out of scope of this PR.

Screenshot 2025-04-03 at 1 37 30 AM


packages/agents/src/mcp/client.ts
```diff
@@ -9,42 +9,129 @@ import type {
...
+
+    await this.mcpConnections[id].init();
+
+    return {
+      id,
+      authUrl: authProvider?.authUrl,
+    };
+  }
+
+  isCallbackRequest(req: Request): boolean {
```
cmsparks, line 113: I don't love how the auth callback flow fits in here, but I think that might be something to improve in a later PR.

examples/mcp-client/src/client.tsx
```diff
@@ -12,63 +12,51 @@ interface Message {
...
-    if (!inputRef.current || !inputRef.current.value.trim()) return;
-
-    const text = inputRef.current.value;
-    const newMessage: Message = {
-      id: Math.random().toString(36).substring(7),
-      text,
-      timestamp: new Date(),
-      type: "outgoing",
-    };
+  function openPopup(authUrl: string) {
```
cmsparks, line 35: I should probably add a case where popup windows don't work ala `useMcp`

edit: I think this is fine to leave for now in the example

cmsparks:
Gonna be pushing up some changes in a bit. Running into some issues deploying remotely

irvinebroque:
this is coming along nicely 👏

cmsparks:
Pushed up changes which mostly consist of cleaning things up. Summary:
* Removed a ton of lifecycle handling in the example. We just start a new session on reload now. I'm assuming that we'll add some nicer MCP lifecycle handling later, but this is sufficient.
* Using DO storage for the auth provider instead of KV because we assume you're running a client in a durable object. This means you don't need to create a KV binding.
* Fixed some bugs preventing remote preview from working
* Fixed bug with multiple dynamic client registrations (there might be a bug upstream in the `typescript-sdk`, but it also could be how I'm using the SDK)

threepointone:
I will review and land this tomorrow morning

irvinebroque (commented):
Running the examples, and it looks like the `start` command for `examples/mcp` clobbers what's running on `localhost:5173`

```
"start": "concurrently \"npx @modelcontextprotocol/inspector\" \"vite dev\" --kill-others"
```

https://github.com/cloudflare/agents/blob/main/examples/mcp/package.json#L6

Swap this out for `vite dev` and examples work fine

packages/agents/package.json
```diff
@@ -72,5 +72,8 @@
     "nanoid": "^5.1.5",
     "partyserver": "^0.0.66",
     "partysocket": "1.1.3"
+  },
+  "devDependencies": {
+    "@modelcontextprotocol/sdk": "^1.8.0"
```
threepointone: this is already a dependency, so why add it as as a dev dep?

packages/agents/src/mcp/client-connection.ts
```diff
@@ -33,32 +34,55 @@ export class MCPClientConnection {
   serverCapabilities: ServerCapabilities | undefined;
 
   constructor(
-    url: URL,
+    public url: URL,
     private info: ConstructorParameters<typeof Client>[0],
-    opts: {
+    private opts: {
```
threepointone: call it `options`

packages/agents/src/mcp/client.ts
```diff
@@ -9,42 +9,149 @@ import type {
...
+      client?: ConstructorParameters<typeof Client>[1];
+      capabilities?: ClientCapabilities;
+    } = {}
+  ): Promise<{ id: string; authUrl: string | undefined }> {
+    const id = opts.reconnect?.id ?? crypto.randomUUID();
+
+    // if we have global auth for the manager AND there's no authProvider override
+    // then let's setup an auth provider
+    const authProvider: AgentsOAuthProvider | undefined = this.auth
+      ? new DurableObjectOAuthClientProvider(
```
threepointone, line 77: this should probably be explicit instead of default to a DO layer

packages/agents/src/mcp/sse-edge.ts
```diff
@@ -2,20 +2,26 @@ import {
...
    * Creates a new EdgeSSEClientTransport, which overrides fetch to be compatible with the CF workers environment
    */
-  constructor(
-    private url: URL,
-    options: SSEClientTransportOptions
-  ) {
+  constructor(url: URL, options: SSEClientTransportOptions) {
     // biome-ignore lint/suspicious/noExplicitAny: Overriding fetch, type doesn't matter here
-    const fetchOverride = (url: any, options = {}) => {
+    const fetchOverride = async (url: any, fetchOpts: any = {}) => {
```
threepointone: I assume this should have the type signature for fetch

threepointone (commented):
just did a small review on this. I think I want to rewrite small bits of this. I'll do it myself and land it. please write a changeset and update the pr decription in the future when you make changes?

threepointone:
This has to be revisited 

- the OAuthProvider has to be explicitly passed in, it probably can't default to the DO Provider, so that'll be a small breaking change, we should try to get it in asap. I've added a warning for now. 
- We can't use ctx.storage because that won't be available on the free plan (iiuc?), we probably have to use the sqlite storage layer there 

landing this to unblock for tomorrow, but ping me and let's chat about it

Issue #148, opened 2025-04-05
MCP server mount method throws error when mounted to root "/"


irvinebroque:
**Repro:** https://github.com/irvinebroque/repro-root-mcp-issue/blob/main/src/index.ts

```ts
const mcpHandler = MyMCP.mount('/');

export default {
	async fetch(request: Request, env: any, ctx: any) {
	  return mcpHandler.fetch(request, env, ctx);
	}
  };
```

```
[error] Client connect error: Endpoint origin does not match connection origin: http://message
```

You can fix this by:

```ts
const mcpHandler = MyMCP.mount('/*');
```

This `mount` API feels odd to me because it comes across like it just wants `string` as input:

https://github.com/cloudflare/agents/blob/347171386ef79588aafcc88f4fd9f9308fde28ed/packages/agents/src/mcp/index.ts#L314

But really it's a `pathname`:

https://developer.mozilla.org/en-US/docs/Web/API/URLPattern/pathname

Seems like if someone passes in `/` we should just throw an error because that's not going to do what someone actually wants it to? We need to add this to docs here: https://developers.cloudflare.com/agents/model-context-protocol/mcp-agent-api/

PR #149, opened 2025-04-05
Automatically change "/" path to "/*" in MCP server mount() method

irvinebroque:
Fixes https://github.com/cloudflare/agents/issues/148

Could instead throw an error here and ask people to set this themselves. But having hard time seeing problematic case if we just made it work as people might expect, who aren't thinking about the `path` argument as something that gets fed into `URLPattern`?


threepointone:
landing this, lgtm

Issue #157, opened 2025-04-07
agents 0.0.53 — Cannot read properties of null (reading '__pk') when accessing this.state


irvinebroque:
Trying to call `this.state` as shown here:

https://developers.cloudflare.com/agents/model-context-protocol/mcp-agent-api/#state-synchronization-apis

Results in this error: `Cannot read properties of null (reading '__pk')`

**Clone and run this example to reproduce:** https://github.com/irvinebroque/remote-mcp-server-13/blob/main/src/index.ts

The exact same code was previously working in prior release of the Agents SDK. Found via customer report + when updatind dependencies to use the latest Agents SDK version (https://github.com/cloudflare/ai/pull/86)

threepointone:
On a train, I'll have a look right after lunch

irvinebroque:
Def after 0.0.45

threepointone:
this looks like it's related to the hibernation PR https://github.com/cloudflare/agents/pull/128 @jmorrell-cloudflare digging in some more, jeremy if you're around let's pair and debug

threepointone:
fixed! thanks @jmorrell-cloudflare

PR #158, opened 2025-04-07
Fix the state issues

jmorrell-cloudflare:
Addresses https://github.com/cloudflare/agents/issues/157

There were a couple of issues here:

- We were managing the connections explicitly, but then deferring to PartyKit's Server class for a lot of things. This led to the `Cannot read properties of null (reading '__pk')` error. We resolve this by deferring to the Agent class' `fetch` implementation so it can manage the websocket connection for us.

- We were not passing the `initialState` to the `#agent`, and `initialState` is not available during the constructor. We can resolve this by re-creating the `#agent` in `onStart`

- As a side effect of letting PartyKit manage the connections, we get some more messages broadcast back to the original worker that are not JSONRPC messages. I've re-worked the code there so that it filters those out.

### DIFF
--- a/packages/agents/src/mcp/index.ts
+++ b/packages/agents/src/mcp/index.ts
@@ -206,30 +206,23 @@ export abstract class McpAgent<
       return new Response("Missing sessionId", { status: 400 });
     }
 
-    const resp = await this.#agent.fetch(request);
+    // For now, each agent can only have one connection
+    // If we get an upgrade while already connected, we should error
+    if (this.#connected) {
+      return new Response("WebSocket already connected", { status: 400 });
+    }
 
-    // Create a WebSocket pair
-    // const webSocketPair = new WebSocketPair();
-    // const [client, server] = Object.values(webSocketPair);
+    // Defer to the Agent's fetch method to handle the WebSocket connection
+    // PartyServer does a lot to manage the connections under the hood
+    const response = await this.#agent.fetch(request);
 
-    // // For now, each agent can only have one connection
-    // // If we get an upgrade while already connected, we should error
-    // if (this.#connected) {
-    //   return new Response("WebSocket already connected", { status: 400 });
-    // }
-    // this.ctx.acceptWebSocket(server);
-    // this.#connected = true;
+    this.#connected = true;
 
     // Connect to the MCP server
     this.#transport = new McpTransport(() => this.getWebSocket());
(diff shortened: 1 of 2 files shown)


packages/agents/src/mcp/index.ts
```diff
@@ -93,15 +93,12 @@ export abstract class McpAgent<
...
+  #agent: Agent<Env, State>;
 
   protected constructor(ctx: DurableObjectState, env: Env) {
     super(ctx, env);
     const self = this;
 
-    // Since McpAgent's _aren't_ yet real "Agents" (they route differently, they don't support
-    // scheduling etc, let's only expose a couple of the methods
-    // to the outer class for now.
     this.#agent = new (class extends Agent<Env, State> {
```
jmorrell-cloudflare, line 102: I left this initialization here. I don't think it does anything since it immediately gets re-written in `onStart` but leaving it out here makes `this.#agent`'s type `Agent<Env, State> | undefined` which requires a lot of useless checks throughout the code to pass typechecking.

Ultimately this should go away if we refactor to be a direct subclass of `Agent`

packages/agents/src/mcp/index.ts
```diff
@@ -190,26 +199,23 @@ export abstract class McpAgent<
...
     // For now, each agent can only have one connection
     // If we get an upgrade while already connected, we should error
     if (this.#connected) {
       return new Response("WebSocket already connected", { status: 400 });
     }
-    this.ctx.acceptWebSocket(server);
+
+    // Defer to the Agent's fetch method to handle the WebSocket connection
+    // PartyServer does a lot to manage the connections under the hood
+    const response = await this.#agent.fetch(request);
```
jmorrell-cloudflare, line 210: I'm a little surprised this works as well as it does, but 🙌 yay PartyKit!

packages/agents/src/mcp/index.ts
geelen: You can change the declaration above on line 96 to be `#agent: Agent<Env, State>!;`. Basically, "trust us TSC, it'll always be there"

geelen (approved):

packages/agents/src/mcp/index.ts
threepointone: https://github.com/cloudflare/agents/pull/161

PR #163, opened 2025-04-08
Fix: Missing agent dep in useCallback

brishin:
I noticed on my end that the `agent` reference would get out of date when using the `call` method - this seems to fix things.


threepointone (approved):
thanks!

brishin:
Certainly! It's a great library.

PR #173, opened 2025-04-10
Fixes to MCP Client with authentication

cmsparks:
Agents changes:
* Explicitly pass in the OAuthProvider (and require it to be an `AgentsOAuthClientProvider`)
* Fix error where client initialization fails if a server returns "Method not found" for capabilities it claimed to support. This is more common in the case of resource templates because there is no explicit resource template capability. Now we warn on servers not advertising their capabilities correctly instead of hard failing. (See #125)

Demo fixes:
* Fix problem in the demo where unauthenticated servers wouldn't cause the capabilities list to be refreshed
* Fix broken demo due to `this.name` not being set. Now the callback URL doesn't need to be passed into the `MCPClientManager` (See #125)

I double checked that the `DurableObjectOAuthClientProvider` still works on the free tier. The SQLite storage backend still supports using the key value storage API, so free users shouldn't have issues here.

### DIFF
--- a/examples/mcp-client/src/server.ts
+++ b/examples/mcp-client/src/server.ts
@@ -5,6 +5,7 @@ import type {
   Prompt,
   Resource,
 } from "@modelcontextprotocol/sdk/types.js";
+import { DurableObjectOAuthClientProvider } from "agents/mcp/do-oauth-client-provider";
 
 type Env = {
   MyAgent: AgentNamespace<MyAgent>;
@@ -32,10 +33,7 @@ export class MyAgent extends Agent<Env, State> {
     resources: [],
   };
 
-  mcp = new MCPClientManager("my-agent", "1.0.0", {
-    baseCallbackUri: `${this.env.HOST}/agents/my-agent/${this.name}/callback`,
-    storage: this.ctx.storage,
-  });
+  mcp = new MCPClientManager("my-agent", "1.0.0");
 
   setServerState(id: string, state: Server) {
     this.setState({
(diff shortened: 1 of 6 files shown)


examples/mcp-client/src/server.ts
```diff
@@ -58,12 +56,22 @@ export class MyAgent extends Agent<Env, State> {
...
   async addMcpServer(url: string): Promise<string> {
     console.log(`Registering server: ${url}`);
-    const { id, authUrl } = await this.mcp.connect(url);
+    const authProvider = new DurableObjectOAuthClientProvider(
+      this.ctx.storage,
+      this.name,
+      `${this.env.HOST}/agents/my-agent/${this.name}/callback`
+    );
+    const { id, authUrl } = await this.mcp.connect(url, {
+      transport: { authProvider },
```
threepointone, line 65: should we pass this here, or in the constructor? what benefit to either approach?

examples/mcp-client/src/server.ts
```diff
@@ -58,12 +56,22 @@ export class MyAgent extends Agent<Env, State> {
...
+    });
     this.setServerState(id, {
       url,
       authUrl,
       state: this.mcp.mcpConnections[id].connectionState,
     });
+    if (this.mcp.mcpConnections[id].connectionState === "ready") {
+      await this.refreshServerData();
+    }
```
threepointone, line 74: what does this do? should it be running more often?

packages/agents/src/mcp/client-connection.ts
```diff
@@ -202,11 +212,31 @@ export class MCPClientConnection {
...
     } while (templatesResult.nextCursor);
     return templatesAgg;
   }
 }
+
+function capabilityErrorHandler<T>(empty: T, method: string) {
+  return (e: { code: number }) => {
+    // server is badly behaved and returning invalid capabilities
+    if (e.code === -32601) {
+      console.warn(
```
threepointone:
```suggestion
      console.error(
```

threepointone (approved):
this looks good, tentatively approving, just a couple of questions before landing

examples/mcp-client/src/server.ts
cmsparks: I decided to pass it into the `connect` method because a transport is intended to be fully owned by a client connection (i.e. two MCP clients can't use the same transport). Because the `authProvider` is part of the transport options, it would also have to be owned by one client (so each client can have a unique redirect URL, oauth client information, tokens, etc).

I think we'd require some intermediate auth provider creation function if we were to pass it into the constructor.

edit: realized my initial comment was wrong

examples/mcp-client/src/server.ts
cmsparks: It just refreshes the agent's state for the tools, prompts, and resources which get sent to the client. It should be sufficient to just update these upon adding a new server or authenticating to a server.

In the future we should trigger this function if the server sent a notification that the tools/prompts/resources list changed, but I haven't added a nice way to hook into those notifications in the MCPClientManager yet.

PR #180, opened 2025-04-16
Add error handling for default get-messages request and message parsing

nickfujita:
Fixes issue where app crashes upon receiving a 40x response from default request to `/get-messages`.
Came across this while testing authentication with the agent, using the method prescribed in `hono-agents`
```
app.use(
  "*",
  agentsMiddleware({
    options: {
      onBeforeConnect: async (req) => {
        const token = req.headers.get("authorization");
        // validate token
        if (!token) return new Response("Unauthorized", { status: 401 });
      },
    },
  })
);
```

- Adds error handling to default request to `/get-messages`, setting the error message on the hook return object, and calling the `onError` callback option if provided.
- Adds the same error handling as the above in the places marked as TODO, to keep things consistent.

### DIFF
--- a/packages/agents/src/ai-react.tsx
+++ b/packages/agents/src/ai-react.tsx
@@ -1,7 +1,7 @@
 import { useChat } from "@ai-sdk/react";
 import type { Message } from "ai";
 import type { useAgent } from "./react";
-import { useEffect, use } from "react";
+import { useEffect, use, useState } from "react";
 import type { OutgoingMessage } from "./ai-types";
 
 type GetInitialMessagesOptions = {
@@ -37,7 +37,8 @@ const requestCache = new Map<string, Promise<Message[]>>();
 export function useAgentChat<State = unknown>(
   options: UseAgentChatOptions<State>
 ) {
-  const { agent, getInitialMessages, ...rest } = options;
+  const { agent, getInitialMessages, onError, ...rest } = options;
+  const [error, setError] = useState<Error|undefined>()
 
   const agentUrl = new URL(
     `${// @ts-expect-error we're using a protected _url property that includes query params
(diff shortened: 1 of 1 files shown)


threepointone (commented):
thanks for starting the PR! What do you think of my recommendation?

packages/agents/src/ai-react.tsx
```diff
@@ -55,11 +56,18 @@ export function useAgentChat<State = unknown>(
...
-      credentials: options.credentials,
-    });
-    return response.json<Message[]>();
+    try {
+      const response = await fetch(getMessagesUrl.toString(), {
+        headers: options.headers,
+        credentials: options.credentials,
+      });
+      return response.json<Message[]>();
+    } catch (e) {
```
threepointone, line 65: I think it's better to actually throw the error here, and expect the user to add an ErrorBoundary

nickfujita: I think that might throw off many users, as it did me, because it doesn't seem to be the expected behavior for a hook to throw an error upon receiving a 40x response. For example, when using [tanstack-query](https://tanstack.com/query/latest/docs/framework/react/quick-start) the standard behavior isn't to throw an error if the request fails, but rather to communicate the issue in the form of an `error` property on the return object, and leaving `data` as undefined (which maybe we should do here instead of returning an empty array 🤔).
I don't think the underlying `userPartySocket` requires an ErrorBoundary either, but rather communitates errors via the `onerror` property in it's return object.

If there was a way to provide a typescript error when using this hook without a ErrorBoundary, that would probably make it more feasible, but I'm not currently aware of a way to add that.

Maybe adding a very clear note in the documentation that there needs to be an ErrorBoundary and updating the agent template repo might prevent users from running into this, but I just can't help but feel that there would be more people filing issues for this same thing in the future 😅

whoiskatrin:
Closing this as we will re-think this functionality

PR #185, opened 2025-04-17
1/ Refactoring to prepare to support streamable transport

jmorrell-cloudflare:
In preparation for supporting the streaming protocol, I've moved some things around to help the code "slot in" without disturbing the existing functionality.

- Rename `McpAgent.mount` to `McpAgent.serveSSE` with `McpAgent.mount` serving as an alias for backward compatibility.
- The DO will now store which protocol has been initiated. A session must always use the same protocol once it's been initialized. This is enforced by baking the protocol into the DO address: ``namespace.idFromName(`sse:${sessionId}`);`` however this is also enforced within the DO logic as well.
- When initializing the websocket connection between the worker and the DO, it used to pass the URL request through. Now we enforce that any request for an SSE transport has an `/sse` path. This will allow us to initialize the correct transport. This gives us space to accommodate an '/mcp' path for the streamable protocol. This change is entirely internal and not exposed to users.
- Better leverage the underlying `Agent` class to manage websocket connections. With the SSE transport, there is only one websocket connection, but this will become more important with the Streamable transport as there may be many more connections.

### DIFF
--- a/packages/agents/src/mcp/index.ts
+++ b/packages/agents/src/mcp/index.ts
@@ -89,7 +89,7 @@ export abstract class McpAgent<
 > extends DurableObject<Env> {
   #status: "zero" | "starting" | "started" = "zero";
   #transport?: Transport;
-  #protocol: TransportType = "unset";
+  #transportType: TransportType = "unset";
 
   /**
    * Since McpAgent's _aren't_ yet real "Agents", let's only expose a couple of the methods
@@ -158,11 +158,13 @@ export abstract class McpAgent<
     })(this.ctx, this.env);
 
     this.props = (await this.ctx.storage.get("props")) as Props;
-    this.#protocol = (await this.ctx.storage.get("protocol")) as TransportType;
+    this.#transportType = (await this.ctx.storage.get(
+      "transportType"
+    )) as TransportType;
     this.init?.();
 
     // Connect to the MCP server
-    if (this.#protocol === "sse") {
+    if (this.#transportType === "sse") {
       this.#transport = new McpSSETransport(() => this.getWebSocket());
       await this.server.connect(this.#transport);
     }
@@ -179,7 +181,7 @@ export abstract class McpAgent<
 
   async _init(props: Props) {
     await this.ctx.storage.put("props", props);
-    await this.ctx.storage.put("protocol", "unset");
+    await this.ctx.storage.put("transportType", "unset");
     this.props = props;
     if (!this.initRun) {
       this.initRun = true;
@@ -224,8 +226,8 @@ export abstract class McpAgent<
         }
 
         // This connection must use the SSE protocol
-        await this.ctx.storage.put("protocol", "sse");
-        this.#protocol = "sse";
+        await this.ctx.storage.put("transportType", "sse");
+        this.#transportType = "sse";
 
         // Connect to the MCP server
         if (!this.#transport) {
@@ -281,7 +283,7 @@ export abstract class McpAgent<
 
     // Since we address the DO via both the protocol and the session id,
     // this should never happen, but let's enforce it just in case
-    if (this.#protocol !== "sse") {
+    if (this.#transportType !== "sse") {
       return new Error("Internal Server Error: Expected SSE protocol");
     }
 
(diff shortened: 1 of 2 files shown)


packages/agents/src/mcp/index.ts
```diff
@@ -79,18 +80,19 @@ class McpTransport implements Transport {
   }
 }
 
+type Protocol = "sse" | "streamable" | "unset";
```
irvinebroque: naming nit but — aren't these transport types, not protocols? (since in MCP "protocol" is overloaded term)?

packages/agents/src/mcp/index.ts
```diff
@@ -194,28 +211,36 @@ export abstract class McpAgent<
...
-    }
-
-    // Defer to the Agent's fetch method to handle the WebSocket connection
-    // PartyServer does a lot to manage the connections under the hood
-    const response = await this.#agent.fetch(request);
+    const path = url.pathname;
+
+    switch (path) {
+      // This session is going to communicate via the SSE protocol
+      case "/sse": {
```
irvinebroque, line 220: Does this end up meaning that you _must_ mount your API handler specifically on `/sse`?

https://github.com/cloudflare/ai/blob/main/demos/remote-mcp-server/src/index.ts#L25

packages/agents/src/mcp/index.ts
```diff
@@ -194,28 +211,36 @@ export abstract class McpAgent<
...
+          this.#transport = new McpSSETransport(() => this.getWebSocket());
+          await this.server.connect(this.#transport);
+        }
 
-    return response;
+        // Defer to the Agent's fetch method to handle the WebSocket connection
+        return this.#agent.fetch(request);
+      }
+      default:
+        return new Response("Internal Server Error: Expected /sse path", {
```
irvinebroque, line 242: Think using `Internal Server Error` in context of a library potentially confusing — and wonder if error message could include guidance?

```suggestion
        return new Response("Client tried to initialize a SSE connection on an endpoint other than /sse. When connecting to an MCP server using SSE as the transport, you must connect to the /sse endpoint", {
```

...but then in writing this out — made me realize wait, are we now constraining SSE to only work at the `/sse` path, when that wasn't a requirement before, and isn't part of the spec?

packages/agents/src/mcp/index.ts
```diff
@@ -194,28 +211,36 @@ export abstract class McpAgent<
...
+          status: 500,
```
irvinebroque, line 243: Would this be 422?

packages/agents/src/mcp/index.ts
jmorrell-cloudflare: yeah, `Transport` was already used by the class. `TransportType` would likely be a better name 👍

packages/agents/src/mcp/index.ts
jmorrell-cloudflare: Not in this case since this is the request coming from the worker to the DO, not the request from the user to the worker. The user never sees this.

We use the path the DO receives to understand which protocol we'll be serving: https://github.com/jmorrell-cloudflare/agents/blob/jmorrell/01-streaming-transport-support/packages/agents/src/mcp/index.ts#L412-L413

We have to use one `fetch` method to serve both since only `fetch` can return a websocket connection. This seemed like a logical way to implement it.

packages/agents/src/mcp/index.ts
jmorrell-cloudflare: This is one of those errors that should never happen since we control both the response and the receiver. If you receive this as a user there would be nothing to do other than file a bug.

packages/agents/src/mcp/index.ts
jmorrell-cloudflare: See other replies. I think a 500 makes sense in this case, but let me know if you think otherwise

packages/agents/src/mcp/index.ts
irvinebroque: Makes sense thanks!

packages/agents/src/mcp/index.ts
irvinebroque: Sounds good

packages/agents/src/mcp/index.ts
irvinebroque: Cool with me

threepointone (approved):

threepointone:
closing this since it was landed in https://github.com/cloudflare/agents/pull/186

PR #186, opened 2025-04-17
2/ Add support for HTTP Streamable MCP Transport

jmorrell-cloudflare:
Note: Stacked on top of #185 *

This adds support for the new Streamable HTTP transport alongside the existing SSE transport. No code changes on the client should be necessary, and you can serve both at once with the same codebase.

This only supports `POST` requests for now. Resumability et al will come later, which we should also not require any further changes.

Setting up a worker to serve these transports on different endpoints would look like this:

```ts
export default {
	fetch(request: Request, env: Env, ctx: ExecutionContext) {
		let url = new URL(request.url);
		let path = url.pathname;
		if (path.startsWith('/sse')) {
			return MyMcpAgent.serveSSE('/sse').fetch(request, env, ctx);
		}
		if (path.startsWith('/mcp')) {
			return MyMcpAgent.serve('/mcp').fetch(request, env, ctx);
		}
	},
};
```

Marking as draft for now since I still have a few things I want to clean up and more validation to do. Initial testing with [the unreleased MCP Inspector branch](https://github.com/modelcontextprotocol/inspector/pull/294) seems to work well!

* Not sure if you can set up a stacked diff on github when both branches are remote so that you can only see the second set of changes here? Please let me know the secret it is

### DIFF
--- a/packages/agents/src/mcp/index.ts
+++ b/packages/agents/src/mcp/index.ts
@@ -213,14 +213,14 @@ export abstract class McpAgent<
     const url = new URL(request.url);
     const path = url.pathname;
 
-    // This session is going to communicate via the SSE protocol
     switch (path) {
+      // This session is going to communicate via the SSE protocol
       case "/sse": {
         // For SSE connections, we can only have one open connection per session
         // If we get an upgrade while already connected, we should error
         const websockets = this.ctx.getWebSockets();
         if (websockets.length > 0) {
-          return new Response("WebSocket already connected", { status: 400 });
+          return new Response("Websocket already connected", { status: 400 });
         }
 
         // This connection must use the SSE protocol
--- a/packages/agents/src/mcp/index.ts
+++ b/packages/agents/src/mcp/index.ts
@@ -1,5 +1,6 @@
 import { DurableObject } from "cloudflare:workers";
 import { Agent } from "../";
+import type { WSMessage } from "../";
 import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import type { Connection } from "../";
 import type { JSONRPCMessage } from "@modelcontextprotocol/sdk/types.js";
@@ -36,7 +37,7 @@ interface CORSOptions {
   maxAge?: number;
 }
 
-class McpTransport implements Transport {
+class McpSSETransport implements Transport {
   onclose?: () => void;
   onerror?: (error: Error) => void;
   onmessage?: (message: JSONRPCMessage) => void;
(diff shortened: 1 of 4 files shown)


packages/agents/src/mcp/index.ts
```diff
@@ -142,14 +312,27 @@ export abstract class McpAgent<
...
-    await this.server.connect(this.#transport);
+    if (this.#protocol === "sse") {
+      this.#transport = new McpSSETransport(() => this.getWebSocket());
+      await this.server.connect(this.#transport);
+    } else if (this.#protocol === "streamable-http") {
+      this.#transport = new McpStreamableHttpTransport(
+        (id) => this.getWebSocketForResponseID(id),
+        (id) => this.#requestIdToConnectionId.delete(id)
+      );
+      await this.server.connect(this.#transport);
```
ashubham, line 273: nit: looks to be common code with line 328. Can move out of the the `if .. else`

packages/agents/src/mcp/index.ts
```diff
@@ -455,30 +717,389 @@ export abstract class McpAgent<
...
+              ws.send(JSON.stringify(message));
+            }
+
+            // closing the websocket will also close the SSE connection
+            ws.close();
+
+            return new Response(null, { status: 202 });
+          }
+
+          for (const message of messages) {
```
ashubham, line 1027: Maybe I am missing something, we already have access to parsedMessages at this point right ? Why do we need to reparse ?

packages/agents/src/mcp/index.ts
```diff
@@ -455,30 +719,389 @@ export abstract class McpAgent<
...
+          const writer = writable.getWriter();
+          const encoder = new TextEncoder();
+
+          // Connect to the Durable Object via WebSocket
+          const upgradeUrl = new URL(request.url);
+          upgradeUrl.pathname = "/streamable-http";
+          const response = await doStub.fetch(
+            new Request(upgradeUrl, {
+              headers: {
+                Upgrade: "websocket",
```
ashubham, line 921: Why do we create a websocket for each interaction? It could just be parallel HTTP POSTs per request id to the DO ? And we stream back responses as they arrive.

jmorrell-cloudflare: Using websockets here allows the DO to hibernate, which could be important for long-running requests.

packages/agents/src/mcp/index.ts
jmorrell-cloudflare: Somewhat poor naming on my part. It's determining which type of message we have: request, response, error, or notification. I'll be revisiting this and seeing if I can use the new helper methods to make this more idiomatic.

packages/agents/src/mcp/index.ts
jmorrell-cloudflare: You can, but then typescript loses track of the type of the transport and thinks it might be `undefined` requiring null checks or an assertion. The type safety and clarity is worth the repetition imo.

packages/agents/src/mcp/index.ts
jmorrell-cloudflare: I ditched the union type, and this should look more familiar now

packages/agents/src/mcp/index.ts
ashubham: Interesting. I would assume the DO would hibernate if it were idle (no incoming http post) after a timeout. That seems simpler than detecting an idle websocket ?

threepointone:
this is breaking types in the /mcp example. let's pair and figure it out.

threepointone:
passed lints/typechecks, now running tests in CI too

threepointone:
The tests pass, but seem to log this workerd error. Wonder what that's about 

```
workerd/io/worker.c++:2097: info: uncaught exception; source = Uncaught; stack = Error: /worker.ts changed, invalidating this Durable Object. Please retry the `DurableObjectStub#fetch()` call.
    at home/runner/work/agents/agents/node_modules/@cloudflare/vitest-pool-workers/dist/worker/lib/cloudflare/test-internal.mjs:508:15
workerd/jsg/exception.c++:140: info: Annotating with brokenness; internalMessage = jsg.Error: /worker.ts changed, invalidating this Durable Object. Please retry the `DurableObjectStub#fetch()` call.; brokennessReason = broken.inputGateBroken
```

PR #202, opened 2025-04-23
Revert Promise.resolve in event handlers

jmorrell-cloudflare:
Following up from https://github.com/cloudflare/agents/pull/186

Adding these back emits these logs in tests:

```
workerd/io/worker.c++:1020: info: console warning; message = An event handler returned a promise that will be ignored. Event handlers should not have a return value and should not be async functions.
```

I'm not sure how to best resolve

### DIFF
--- a/.changeset/good-coins-repeat.md
+++ b/.changeset/good-coins-repeat.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
+await stream writer calls in websocket handlers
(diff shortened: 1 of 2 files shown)


threepointone:
I'll take this over and fix it, thanks for kicking it off!

threepointone:
pushed changes, wanna test and make sure I didn't break anything?

threepointone:
hide whitespace when looking at the diff, and my changes will be more obvious

threepointone:
hide whitespace when looking at the diff, and my changes will be more obvious

Issue #194, opened 2025-04-21
Need Help: How to push props to MCPAgent?


Puliczek:
Hi,

I am trying to pass a `userId` (a prop) to `McpAgent`, but I'm having trouble getting it to work. Below is the code snippet of my current setup:

For the router:
```typescript
return MyMCP.mount("/sse").fetch(request, env, ctx);
```

For the `McpAgent`:
```typescript
export class MyMCP extends McpAgent<Env, {}, { userid: string }> {
  // Class implementation
}
```

I want to pass the `userId` to `MyMCP` from the router. The `userId` is obtained from the URL, for example, `/sse?userId=123` or `/123/sse`. Either format is acceptable for me, but I haven't been able to get it to work correctly.

Could someone provide a simple example of how to achieve this?

Thank you!

irvinebroque:
Could try implementing your own `mount()` method, and passing the userID into the underlying Durable Object

https://github.com/cloudflare/agents/blob/main/packages/agents/src/mcp/index.ts#L371

But generally curious why in your setup a user ID is passed in like this? Typically you'd get back this information from the upstream identity provided in the OAuth flow?

Puliczek:
@irvinebroque Thanks, I'll give it a try :)

I've stopped using OAuth because it's too complex for my needs, and cursor doesn't support it. Instead, I plan to use a simple GUID as userId to select the namespace in Vectorized.

I want to get a URL parameter value in MCPAgent, but I couldn't figure out how. My idea was to pass it through props instead :)

Here's more detail about what I'm trying to achieve: https://developers.cloudflare.com/agents/api-reference/calling-agents/

I just want an MCPAgent per user based on a URL parameter, but it doesn't work with MCPAgents.

irvinebroque:
Makes sense — maybe start here: https://github.com/cloudflare/ai/blob/main/demos/mcp-server-bearer-auth/src/index.ts

Right now the `MCPAgent` class is coupled to the session in ways that make it a bit tricky to get a per-user MCP. @jmorrell-cloudflare @geelen working on a vNext that makes this kind of pattern easier

Puliczek:
Thanks, @irvinebroque! That was the missing piece I needed:

My Router:
```typescript
ctx.props = {
  userId: userId,
};

return MyMCP.mount("/sse").fetch(request, env, ctx);
```

When I set up ```ctx.props```, it's visible in MCPAgent with this code:
```typescript
console.log("userid: " + this.props.userid);
```

[...]

PR #203, opened 2025-04-23
Fix streamable hibernation issues

jmorrell-cloudflare:
Fixes an error introduced in https://github.com/cloudflare/agents/pull/186

The first the the Streamable worker does on an incoming request with a `sessionId` is ask the DO if it's been initialized. If it has not, then the DO has never received an initialize message, and we should reject it.

In the current version once a DO hibernates, this always returns `false`.

However, we're conflating two different types of "initialization" here:

- We call "init" on the `McpAgent` class to allow the user to define the `McpServer`
- There is an "initialization" message that is the first message the Streamable server receives

This fixes the issue by having an explicit `setInitialized()` method, and tracking the second kind of authentication in DO storage.

There was also an additional error with the `this.init?.()` call where it was not `await`'d. Once you await you get errors because it is being run twice and trying to define the same tool twice. These were being silently ignored.

### DIFF
--- a/packages/agents/src/mcp/index.ts
+++ b/packages/agents/src/mcp/index.ts
@@ -259,7 +259,7 @@ export abstract class McpAgent<
     this.#transportType = (await this.ctx.storage.get(
       "transportType"
     )) as TransportType;
-    this.init?.();
+    await this._init(this.props);
 
     // Connect to the MCP server
     if (this.#transportType === "sse") {
@@ -285,7 +285,11 @@ export abstract class McpAgent<
 
   async _init(props: Props) {
     await this.ctx.storage.put("props", props ?? {});
-    await this.ctx.storage.put("transportType", "unset");
+
+    // if transportType is not set, set it to unset
+    if (!this.ctx.storage.get("transportType")) {
+      await this.ctx.storage.put("transportType", "unset");
+    }
     this.props = props;
     if (!this.initRun) {
       this.initRun = true;
@@ -293,7 +297,12 @@ export abstract class McpAgent<
     }
   }
 
-  isInitialized() {
+  async isInitialized() {
+    if (this.#status !== "started") {
+      // This means the server "woke up" after hibernation
+      // so we need to hydrate it again
+      await this.#initialize();
+    }
     return this.initRun;
   }
 
--- a/packages/agents/src/mcp/index.ts
+++ b/packages/agents/src/mcp/index.ts
@@ -285,8 +285,6 @@ export abstract class McpAgent<
 
   async _init(props: Props) {
     await this.ctx.storage.put("props", props ?? {});
-
-    // if transportType is not set, set it to unset
     if (!this.ctx.storage.get("transportType")) {
       await this.ctx.storage.put("transportType", "unset");
     }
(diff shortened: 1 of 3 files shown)


packages/agents/src/tests/mcp-streamable-http.test.ts
```diff
@@ -269,7 +269,6 @@ describe("McpAgent Streamable HTTP Transport", () => {
 
   it("should reject invalid session ID", async () => {
     const ctx = createExecutionContext();
-    const sessionId = await initializeServer(ctx);
```
jmorrell-cloudflare, line 272: This is unused in the rest of the test and can be safely removed without affecting the test logic

packages/agents/src/mcp/index.ts
```diff
@@ -285,16 +285,22 @@ export abstract class McpAgent<
...
   }
 
-  isInitialized() {
-    return this.initRun;
+  async setInitialized() {
+    await this.ctx.storage.put("initialized", true);
+  }
+
+  async isInitialized() {
+    return (await this.ctx.storage.get("initialized")) === true;
```
jmorrell-cloudflare, line 303: This is a little odd because this is only used by the streamable transport, and not the SSE transport.

packages/agents/src/mcp/index.ts
```diff
@@ -285,16 +285,22 @@ export abstract class McpAgent<
 
   async _init(props: Props) {
     await this.ctx.storage.put("props", props ?? {});
-    await this.ctx.storage.put("transportType", "unset");
+    if (!this.ctx.storage.get("transportType")) {
+      await this.ctx.storage.put("transportType", "unset");
+    }
```
jmorrell-cloudflare, line 290: Without this guard this would be overwritten when the DO wakes up. The complexity of our "boot" logic is higher than I would like.

threepointone (approved):

packages/agents/src/mcp/index.ts
```diff
@@ -898,7 +904,7 @@ export abstract class McpAgent<
           const isInitialized = await doStub.isInitialized();
 
           if (isInitializationRequest) {
-            await doStub._init(ctx.props);
```
BrandonNoad, line 901: I have a question about this change. `ctx` is no longer used in `static serve`. So there doesn't seem to be a way to provide additional context from the original request to `init` when defining your MCP tools (e.g., header values)?

Also, wasn't the OAuth flow relying on that? https://github.com/cloudflare/workers-oauth-provider/blob/main/src/oauth-provider.ts#L1857

See my comment here for additional details: https://github.com/cloudflare/agents/issues/194#issuecomment-2846010072

jmorrell-cloudflare: Good catch! You're absolutely right. Let me add that back in

jmorrell-cloudflare: Note that this is only an issue if you need to access the results of auth in your MCP tools. I *think* the auth flow itself will still work.

jmorrell-cloudflare: https://github.com/cloudflare/agents/pull/246

PR #208, opened 2025-04-24
fix: coordinate request caching and suspense

a-type:
... for initial messages

Refactors the usage of suspense and effects to ensure the cached request promise remains in the cache during React dev mode / strict mode, which runs all effects twice.

Mainly this consists of making sure the effect which cleans up the cache is also the one which adds to the cache, so that multiple arbitrary invocations of the effect will result in the intended state with the request cached after the dust settles.

To accomplish this I modified the code around `use` to assign the request promise to a variable which can be cached by the effect.

While not elegant, I found that I had to also keep caching the equest in `doGetInitialMessages`, as this initial cache, while appearing redundant, covers any possible concurrent component renders before the first effects are run.

Since all this behavior is rather esoteric and dependent on subtle React behaviors, I've opted to include a test for correct behavior. I used this test to validate the problem originally, so I'm confident the fix is right. To observe the original problem, comment out line 103 of `ai-react.tsx`, which re-caches the request, and see that in the test the initial messages re refetched on rerender (note: you have to restart Vitest to pick up on source changes).

### DIFF
--- a/packages/agents/src/react-tests/use-agent-chat.test.tsx
+++ b/packages/agents/src/react-tests/use-agent-chat.test.tsx
@@ -1,0 +1,91 @@
+import { act, StrictMode, Suspense } from "react";
+import { describe, expect, it, vi } from "vitest";
+import { render } from "vitest-browser-react";
+import { useAgentChat } from "../ai-react";
+import type { useAgent } from "../react";
+
+// mock the @ai-sdk/react package
+vi.mock("@ai-sdk/react", () => ({
+  useChat: vi.fn((args) => ({
+    messages: args.initialMessages,
+    setMessages: vi.fn(),
+  })),
+}));
+
+/**
+ * Unit tests for the hook functionality which mock the network
+ * layer and @ai-sdk dependencies.
+ */
+describe("useAgentChat", () => {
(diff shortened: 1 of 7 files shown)


a-type:
The new test suite for React requires Puppeteer. I attempted to update relevant scripts and workflows, but will have to see if it works in real life.

packages/agents/src/react-tests/use-agent-chat.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+    // mocking the agent with a subset of fields used in useAgentChat
+    const mockAgent: ReturnType<typeof useAgent> = {
+      id: "fake-agent",
+      name: "fake-agent",
+      _url: "ws://localhost:3000",
+      _pkurl: "ws://localhost:3000",
+      addEventListener: vi.fn(),
+      removeEventListener: vi.fn(),
+      send: vi.fn(),
+    } as any;
```
threepointone:
```suggestion
      send: vi.fn(),
      // biome-ignore lint/suspicious/noExplicitAny: <explanation>
    } as any;
```

threepointone:
This looks great! I'mma land this, thank you so much!

PR #212, opened 2025-04-25
do not remove search params and hash from mcp endpoint message

pbteja1998:
Reference: https://github.com/modelcontextprotocol/typescript-sdk/pull/177


threepointone:
landing because this seems alright. thanks for the PR!

threepointone (approved):

Issue #210, opened 2025-04-24
Env type on fetch causes type errors


jmorrell-cloudflare:
I think all of our MCP demo apps have had `@ts-ignore` added to work around this issue. 

<img width="669" alt="Image" src="https://github.com/user-attachments/assets/f885cc4f-adb7-4ef7-9980-4bfd0aab7dc2" />

```
Argument of type 'Env' is not assignable to parameter of type 'Record<string, DurableObjectNamespace<McpAgent<unknown, unknown, Record<string, unknown>>>>'.
  Index signature for type 'string' is missing in type 'Env'.
```

This comes from the type signature of the `fetch` function we're returning in the static methods `serve` and `serveSSE`. https://github.com/cloudflare/agents/blob/main/packages/agents/src/mcp/index.ts#L537

```typescript
fetch: async (
        request: Request,
        env: Record<string, DurableObjectNamespace<McpAgent>>,
        ctx: ExecutionContext
      ): Promise<Response>
```

Static methods do not share the type parameters of the class. I think we can likely resolve this by adding a parameter to the `fetch` functions themselves. Example: https://github.com/cloudflare/agents/compare/671bdf7...jmorrell/env-type-issue

jmorrell-cloudflare:
I think we likely also need a check here if the binding that we get passes isn't the DO we expect: https://github.com/cloudflare/agents/blob/main/packages/agents/src/mcp/index.ts#L545 and a good error message with guidance since this is likely happening in development because they haven't set up a DO yet.

jmorrell-cloudflare:
Fixed in #219

PR #219, opened 2025-04-25
Fix type error for McpAgent.fetch

jmorrell-cloudflare:
An attempt at fixing https://github.com/cloudflare/agents/issues/210

No more red squiggles or `@ts-ignore` in our demo code!

Monosnap index ts — remote-mcp-authless 2025-04-25 08-45-25

### DIFF
--- a/packages/agents/src/mcp/index.ts
+++ b/packages/agents/src/mcp/index.ts
@@ -542,12 +542,19 @@ export abstract class McpAgent<
         if (corsResponse) return corsResponse;
 
         const url = new URL(request.url);
+        const bindingValue = env[binding as keyof typeof env] as unknown;
 
-        // Ensure we have an MCP agent as a binding
-        // TODO: How can we address this in a type-safe way??
-        const namespace = env[
-          binding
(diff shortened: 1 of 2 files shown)


packages/agents/src/mcp/index.ts
```diff
@@ -532,17 +532,32 @@ export abstract class McpAgent<
...
+          );
+          return new Response("Invalid binding", { status: 500 });
+        }
+
+        // Ensure that the biding is to a DurableObject
+        if (bindingValue.toString() !== "[object DurableObjectNamespace]") {
+          return new Response("Invalid binding", { status: 500 });
+        }
+
+        const namespace = bindingValue as DurableObjectNamespace<McpAgent>;
```
jmorrell-cloudflare, line 560: This checks to make sure we have a binding, and that it's a DO binding, but then assumes that it must be an `McpAgent` because I couldn't figure out how to dynamically test that.

Not perfect, but a lot better than what we have now.

threepointone (approved):

PR #238, opened 2025-04-30
feat(obs): basic agent instrumentation

zebp:
Adds a new `observability` property to Agents that receives events during the Agent's exeuction that can be used for configurable observability. The default observability implementation prints the events to console, but in the future this could be hooked up to a UI or some other observability tooling.

### DIFF
--- a/packages/agents/src/observability/index.ts
+++ b/packages/agents/src/observability/index.ts
@@ -1,0 +1,106 @@
+import type { Message } from "ai";
+import type { Schedule } from "../index";
+import { getCurrentAgent } from "../index";
+
+type BaseEvent<
+  T extends string,
+  Payload extends Record<string, unknown> = {},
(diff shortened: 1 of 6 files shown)


packages/agents/src/observability/index.ts
```diff
@@ -0,0 +1,111 @@
...
+ */
+export const genericObservability: Observability = {
+  emit(event) {
+    // In local mode, we display a pretty-print version of the event for easier debugging.
+    if (isLocalMode()) {
+      console.log(event.displayMessage);
+      return;
+    }
+
+    console.log(event);
```
zebp, line 88: I'm not sure if we should keep this log as the default for non-local mode, there could be billing implications if the user has Workers Logs enabled ($0.30 per million logs) @threepointone what do you think?

PR #249, opened 2025-05-03
chore: add top level default types to package.json

dexxiez:
- Add top level default types to package.json for both modules
- Works in conjunction with explicit exports object
- This lets the NPM registry know that this project indeed does support TS.

Don't bother with changeset, when it happens it happens.


threepointone (approved):
But I like changesets for giving credit in changelogs! Thanks for the PR, landing.

PR #251, opened 2025-05-04
Ensure isLoading is false after you `stop` an ongoing chat agent request

brettimus:
Fixes  cloudflare/agents-starter#69 (tested locally) - big thank you to @Gyurmatag

***

So in https://github.com/cloudflare/agents/pull/105, I forgot to also close the controller after sending an abort signal. This meant the `isLoading` from the chat agent hook didn't turn back to false.


threepointone (approved):
Thank you!!!

Issue #250, opened 2025-05-04
mcp-client: auth callback fails due to hibernation


klaemo:
So, I'm trying to connect to an authenticated mcp server from my chat agent. I basically threw  the mcp-client example into the starter agent. And it almost worked 😅 But it fails to recognize the authentication callback url when i get redirected back from the mcp server's auth url. 
I added some logging and noticed the `callbackUrls` of the agent's `this.mcp` `MCPClientManager`  was empty. This is why `this.mcp.isCallbackRequest(request)` always returns false.

Disabling hibernation of the agent `MyAgent.options.hibernate = false;` seems to fix it.

This is reproducible with the examples/mcp-client example

threepointone:
@cmsparks

klaemo:
besides, i'm curious how to persist auth state across DO hibernations/restarts. here's the scenario i have in mind:
- user starts (chat) session
- user sets up a connection to an mcp server in this session including auth
- user goes away for a while or schedules task for later that's going to use the mcp server
- DO goes to sleep
- DO wakes up again and should be able to use the mcp server on the user's behalf

what i'm doing now is reconnecting to the previously connected mcp server `onStart`, but i couldn't figure out how to resurrect the `DurableObjectOAuthClientProvider`'s state. 🤔

cmsparks:
Ah yeah, I think that problem will happen if the auth popup happens to open in the same tab (I _think_ it might depend on your browser settings). In most cases, the popup should be a popup in a separate window, but if not the DO will hibernate because the WS connection ends.

In the example, I didn't implement reconnection after hibernation because we were moving the MCP client to become a first class attribute of the `Agent` class. In that scenario, ideally the agent just handles the lifecycle/reconnection, so it might not be worth overcomplicating the example.

> besides, i'm curious how to persist auth state across DO hibernations/restarts. here's the scenario i have in mind:
> 
>     * user starts (chat) session
> 
>     * user sets up a connection to an mcp server in this session including auth
> 
>     * user goes away for a while or schedules task for later that's going to use the mcp server
> 
>     * DO goes to sleep
> 
>     * DO wakes up again and should be able to use the mcp server on the user's behalf
> 
> 
> what i'm doing now is reconnecting to the previously connected mcp server `onStart`, but i couldn't figure out how to resurrect the `DurableObjectOAuthClientProvider`'s state. 🤔

[...]

PR #253, opened 2025-05-05
fix: allow overriding fetch and request headers in SSEEdgeClientTransport

adesege:
This PR enhances SSEEdgeClientTransport to support custom fetch options and merges custom request headers with workerOptions headers.

In my use case, I was trying to pass along CF access id and secrets as headers to a remote mcp server but those headers were not sent because SSEEdgeClientTransport is overriding the fetch function and doesn't call user-provided fetch function when it exists.

# How to test
1. Connect to an mcp server behind Cloudflare access using a clone of this branch.
```ts
manager.connect("https://mcp.example.com/sse", {
      transport: {
        eventSourceInit: {
          fetch: (url, init) => fetch(url, {
            ...init,
            headers: {
              ...init?.headers,
              'CF-Access-Client-Id': env.CF_ACCESS_CLIENT_ID,
              'CF-Access-Client-Secret': env.CF_ACCESS_CLIENT_SECRET
            }
          })
        },
        //requestInit: {
          //headers: {
            //'CF-Access-Client-Id': env.CF_ACCESS_CLIENT_ID,
            //'CF-Access-Client-Secret': env.CF_ACCESS_CLIENT_SECRET,
          //}
        //}
      }
    })
```
2. Assert that you do not get a 403 error

PR #254, opened 2025-05-05
Move lifecycle and auth handling for MCP servers into the Agent class

cmsparks:
Now that the MCP Client manager is a first class part of the Agent class, we should have it handle:
* Lifecycle (saving connection data/state, reconnecting, etc)
* Auth (callbacks + oauth provider)

This makes connecting to MCP servers _significantly_ easier and fixes some bugs with the existing mcp client example. In particular, reconnection as discussed in #250


packages/agents/src/index.ts
```diff
@@ -776,6 +897,76 @@ export class Agent<Env, State = unknown> extends Server<Env> {
...
+   * @param callbackHost Base host for the agent, used for the redirect URI.
+   * @param agentsPrefix agents routing prefix if not using `agents`
+   * @returns authUrl
+   */
+  async addMcpServer(
+    url: string,
+    callbackHost: string,
+    agentsPrefix = "agents"
+  ): Promise<{ id: string; authUrl: string | undefined }> {
+    const callback_url = `${callbackHost}/${agentsPrefix}/${camelCaseToKebabCase(this.#ParentClass.name)}/${this.name}/callback`;
```
cmsparks: I don't think there's any way to get around requiring the user to pass in the host because calls to `addMcpServer` isn't necessarily going to occur in the context of a request.

packages/agents/src/react.tsx
```diff
@@ -39,6 +39,7 @@ export type UseAgentOptions<State = unknown> = Omit<
   name?: string;
   /** Called when the Agent's state is updated */
   onStateUpdate?: (state: State, source: "server" | "client") => void;
+  onMcpUpdate?: (mcpServers: MCPServerState) => void;
```
cmsparks: Not sure if we want to expose an `addMcpServer()` function as a native capability of `useAgent`, but I decided to just provide a hook into the reading MCP server state. We could always add that functionality later.

packages/agents/src/index.ts
```diff
@@ -159,6 +166,43 @@ function getNextCronTime(cron: string) {
...
+      server_url: string;
+      auth_url: string | null;
+      state:
+        | "authenticating"
+        | "connecting"
+        | "ready"
+        | "discovering"
+        | "failed";
+    };
+  };
```
klaemo: could be nice to have the server as it's own importable type in case someone wants pass it around in their app

packages/agents/src/index.ts
```diff
@@ -776,6 +908,80 @@ export class Agent<Env, State = unknown> extends Server<Env> {
...
+   * @param url MCP Server SSE URL
+   * @param callbackHost Base host for the agent, used for the redirect URI.
+   * @param agentsPrefix agents routing prefix if not using `agents`
+   * @returns authUrl
+   */
+  async addMcpServer(
+    url: string,
+    callbackHost: string,
+    agentsPrefix = "agents"
+  ): Promise<{ id: string; authUrl: string | undefined }> {
```
klaemo: i think it could be useful to be able to give the server a name. i kind of had that use case because i wanted to dynamically show something like `Please authenticate with ${serviceName}` or `Connected to ${serviceName}`. i could keep a mapping of url -> name/metadata somewhere, but meh...

klaemo:
this is awesome, btw! :)

packages/agents/src/index.ts
cmsparks: Yeah good point, I might add that. 

I wish I could grab the serverName from the initialization handshake, but unfortunately that requires authorizing beforehand, which wouldn't solve your case: https://modelcontextprotocol.io/specification/2025-03-26/basic/lifecycle#initialization

packages/agents/src/index.ts
```diff
@@ -776,6 +908,80 @@ export class Agent<Env, State = unknown> extends Server<Env> {
...
+   *
+   * @param url MCP Server SSE URL
+   * @param callbackHost Base host for the agent, used for the redirect URI.
+   * @param agentsPrefix agents routing prefix if not using `agents`
+   * @returns authUrl
+   */
+  async addMcpServer(
+    url: string,
+    callbackHost: string,
+    agentsPrefix = "agents"
```
cmsparks: - [x] Add additional MCP connect options here

packages/agents/src/index.ts
threepointone: well that's super annoying. should we consider ading it as property (static or otherwise) on the Agent class? ping me and e can talk about it.

packages/agents/src/react.tsx
threepointone: yeah we probably don't want to explicitly add addMcpServer here, and it's trivial to expose it with rpc anyway

packages/agents/src/react.tsx
```diff
@@ -94,6 +95,9 @@ export function useAgent<State = unknown>(
           options.onStateUpdate?.(parsedMessage.state as State, "server");
           return;
         }
+        if (parsedMessage.type === "cf_agent_mcp_servers") {
+          options.onMcpUpdate?.(parsedMessage.mcp as MCPServerState);
```
threepointone:
```suggestion
          options.onMcpUpdate?.(parsedMessage.mcp as MCPServerState);
          return;
```

threepointone (commented):
I did a quick skim on this and it looks great, I'll do a deeper dive and test it all out soon. awesome work.

packages/agents/src/react.tsx
courtney-sims: Should we add a comment to this property too?

cmsparks: added!

cmsparks:
~~- [ ] fix bug where mcp server updates (i.e. list tool updates) don't trigger `cf_agent_mcp_server` events.~~

Actually I'm going to hold off on this until a later PR because it requires adding notification handlers to the client. This requires me to think a bit more thought about the API design because we should expose hooks into server -> client requests/notifications to end users. I think it's fine to leave this "bug" because most clients and servers don't support dynamic server -> client interactions anyways.

packages/agents/src/index.ts
cmsparks: I decided to:
* Remove client capabilities options (because I realized they're redundant with the client options type)
* Prevent the user from configuring transport options from this API because it might change once we add support for the streamable transport.
* keep the client options (which includes capabilities options)

cmsparks: Actually refined this a bit more. Now:
* we allow headers as transport options (because of #253)
* persist options in sqlite storage so we recreate clients with the same options upon reconnection

cmsparks:
Made two changes:
* Consolidated the connection code into `#connectToMcpServerInternal`
* Added transport headers into the options struct

threepointone (approved):
this looks fine. going to land it and release it because I think this needs its tires kicked to shake out any usability issues. great stuff, lfg.

PR #269, opened 2025-05-11
Add type support to react useAgent().call

ruifigueira:
Proposal to add type support to react useAgent() RPC calls. It can infer argument tyoes and ensure we pass all required values.

Usage example:

```ts
// server.ts
class MyAgent extends Agent<typeof env, {}> {
  @callable()
  sayHello(name?: string): string {
    return `Hello, ${name ?? 'World'}!`;
  }
  
  @callable()
  async perform(task: string, p1?: number): Promise<void> {
    // do something
  }

  // not annotated with @callable()
  nonRpc(): void {
  }
}

// index.tsx
import type { MyAgent } from './server.ts';

// ...

const agent = useAgent<{}, MyAgent>({ agent: 'my-agent' });
// return type is promisified
agent.call('sayHello') satisfies Promise<string>;

// @ts-expect-error first argument is not a string
await agent.call('sayHello', [1]);

await agent.call('perform', ['some task', 1]);
await agent.call('perform', ['another task']);
// @ts-expect-error requires parameters
await agent.call('perform');

// we cannot exclude it because typescript doesn't have a way
// to exclude based on decorators
await agent.call('nonRpc');
```

Intellisense in vscode:

image

Unfortunately, I didn't find a way to restrict calls to methods decorated with `@callable()`, so it will also suggest non-RPC methods.
Nevertheless, I excluded the ones declared in the base `Agent` class to ensure intellisense is as clean as possible.

If there are public methods to exclude, it's always possible to exclude the non-RPC ones with `Omit<MyAgent, 'noRpc'>`:

image

### DIFF
--- a/packages/agents/src/react.tsx
+++ b/packages/agents/src/react.tsx
@@ -77,9 +77,9 @@ type AgentPromiseReturnType<T extends AgentMethods<any>, K extends keyof T> =
  * @param options Connection options
  * @returns WebSocket connection with setState and call methods
  */
-export function useAgent<State, AgentT extends {
+export function useAgent<AgentT extends {
   get state(): State;
-}>(
+}, State>(
   options: UseAgentOptions<State>
 ): PartySocket & {
   agent: string;
--- a/packages/agents/src/react.tsx
+++ b/packages/agents/src/react.tsx
@@ -88,6 +88,18 @@ type AgentPromiseReturnType<T extends AgentMethods<any>, K extends keyof T> =
  * @param options Connection options
  * @returns WebSocket connection with setState and call methods
  */
+export function useAgent<State = unknown>(
+  options: UseAgentOptions<State>
+): PartySocket & {
+  agent: string;
+  name: string;
+  setState: (state: State) => void;
+  call: <T = unknown>(
+    method: string,
+    args?: unknown[],
+    streamOptions?: StreamOptions
+  ) => Promise<T>;
+};
 export function useAgent<
   AgentT extends {
     get state(): State;
@@ -109,18 +121,6 @@ export function useAgent<
       args: Parameters<RequiredAgentMethods<AgentT>[T]>,
       streamOptions?: StreamOptions
     ) => AgentPromiseReturnType<AgentT, T>);
-};
-export function useAgent<State = unknown>(
-  options: UseAgentOptions<State>
-): PartySocket & {
-  agent: string;
-  name: string;
-  setState: (state: State) => void;
-  call: <T = unknown>(
-    method: string,
-    args?: unknown[],
-    streamOptions?: StreamOptions
-  ) => Promise<T>;
 };
 export function useAgent<State>(
   options: UseAgentOptions<unknown>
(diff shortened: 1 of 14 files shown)


packages/agents/src/react.tsx
```diff
@@ -134,7 +195,7 @@ export function useAgent<State = unknown>(
   }) as PartySocket & {
     agent: string;
     name: string;
-    setState: (state: State) => void;
+    setState: (state: unknown) => void;
```
threepointone: why change this signature?

packages/agents/src/tests-d/typed-use-agent.test-d.ts
```diff
@@ -0,0 +1,49 @@
...
+  f5: (a: string | undefined) => void;
+  f6: () => Promise<void>;
+}
+
+// @ts-expect-error state doesn't match type A state
+const a2 = useAgent<{ foo: 'bar' }, A>({
+  agent: 'test',
+});
+
+const a1 = useAgent<{}, A>({
```
threepointone: is there any way we can flip the order of this signature without a breaking change? so that it's either useAgent<State> or useAgent<MayAgent, State>

threepointone (commented):
this is great, thanks so much! a couple of questions

packages/agents/src/tests-d/typed-use-agent.test-d.ts
ruifigueira: Hmm, my first approach was without the `State` parameter, and I didn't revert those changes when I added it to the new overload function.

I'll revert those changes.

packages/agents/src/react.tsx
ruifigueira: Hmm, my first approach was without the State parameter, and I didn't revert those changes when I added it to the new overload function.

I'll revert those changes.

packages/agents/src/tests-d/typed-use-agent.test-d.ts
ruifigueira: I did it in fe46392d506410f12db4a8f769ff497851116749. Actually `State` could be infered from `Agent`, but it breaks `useAgent<State>` calls if I omit it or set a default type to it, so we'll need to explicitly provide both `Agent` and `State` types.

threepointone:
I fixed some lint checks, and rebased on top of previous changes. There are a bunch of explicit anys, which I try to keep out of the codebase. If possible, could you replace them? If not possible, it's fine to skip the check with a `//biome-ignore` statement 

You can run npm run check at the root to see whether it would pass checks btw

ruifigueira:
I had to ignore those errors for now. Maybe in the future we can improve type checking for RPC and restrict types to the ones that are [JSON compatible](https://hackernoon.com/mastering-type-safe-json-serialization-in-typescript).

threepointone:
there's still a type failure in rpc.tsx, which shows that it fails if you don't have an explicit type param passed, can we make it so that's still acceptable? a default <Agent, State> value?

ruifigueira:
It seems that switching the order of overload functions fixes it.

threepointone:
This is awesome, thanks for iterating on it. I'll do one final review and test it out tomorrow and land it. Great PR!

PR #273, opened 2025-05-12
Expose getMcpServerState method

cmsparks:
From some internal feedback, someone requested to be able to access the MCPServerState internally from within the Agent class. This seems pretty reasonable to add.

#244 adds a similar `listMcpConnections` method. These are mostly similar, but happy to change the data returned as necessary.


threepointone (approved):

threepointone:
dope, thanks!

packages/agents/src/index.ts
```diff
@@ -1049,12 +1052,12 @@ export class Agent<Env, State = unknown> extends Server<Env> {
...
       JSON.stringify({
         type: "cf_agent_mcp_servers",
-        mcp: this._getMcpServerStateInternal(),
+        mcp: this.getMcpServerState(),
       })
     );
   }
 
-  private _getMcpServerStateInternal(): MCPServersState {
+  getMcpServerState(): MCPServersState {
```
threepointone, line 1060: I landed this, but thinking... how do you feel about just calling this "getMcpServers"?

cmsparks: I'm cool with that, that's probably a better name

threepointone: ok done https://github.com/cloudflare/agents/pull/279

PR #274, opened 2025-05-12
Stub for Agent RPC

ruifigueira:
A proposal for a stub object for Agent RPC that better mimics method calls.

Usage example:

```ts
const { stub } = useAgent<MyAgent, {}>({ agent: "my-agent" });

// return type is promisified
stub.sayHello() satisfies Promise<string>;

// @ts-expect-error first argument is not a string
await stub.sayHello(1);

await stub.perform("some task", 1);
await stub.perform("another task");
// @ts-expect-error requires parameters
await stub.perform();
```

### DIFF
--- a/packages/agents/src/react.tsx
+++ b/packages/agents/src/react.tsx
@@ -81,6 +81,15 @@ type AgentPromiseReturnType<T extends AgentMethods<any>, K extends keyof T> =
     ? ReturnType<T[K]>
     : Promise<ReturnType<T[K]>>;
 
+// biome-ignore lint: suppressions/parse
+type AgentStub<T extends AgentMethods<any> | unknown> = {
+  // biome-ignore lint: suppressions/parse
+  [K in keyof T]: T extends AgentMethods<any>
+    ? (...args: Parameters<T[K]>) => AgentPromiseReturnType<T, K>
+    : // biome-ignore lint: suppressions/parse
+      any;
+};
+
 /**
  * React hook for connecting to an Agent
  * @template State Type of the Agent's state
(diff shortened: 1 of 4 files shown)


packages/agents/src/tests-d/use-agent-stub.test-d.ts
```diff
@@ -0,0 +1,43 @@
...
+  prop: string;
+  f1: () => number;
+  f2: (a: string) => void;
+  f3: (a: number, b: string) => Promise<string>;
+  f4: (a?: string) => void;
+  f5: (a: string | undefined) => void;
+  f6: () => Promise<void>;
+}
+
+const { stub } = useAgent<A, {}>({
```
threepointone, line 15: love this idea. how about instead of calling this "stub", we call it rpc. so it'll be agent.rpc.f1()

ruifigueira: Sure, that even was the initial name but I changed it before submitting the PR 😅 I'll do it at the end of the day

threepointone: well, the nice thing about calling it "stub" is that's what we call it eslewhere in workers already. so maybe we it's fine to call it stub here. lemme review and land this!

packages/agents/src/react.tsx
```diff
@@ -250,6 +262,17 @@ export function useAgent<State>(
...
+  agent.stub = new Proxy<any>(
+    {},
+    {
+      get: (target, method) => {
+        return (...args: unknown[]) => {
+          return call(method as string, args);
+        };
+      },
+    }
+  ) as AgentStub<unknown>;
```
threepointone, line 275: great implementation, so simple

threepointone (approved):
another banger, great PR

PR #275, opened 2025-05-12
Add ContextRouter

cmsparks:
I've expanded the scope of this PR per discussion w/ @threepointone on MCP Routers. This PR adds the `ContextRouter` interface, which does the following:
* Exposes a listTools and listResources function to filter tools and resources
* Exposes a systemPrompt function, which composes the tool, resource, and MCP state into an LLM consumable system prompt. This is meant to be a snippet which you can include in your existing system prompt, rather than a standalone prompt.
* Exposes a setMessages() method, which updates the ContextRouter

I have purposely excluded prompts and resource templates from the ContextRouter interface because those features are not intended to be model controlled. Tools are intended to be LLM controlled, and resources may be LLM controlled.

Given this interface, I've implemented two ContextRouters:
* BaseContextRouter, which preserves the same functionality as before (no filtering of tools) and provides an additional system prompt
* LLMContextRouter, which lets an LLM filter tools and resources using tool calling.

Here is a sample agent using the `LLMContextRouter`:

```ts
export class MyAgent extends Agent<Env, never> {
  private contextRouter;
  private model;
  constructor(ctx: DurableObjectState, env: Env) {
    super(ctx, env);
    const ai = createWorkersAI({
      binding: env.AI,
    });
    this.model = ai("@cf/meta/llama-4-scout-17b-16e-instruct")
    this.contextRouter = new LLMContextRouter(this.model);
    this.mcp = new MCPClientManager("my-agent", "1.0.0", this.contextRouter);
  }

  async generateResponse(message: CoreMessage[]) {
    await this.contextRouter.setMessages(messages);

    const res = await generateText({
      model,
      tools: this.mcp.unstable_getAITools()
      // ...
    })
  }
}
```


packages/agents/src/mcp/client.ts
```diff
@@ -301,6 +302,15 @@ export class MCPClientManager {
       options
     );
   }
+
+  /**
+   * Utility function to fetch MCP information which you can add to an LLM's system prompt.
+   *
+   * @param includeResources Whether to include resources in the prompt. This may be useful if you include a tool to fetch resources OR if the servers you connect to interact with resources.
+   */
+  unstable_getMcpPrompt(includeResources = true): string {
```
Gyurmatag: Hey @cmsparks can we add a function here that will resolve this problem?
https://github.com/cloudflare/agents-starter/pull/80#discussion_r2085090892
Thanks very much!

cmsparks: Continuing conversation in the other agents-starter PR as that seems to be a more fitting location.

packages/agents/src/mcp/client-prompt.ts
```diff
@@ -0,0 +1,34 @@
+import type { MCPClientConnection } from "./client-connection";
+import type { Resource } from "@modelcontextprotocol/sdk/types.js";
+
+export function unstable_getMcpPrompt(
+  conns: Record<string, MCPClientConnection>,
+  includeResources: boolean
+) {
+  return `<integrations_list>
+        You have access to multiple integrations via Model Context Protocol (MCP). These integrations provide you with tools which you can use to execute to complete tasks or retrieive information.
```
courtney-sims: retrieive -> retrieve
which you can use to execute -> which you can execute

packages/agents/src/mcp/client.ts
```diff
@@ -308,12 +287,12 @@ export class MCPClientManager {
    *
    * @param includeResources Whether to include resources in the prompt. This may be useful if you include a tool to fetch resources OR if the servers you connect to interact with resources.
```
courtney-sims, line 290: This function no longer has a parameter

mattzcarey:
@wilhemklopp they are using your prompt area idea.

mattzcarey:
I like this approach in that you can expose some very clean functionality for the user. I dont like that you are hardcoding a workflow to create an agent (llm + set some active tools). ie, 

1. start the contextmanager. 
2. it starts servers
3. it picks tools/resources to activate. 

imo the agent should be free to create itself over time and we should accomodate. It should have a tool which can start server, stop server and the tools should be updated for the next step.

cmsparks:
> imo the agent should be free to create itself over time and we should accomodate. It should have a tool which can start server, stop server and the tools should be updated for the next step.

The ContextRouter itself is more of a way for users to define their own ways of filtering tools/resources. This could be an LLM workflow to filter tools (`LLMContextRouter` included in this PR), well defined state machine, or agentic tool inclusion like you describe. 

This design doesn't preclude implementing a ContextRouter that exposes filtering tools to the Agent. For example, to support the usecase you describe, something like should work: 

```ts
class ToolContextRouter extends BaseContextRouter {
  private activeServers = []
  
  listTools() {
      // filter list tools to only include active servers
      const filteredConnections = {}
      for (const serverId in activeServers) {
        filteredConnections[serverId] = this.connectionManager[serverId]
      }
      
      return getNamespacedData(filteredConnections, "tools");
  }

  // returns tools of activeServers and two utility tools to add/remove servers from context
  getAITools() {
    return {
      ...super.getAITools(),
      addMcpServer: {
          // adds a server to this.activeServers
      },
      removeMcpServer: {
          // removes a server from this.activeServers
      }
    }
  }
  
  systemPrompt() {
    // describes the servers that are accessible and how to use addMcpServer/removeMcpServer
  }
}
```

This is an abbreviated implementation, but I think I'll play around with adding something like the above as an included ContextRouter implementation in this PR like you describe.

cmsparks:
Gonna close this for now bc this is probably stale and I probably should rethink the implementation anyways

PR #289, opened 2025-05-14
Type-safe serializable RPC methods

ruifigueira:
Last one (I guess...)!

In "typed" mode, onlly serializable parameters and return values or their promises are supported, so this is not retro-compatible with my previous changes.

However, I kept retro-compatibility with type-unsafe agent proxies.

Another breaking change is on `unstable_callable`, which enforces decorated method to be serializable.
I can revert those breaking changes, it's just a couple of lines.

### DIFF
--- a/.changeset/gorgeous-donkeys-bake.md
+++ b/.changeset/gorgeous-donkeys-bake.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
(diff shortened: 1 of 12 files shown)


ruifigueira:
I reverted `unstable_callable` changes, streaming use cases were not properly handled.

threepointone:
Another great one, thanks!

PR #291, opened 2025-05-17
docs(readme): fix incorrect usage of npm create for template

sijan2:
The README previously included an incorrect double-dash (`--`) in the  `npm create cloudflare@latest` command, which breaks execution when  copied directly. This commit removes the extra `--` to ensure the  Cloudflare agents-starter template initializes correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ This project is actively evolving. Current focus areas:
 
 ```bash
 # Create a new project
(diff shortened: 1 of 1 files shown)


threepointone:
This isn't right, the extra -- is required to pass the params correctly

sijan2:
sorry closing it i had npm alias with pnpm thats why it was  causing the error.

PR #305, opened 2025-05-25
MCP example: url mixup fix

anthonyringoet:
MCP vs. Agent url mixup. Agent using port `5174`, as setup in `vite.config.ts`


threepointone (approved):
whoops, thanks!

PR #306, opened 2025-05-25
MCP example url mixup number 2

anthonyringoet:
Sorry @threepointone, there was definitely a mixup but also in that last PR.
MCP server is running on 5174 in the example and should be removed I guess.

This lets it actually successfully run the example


threepointone (approved):

threepointone:
Ok!

PR #313, opened 2025-05-29
Refactor MCP server connection logic

threepointone:
Moved MCP server connection logic from the Agent's startup process to dedicated methods `connectMCPServers` and `disconnectMCPServers`. This improves code organization and allows explicit control over MCP server connections.

TODO: rewrite client demo //cc @cmsparks


cmsparks (approved):
This LGTM. I can push up some changes changing the client demo to support this.

threepointone:
ok cool, I'll wait to land this until you make the client demo changes

threepointone:
sigh, this is going to be a serious breaking change, need a better way of messaging this before releasing...

courtney-sims:
What about a migration guide? We could link it in Discord.

cmsparks:
Yeah.. I guess if we're going to make a breaking change to this interface, is there any other MCP breaking changes we'd want to bundle in with this? I don't think there are, but probably a good thing to consider

cmsparks:
Pushed some changes to:
* update the client demo and swap to using RPC
* fix state propogation for disconnection

packages/agents/src/mcp/client.ts
```diff
@@ -65,6 +65,9 @@ export class MCPClientManager {
   }> {
     const id = options.reconnect?.id ?? nanoid(8);
 
+    console.log("connecting!");
+    console.log(id);
+
```
threepointone: delete?

cmsparks: Yup, just removed

examples/mcp-client/src/server.ts
```diff
@@ -9,15 +14,19 @@ type Env = {
...
-      return new Response("Ok", { status: 200 });
-    }
+  @unstable_callable()
+  async disconnectServers() {
+    await this.disconnectMCPServers();
+  }
+
+  @unstable_callable()
+  async connectServers() {
+    await this.connectMCPServers();
```
courtney-sims, line 24: Is there any situation where a user might want to connect only one or some set of servers rather than all that have been added?

cmsparks: I thought about this and the answer is probably yes. But we could always add that as a option in the future

threepointone:
ok I thought this through and spoke to some folks, and it'll just break too many existing users. Instead I'm just going to mark it as deprecated and add a new method or two. (maybe registerMCPAServer)

PR #315, opened 2025-06-03
Add MCP client tests & fix bug where auth fails between disconnections

cmsparks:
This adds a few basic tests for connecting to MCP servers, specifically:
* No Auth
* Bearer Auth
* Simple OAuth
* OAuth with disconnection mid auth flow (simulating hibernation after the user has clicked on the auth_url)
* OAuth with disconnection before auth flow (simulating hibernation before the use clicked on the auth_url, but after the server has been added)

This change also fixes a bug with the latter 2 scenarios, where disconnection before or during an auth flow would cause a token exchange error.

This is stacked on #313


BoomchainLabs (approved):

whoiskatrin (approved):

whoiskatrin:
Closing this, as we've implemented most of it

PR #319, opened 2025-06-12
add lightweight .queue

threepointone:


packages/agents/src/index.ts
```diff
@@ -665,6 +672,73 @@ export class Agent<Env, State = unknown> extends Server<Env> {
     throw new Error("Not implemented");
   }
 
+  /**
+   * Queue a task to be executed in the future
+   * @param payload Payload to pass to the callback
+   * @param callback Name of the method to call
+   * @returns The ID of the queued task
+   */
+  async queue<T = unknown>(payload: T, callback: keyof this): Promise<string> {
```
jahands: it's weird to me that `schedule()` params have the callback before the payload and this one has it after.

Could we be consistent and change this to `(callback: keyof this, payload: T)` ?

threepointone: oops, just a mistake, good catch

PR #332, opened 2025-06-25
MCP connect / reconnect refactor

mchockal:
The PR has fixes for two issues that I've experienced.

**1. onStart() blocked by mcp server reconnects results in DO instance overload**
Awaiting on MCP Server reconnect in onStart() blocks longer when the number of mcp servers increase. I've run into issues where this took so long that the end user request to agent gets cancelled in the worker, even when there's zero additional work done in the user-defined onStart(). 

- Proposed solution(tested) : Don't block on mcp server reconnect in onStart() 
- Gotcha: The one time I can see this causing an issue is when the tool determination happens as the first step in onRequest(), or onMessage() ..and if the mcp server reconnection is incomplete, the llm might not see all available tools and can result in a one-time incorrect tool mapping / 'no appropriate tool found' failure. I haven't run into this issue in my testing, but this IS possible.

**2. MCP client reconnects not working as expected after DO wakes up.** 
  ```
  const { id, authUrl, clientId } = await this.mcp.connect(url, {
      client: options?.client,
      reconnect,
      transport: {
        ...headerTransportOpts,
        authProvider,
      },
    }); 
```

authUrl and clientId are returned only for mcp server connect the very first time.
During reconnect, only id is returned, and the other two are null.

The INSERT OR REPLACE query that follows this statement in `_connectToMcpServerInternal` works as expected for addMcpServer. But the same `_connectToMcpServerInternal` is called in `onStart()` for mcp server reconnect. So the sql statement overwrites the clientId (and authURL, but this isn't important) , resetting them to undefined, thereby creating a NEW mcp client connection for an existing server with a different clientID. This is another one reason why we see 400 Auth Error (its similar to worker shutting down in the middle of a auth request, so a new auth flow is kicked off, but the token exchange is expecting a different clientID.

### CODE REVIEW

**Comment by threepointone** [comment]
thanks for this! will review

File: packages/agents/src/index.ts [hunk 1]
```diff
@@ -543,18 +542,18 @@ export class Agent<Env, State = unknown> extends Server<Env> {
...
-          this.broadcast(
-            JSON.stringify({
-              mcp: this.getMcpServers(),
-              type: "cf_agent_mcp_servers",
-            })
-          );
-
-          await this._tryCatch(() => _onStart());
+          ).then((results) => {
+            console.log('MCP reconnect results:', results)
```
threepointone:
```suggestion
```

threepointone (approved):
this looks good to me, approving for now, will let @cmsparks have a look before I land and release

packages/agents/src/index.ts
```diff
@@ -528,9 +528,8 @@ export class Agent<Env, State = unknown> extends Server<Env> {
           `;
 
           // from DO storage, reconnect to all servers not currently in the oauth flow using our saved auth information
-          await Promise.allSettled(
+          Promise.allSettled(
```
threepointone, line 531: how does removing the await here fix the issue?

packages/agents/src/index.ts
```diff
@@ -528,32 +528,29 @@ export class Agent<Env, State = unknown> extends Server<Env> {
           `;
 
           // from DO storage, reconnect to all servers not currently in the oauth flow using our saved auth information
-          await Promise.allSettled(
-            servers
-              .filter((server) => server.auth_url === null)
```
threepointone, line 533: oh I see you removed this filter as well

threepointone:
I did a format and rebase to pass lint check

cmsparks:
This LGTM

I fixed the same bug here in a different way, but this functionally fixes it in the same way: https://github.com/cloudflare/agents/pull/315/files#diff-565d71ad87e03693406176373280e6737d6676d6c1d97717ae346298ba719efaR110

cmsparks (approved):

PR #335, opened 2025-06-25
Basic A2A Example

mattzcarey:
## AI Summary

- Added examples/a2a/ - Agent-to-Agent framework integration example
- Custom esbuild configuration to handle Express dependencies as external
- Hono-based implementation replacing Express for Cloudflare Workers compatibility
- Agent (DO) version of the A2A TaskStore + AgentExecutor

## Hacky bits
`express` is a dependency of @a2a-js/sdk, there is a hacky build script to remove it since we dont actually use it. I'm sure there is a better way to use esbuild in wrangler directly but I couldnt find it.

Is this the best way to do the routing? 
Should I be using a websocket to handle the streaming since this will not have proper streaming support I dont think.


whoiskatrin:
LGTM as the first iteration

Re: websockers. Let's keep current approach for A2A spec compliance, we can add WebSocket option for better streaming experience later and let the clients choose between HTTP SSE and WebSocket
I think that current implementation works for the spec

PR #342, opened 2025-07-03
email for agents

whoiskatrin:
this adds the ability for agents to receive and respond to real emails. Incoming messages are parsed and routed based on the address or headers, and agents can now automatically reply while keeping track of recent emails. We also added basic state management, loop prevention for auto-replies, and a simple API for testing and debugging. 

+ example agent

### DIFF
--- a/packages/agents/src/index.ts
+++ b/packages/agents/src/index.ts
@@ -9,7 +9,6 @@ import type {
   Tool,
 } from "@modelcontextprotocol/sdk/types.js";
 import { parseCronExpression } from "cron-schedule";
-import { createMimeMessage } from "mimetext";
 import { nanoid } from "nanoid";
 import { EmailMessage } from "cloudflare:email";
 import {
@@ -1440,6 +1439,7 @@ export async function sendEmailWithRouting(
   fromName: string,
   options: EmailSendOptions
 ): Promise<void> {
+  const { createMimeMessage } = await import("mimetext");
   const msg = createMimeMessage();
   msg.setSender({ addr: from, name: fromName });
   msg.setRecipient(options.to);


packages/agents/src/index.ts
```diff
@@ -1173,17 +1193,181 @@ export async function routeAgentRequest<Env>(
...
+      return {
+        agentName: localPart,
+        agentId: subAddress,
+      };
+    }
+
+    return {
+      agentName: localPart,
+      agentId: defaultAgentName || localPart,
+    };
```
deloreyj: With a default on one of these properties we could go one of two ways
1. Multiple Agent namespaces each with a single default agent instance
2. A single default Agent namespace with many instances identified by the `localPart`

Option 2 seems like the much more common use case to me, but I believe this is implementing option 1. Would you consider swapping so the default gets applied to the agentName instead? 

I'd imagine using option 2 when I have a catch-all email route that points at an Email Worker with a single EmailAgent namespace. In that setup it makes sense to me for the email address to be the agentId and hardcode the default namespace.

whoiskatrin: swapped, let me know what's your experience like

examples/email-agent/test-mail.ts
```diff
@@ -0,0 +1,161 @@
+/**
```
ghostwriternr: I think maybe this test script can be updated/discarded? Currently, this seems to demonstrate the overall flow from an explanation pov, but doesn't actually trigger the real flow and/or test the user's setup.

examples/email-agent/test-comprehensive.ts
```diff
@@ -0,0 +1,176 @@
+/**
```
ghostwriternr: Same comment as test-email. I tried going through the instructions, and the endpoints like `/stats` and `toggle-auto-reply` don't exist anymore (maybe they did during development and not needed anymore?)

ghostwriternr: I personally have never used email workers, so this has been a helpful exercise to test this. But I had luck triggering the email worker in this example by using the [curl example from the docs](https://developers.cloudflare.com/email-routing/email-workers/local-development/):
```
curl --request POST 'http://localhost:8787/cdn-cgi/handler/email' \
  --url-query 'from=sender@example.com' \
  --url-query 'to=recipient@example.com' \
  --header 'Content-Type: application/json' \
  --data-raw 'Received: from smtp.example.com (127.0.0.1)
        by cloudflare-email.com (unknown) id 4fwwffRXOpyR
        for <recipient@example.com>; Tue, 27 Aug 2024 15:50:20 +0000
From: "John" <sender@example.com>
Reply-To: sender@example.com
To: recipient@example.com
Subject: Testing Email Workers Local Dev
Content-Type: text/html; charset="windows-1252"
X-Mailer: Curl
Date: Tue, 27 Aug 2024 08:49:44 -0700
Message-ID: <6114391943504294873000@ZSH-GHOSTTY>

Hi there'
```

We can perhaps tailor this to our example and update the README.

packages/agents/src/index.ts
```diff
@@ -1173,17 +1193,183 @@ export async function routeAgentRequest<Env>(
...
+  let EmailMessage: any;
+  try {
+    const cloudflareEmail = await import("cloudflare:email");
+    // biome-ignore lint/suspicious/noExplicitAny: EmailMessage constructor is dynamically imported
+    EmailMessage = cloudflareEmail.EmailMessage;
+  } catch (error) {
+    throw new Error(
+      "cloudflare:email module not available. This function must be called in a Cloudflare Workers environment."
+    );
+  }
```
ghostwriternr: This currently throws this error:
```sh
[wrangler:info] Ready on http://localhost:8787
⎔ Starting local server...
📮 Email received via email handler
📧 Received email from: sender@example.com to: recipient@example.com
📧 Parsed email: {
  from: 'sender@example.com',
  subject: 'Testing Email Workers Local Dev',
  textLength: 0
}
📊 Agent state updated: { emailCount: 2, totalEmails: 2 }
🤖 Sending auto-reply to: sender@example.com
✘ [ERROR] ❌ Failed to send auto-reply: Error: Dynamic require of "path" is not supported

      at __require
  (file:///Users/naresh/github/cloudflare/agents/examples/email-agent/.wrangler/tmp/dev-m3G3q1/index.js:12:9)
      at ../../node_modules/mime-types/index.js
  (file:///Users/naresh/github/cloudflare/agents/node_modules/mime-types/index.js:16:15)
      at __require2
  (file:///Users/naresh/github/cloudflare/agents/packages/agents/dist/chunk-KEXKKQVW.js:15:48)
      at ../../packages/agents/dist/mimetext.node.es-657RU4ZP.js
  (file:///Users/naresh/github/cloudflare/agents/examples/email-agent/.wrangler/tmp/dev-m3G3q1/index.js:17371:18)
      at __init
  (file:///Users/naresh/github/cloudflare/agents/examples/email-agent/.wrangler/tmp/dev-m3G3q1/index.js:15:56)
      at null.<anonymous>
  (file:///Users/naresh/github/cloudflare/agents/packages/agents/src/index.ts:1335:39)
      at async sendEmailWithRouting
  (file:///Users/naresh/github/cloudflare/agents/packages/agents/src/index.ts:1335:33)
```

when hitting:
```sh
curl --request POST 'http://localhost:8787/cdn-cgi/handler/email' \
  --url-query 'from=sender@example.com' \
  --url-query 'to=recipient@example.com' \
  --header 'Content-Type: application/json' \
  --data-raw 'Received: from smtp.example.com (127.0.0.1)
        by cloudflare-email.com (unknown) id 4fwwffRXOpyR
        for <recipient@example.com>; Tue, 27 Aug 2024 15:50:20 +0000
From: "John" <sender@example.com>
Reply-To: sender@example.com
To: recipient@example.com
Subject: Testing Email Workers Local Dev
Content-Type: text/html; charset="windows-1252"
X-Mailer: Curl
Date: Tue, 27 Aug 2024 08:49:44 -0700
Message-ID: <6114391943504294873000@ZSH-GHOSTTY>

Hi there'
```

Looks like the fix requires 3 things:
1. Updating `packages/agents/scripts/build.ts`, `packages/hono-agents/scripts/build.ts` to include "cloudflare:email" in the list of external packages.
2. Changing these dynamic imports to top-level imports and completely remove this highlighted block.
```ts
import { createMimeMessage } from "mimetext";
import { EmailMessage } from "cloudflare:email";
```
```suggestion
```
3. Add `"mimetext": "^3.0.27",` to `packages/agents/package.json`. `npm run build` works without this, but its only because we are using mimetext in one of the examples and so this monorepo package-lock.json still sees it. But there is an obtuse runtime error about 'path' being dynamically imported, which recursively comes from this package being missing in the agents package dependencies.

packages/agents/src/index.ts
```diff
@@ -1173,17 +1193,183 @@ export async function routeAgentRequest<Env>(
...
- */
+export type EmailResolver<Env> = (
+  email: ForwardableEmailMessage,
+  env: Env
+) => Promise<{
+  agentName: string;
+  agentId: string;
+} | null>;
+
+async function defaultEmailResolver<Env>(
```
ghostwriternr: This is a DX comment, but I think is handy to improve. I was wondering if we could leave it this way too, but this particular part confused me a lot when I first looked at this PR and my only way out was to read the source code to understand.

The current behaviour for the user to get one of the pre-configured routers is:
1. For address-based routing, they've gotta do `const resolver = createEmailAddressResolver("default");`
2. For header-based routing, they just skip creating a resolver and use the 'default' behaviour
3. For catch-all routing (which imo feels more simple/default-y), they've gotta do `const resolver = createCatchAllResolver("EmailAgent", "main");`.

But because the default behaviour is 'hidden' + inconsistencies in how they're all invoked (questions I asked myself: why do I need 2 params for a simple catch-all resolver? what is the default behaviour? can I write my own resolver?), it gets a bit confusing. Furthermore, `EmailRoutingOptions` actually needs `defaultAgentName` and `defaultAgentId` IF the developer actually doesn't pass any of the branches defined in the hidden `defaultEmailResolver` method, and we seem to have implicit logic to access `env[routingInfo.agentName]` too.

My proposal to improve this would be to change the signature of `routeAgentEmail` to:
```ts
routeAgentEmail<Env>(
  email: ForwardableEmailMessage,
  env: Env,
  resolver: EmailResolver<Env> // not optional anymore
): Promise<void>
```
This will indicate to the developer clearly that a resolver is always needed. And we can add a docstring + docs to indicate they can use one of our pre-packaged ones. And get rid of the need for them to pass `defaultAgentName` and `defaultAgentId`, which was solving this redundant purpose anyway.

That way, the usage becomes a bit cleaner:
```
  // Three explicit choices - pick exactly one
  const resolver = createAddressBasedResolver("default");
  const resolver = createHeaderBasedResolver();
  const resolver = createCatchAllResolver("EmailAgent", "main");

  // Or write your own
  const resolver = async (email, env) => {
    // custom logic
    return { agentName: "MyAgent", agentId: "some-id" };
  };
```

I do have a feeling its possible to clean up the params for these resolvers further by using sane defaults and letting the user pass optional names they prefer. And possibly even package into a single objects so it becomes a nice builder pattern - but that's completely optional : P
```ts
  - DefaultEmailResolvers.addressBased() - they expect specific format
  - DefaultEmailResolvers.headerBased() - they expect headers to exist
  - DefaultEmailResolvers.catchAll() - they want everything to go somewhere
```

packages/agents/src/index.ts
```diff
@@ -1173,17 +1193,183 @@ export async function routeAgentRequest<Env>(
...
+      agentId: options.defaultAgentId,
+    };
+  }
+
+  if (!routingInfo) {
+    console.warn("No routing information found for email, dropping message");
+    return;
+  }
+
+  const namespace = env[routingInfo.agentName as keyof Env] as AgentNamespace<
```
ghostwriternr: I haven't tested this path yet (needs setting up workers email properly I think) - but I wonder if this is accurate.

ghostwriternr (changes requested):
Overall, I quite like this PR! The multiple resolvers being exposed is super handy, and pre-packaged handling for stuff like threads and auto-replies is super smart. I think it still needs to be tested end-to-end further and the example documentation updated a bit with proper [Email Workers](https://developers.cloudflare.com/email-routing/email-workers/) docs referenced as needed.

packages/agents/src/index.ts
whoiskatrin: fixed, will push soon

packages/agents/src/index.ts
whoiskatrin: fixed

ghostwriternr (approved):
Thanks! LGTM 🚀

ghostwriternr (approved):

PR #345, opened 2025-07-10
potentially a fix for getCurrentAgent returns undefined 

whoiskatrin:
we've implemented auto-wrapping logic that:
- Enumerates all methods in the class
- Filters out built-in framework methods
- Automatically wraps remaining custom methods


docs/context-management.md
```diff
@@ -0,0 +1,124 @@
+# Context Management
```
ghostwriternr, line 1: I think we can have a very succinct version of this in the main README. This one has a lot of implementation details, which can be skipped by devs imo.

packages/agents/src/index.ts
```diff
@@ -247,6 +247,59 @@ export function getCurrentAgent<
...
+}
+
+/**
+ * Decorator to automatically wrap methods with agent context
+ * Ensures getCurrentAgent() works properly within the decorated method
+ * @param connection Optional connection context
+ * @param request Optional request context
+ * @returns Method decorator
+ */
+export function withContext(connection?: Connection, request?: Request) {
```
ghostwriternr: Is this method still needed? Since we're auto-wrapping using withAgentContext, we could likely skip this one.

packages/agents/src/index.ts
```diff
@@ -677,6 +735,112 @@ export class Agent<Env, State = unknown> extends Server<Env> {
...
+          // Check if it's a custom method (not built-in)
+          const descriptor = Object.getOwnPropertyDescriptor(proto, methodName);
+          if (descriptor && typeof descriptor.value === "function") {
+            // Wrap the custom method with context
+            this[methodName as keyof this] = withAgentContext(
+              this,
+              (this[methodName as keyof this] as Function).bind(this)
+            ) as any;
+          }
+        }
```
ghostwriternr: This could become very tricky to maintain as it becomes a target to update any time we add/remove/rename any of our methods (both from Agents & the underlying partyserver). Can we switch to getting the list of methods from the class prototypes? Something along the lines of:
```suggestion
      // Collect all methods from base prototypes (Agent and Server)
      const basePrototypes = [Agent.prototype, Server.prototype];
      const baseMethods = new Set<string>();

      for (const baseProto of basePrototypes) {
        let proto = baseProto;
        while (proto && proto !== Object.prototype) {
          const methodNames = Object.getOwnPropertyNames(proto);
          for (const methodName of methodNames) {
            baseMethods.add(methodName);
          }
          proto = Object.getPrototypeOf(proto);
        }
      }

      // Get all methods from the current instance's prototype chain
      let proto = Object.getPrototypeOf(this);
      let depth = 0;
      while (proto && proto !== Object.prototype && depth < 10) {
        const methodNames = Object.getOwnPropertyNames(proto);

        for (const methodName of methodNames) {
          // Skip if it's a private method or not a function
          if (
            methodName.startsWith("_") ||
            typeof this[methodName as keyof this] !== "function"
          ) {
            continue;
          }

          // If the method doesn't exist in base prototypes, it's a custom method
          if (!baseMethods.has(methodName)) {
            const descriptor = Object.getOwnPropertyDescriptor(proto, methodName);
            if (descriptor && typeof descriptor.value === "function") {
              // Wrap the custom method with context
              this[methodName as keyof this] = withAgentContext(
                this,
                (this[methodName as keyof this] as Function).bind(this)
              ) as any;
            }
          }
        }
```

packages/agents/src/tests/vitest.config.ts
```diff
@@ -2,6 +2,8 @@ import { defineWorkersConfig } from "@cloudflare/vitest-pool-workers/config";
 
 export default defineWorkersConfig({
   test: {
+    setupFiles: ["./test-setup.ts"],
```
ghostwriternr, line 5: If queueMicrotask doesn't need to be polyfilled, can remove this.

packages/agents/src/index.ts
```diff
@@ -677,6 +735,112 @@ export class Agent<Env, State = unknown> extends Server<Env> {
...
+        // Use a try-catch to avoid queueMicrotask issues
+        try {
+          console.warn(
+            "Auto-wrapping custom methods failed:",
+            error instanceof Error ? error.message : String(error)
+          );
+        } catch (logError) {
+          // Ignore logging errors in environments where console.warn might fail
+        }
+      }
```
ghostwriternr: Maybe nothing to resolve and asking for my own understanding: Why do we need this? My understanding is that queueMicrotask works perfectly in workerd.

I'm guessing the polyfill you've added to test-setup.ts is maybe needed due to vitest's console implementation (I don't know this for sure - guessing from the changes)?

packages/agents/src/index.ts
threepointone: this function isn't used anywhere?

ghostwriternr (changes requested):

docs/context-management.md
whoiskatrin: discussed, resolved

packages/agents/src/index.ts
whoiskatrin: No, you are actually right here. queueMicrotask issue was resolved with the polyfill in test-setup.ts, but this bit ehre is just protecting console.warn which should work fine in all environment anywa, will change

packages/agents/src/index.ts
ghostwriternr: As discussed, we should likely resolve this comment proper - all other changes look good to me!

ghostwriternr (changes requested):

threepointone:
aight, fixed some stuff: 

- enumerate methods from prototype 
- make it work in tests 
- make it work with callables

landing this

threepointone (approved):

PR #352, opened 2025-07-11
Email agent

threepointone:
This carries on from https://github.com/cloudflare/agents/pull/342 (thanks @whoiskatrin) 

- I think local dev might actually be a bit broken 
- routeAgentEmail needs some work to be resilient to uppercase/lowercase in resolution 
- examples/email-agent needs a nice ui demo page 

But I think this could land so it doesn't fester


threepointone:
landing this

PR #354, opened 2025-07-14
fix: dequeue items in DB after each task is complete

jahands:
Prevents a single failure from causing all items in the queue from being retried (including previously processed items that were successful).


jahands:
@threepointone I don't have perms to merge in this repo so feel free to merge whenever

PR #359, opened 2025-07-15
Fix email routing being case sensitive

### LINKED ISSUE #356: Is createAddressBasedEmailResolver supposed to be case sensitive?
I had a heck of a time getting createAddressBasedEmailResolver to work, finally realized that I had to have my email formatted as <CaseSensitiveAgent+InstanceName@domain.com>. Is this intentional?

ghostwriternr:
Add support for both CamelCase and kebab-case agent names in email routing, similar to how partyserver does. The fix caches a mapping of agent names to handle case variations, ensuring emails can be routed correctly regardless of the case format used.

Also add tests covering all email resolver types and case variations. Intentionally testing only the resolvers, so we can add more exhaustive tests for the email resolver built on top of vitest-pool-workers later.

fixes #356

### DIFF
--- a/.changeset/quick-maps-walk.md
+++ b/.changeset/quick-maps-walk.md
@@ -2,4 +2,4 @@
 "agents": patch
 ---
 
-Fix email routing to handle both CamelCase and kebab-case agent names
+Fix email routing to be case-insensitive for agent names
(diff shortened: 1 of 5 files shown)


threepointone (approved):
fantastic, thank you for writing tests!

Issue #358, opened 2025-07-14
Feature Request: HTTP Streamable Support in mcp-client Agent Class


arre-ankit:
## Summary
Currently, the `mcp-client` in Agent Class only supports Server-Sent Events (SSE) for streaming connections. I would like to request support for HTTP Streamable connections to enable integration with Cloudflare mcp-client.

## Current Limitation
The existing implementation is limited to SSE, which prevents connection to HTTP Streamable remote MCP servers.

## Requested Feature
Add HTTP Streamable support to the `mcp-client` in Agent Class to enable broader compatibility with different MCP server implementations.

## Use Case
This feature would allow developers to:
- Connect to HTTP Streamable remote MCP servers
- Integrate with Cloudflare `mcp-client` more effectively
- Have more flexibility in choosing streaming protocols for their MCP implementations

## Additional Context
- This feature request was previously discussed in [Discord](https://discord.com/channels/595317990191398933/1377707912247644170/1379241995909726240)
- The response indicated this is already on the roadmap
- This would significantly improve the versatility of the mcp-client for various deployment scenarios

cc @threepointone @whoiskatrin

whoiskatrin:
@arre-ankit take a look at https://github.com/cloudflare/agents/pull/364, i took a stab at this, let me know if that's what you had in mind, happy to discuss

arre-ankit:
Yes, this is exactly what I had in mind — thanks for taking a stab at it! Looks great. 🙌

whoiskatrin:
Closing, added this in the latest release

PR #364, opened 2025-07-16
Feature/http streamable support

whoiskatrin:
add HTTP Streamable transport support for the MCP client, addressing issue #358 
HTTP Streamable is now the default transport with SSE remaining available for backward compatibility
formatted and added new tests


packages/agents/README.md
```diff
@@ -393,6 +393,76 @@ This creates:
 - Intuitive input handling
 - Easy conversation reset
 
+### 🔗 MCP (Model Context Protocol) Integration
```
kentcdodds, line 396: Love to see this!

packages/agents/README.md
```diff
@@ -393,6 +393,76 @@ This creates:
...
+          return {
+            location,
+            temperature: "22°C",
+            conditions: "Sunny"
+          };
+        }
+      }
+    };
+  }
+}
```
kentcdodds: Interesting! So we no longer recommend using the standard SDK?

whoiskatrin: We absolutely still recommend the standard MCP SDK! These *EdgeClientTransport classes are thin compatibility wrappers that extend the official SDK classes (SSEClientTransport and StreamableHTTPClientTransport) to work around Cloudflare Workers limitations (like the new Function() issue in #689)

whoiskatrin: The Edge classes just override the fetch behavior and add auth integration; all the core MCP protocol logic still comes from the official SDK. Think of them as Cloudflare Workers adapters rather than replacements

kentcdodds: If that's the case, then it's surprising to me that the first documentation of this (and likely the first thing people trying to use this for MCP would see) does not use the official SDK 🤔

packages/agents/src/mcp/client-connection.ts
```diff
@@ -18,6 +19,15 @@ import {
...
 import type { AgentsOAuthProvider } from "./do-oauth-client-provider";
 import { SSEEdgeClientTransport } from "./sse-edge";
+import { StreamableHTTPEdgeClientTransport } from "./streamable-http-edge";
+
+export type MCPTransportOptions = (
+  | SSEClientTransportOptions
+  | StreamableHTTPClientTransportOptions
+) & {
+  authProvider?: AgentsOAuthProvider;
+  type?: "sse" | "streamable-http";
```
cmsparks, line 29: Out of scope of this PR, but might be cool if we had an `auto` option. The functionality would be like `useMcp`, where it first tries `streamable-http`, then attempts to fall back to `sse`.

Though might not be relevant once everyone starts moving to streamable.

cmsparks (approved):
LGTM!

packages/agents/README.md
whoiskatrin: it's a miss on my side completely, fixed

packages/agents/src/mcp/client-connection.ts
whoiskatrin: Added this an issue, will do a separate PR! Thanks for taking a look!

examples/mcp-client/README.md
```diff
@@ -15,8 +22,36 @@ Then, follow the steps below to setup the client:
...
+await mcpClient.connect(serverUrl, {
+  transport: {
+    type: "sse",
+    authProvider: myAuthProvider
+  }
+});
+
+// Or using Agent.addMcpServer() (as shown in the example)
+export class MyAgent extends Agent<Env, never> {
+  mcp = new MCPClientManager("my-agent", "1.0.0");
```
threepointone:
```suggestion
```
we already do this inside Agent

packages/agents/src/tests/mcp/client/client-manager.test.ts
```diff
@@ -0,0 +1,224 @@
...
+          }
+        });
+
+        expect(result).toBeDefined();
+        expect(result.id).toBeDefined();
+        expect(typeof result.id).toBe("string");
+        expect(result.id.length).toBe(8);
+      } catch (error) {
+        // Expected to fail in test environment, but transport should be created
+        expect(error).toBeDefined();
```
threepointone: we should probably run a server during test and check against it

packages/agents/src/tests/mcp/client/client-manager.test.ts
```diff
@@ -0,0 +1,224 @@
...
+          transport: {
+            // No type specified, should default to streamable-http
+          }
+        });
+
+        expect(result).toBeDefined();
+        expect(result.id).toBeDefined();
+      } catch (error) {
+        // Expected to fail in test environment
+        expect(error).toBeDefined();
```
threepointone: how do we test that it's default to use http transport?

packages/agents/src/tests/mcp/client/client-manager.test.ts
```diff
@@ -0,0 +1,224 @@
...
+        });
+
+        expect(result).toBeDefined();
+        expect(result.id).toBe("existing-server-id");
+      } catch (error) {
+        // Expected to fail in test environment
+        expect(error).toBeDefined();
+      }
+    });
+
```
threepointone: we should actually test for reconnections here

packages/agents/src/tests/mcp/client/client-manager.test.ts
```diff
@@ -0,0 +1,224 @@
...
+      expect(Object.keys(manager.mcpConnections)).toHaveLength(0);
+
+      try {
+        await manager.connect(mockServerUrl, {
+          transport: {
+            type: "streamable-http"
+          }
+        });
+      } catch (error) {
+        // Expected to fail, but connection should be created
```
threepointone: really? odd behaviour

packages/agents/src/tests/mcp/client/client-manager.test.ts
```diff
@@ -0,0 +1,224 @@
...
+        });
+
+        await manager.connect(mockServerUrl, {
+          transport: {
+            type: "sse"
+          }
+        });
+
+        expect(Object.keys(manager.mcpConnections).length).toBeGreaterThan(0);
+
```
threepointone: should this be 2?

packages/agents/src/tests/mcp/client/client-manager.test.ts
```diff
@@ -0,0 +1,224 @@
...
+        // Should have auth-related properties when auth provider is present
+        expect(result).toBeDefined();
+        expect(result.id).toBeDefined();
+      } catch (error) {
+        // Expected to fail in test environment
+        expect(error).toBeDefined();
+      }
+    });
+  });
```
threepointone: the auth stuff is super involved, probably worth adding a whole other test suite for all of it

packages/agents/src/tests/mcp/transports/streamable-http.test.ts
```diff
@@ -0,0 +1,306 @@
...
+      );
+
+      expect(response.status).toBe(400);
+      const errorData = await response.json();
+      expectErrorResponse(
+        errorData,
+        -32600,
+        /Initialization requests must not include a sessionId/
+      );
+    });
```
threepointone: this error message implies something else (like, it would fail an initial init request if you sent an id). how do we test that a second init request fails anyway?

threepointone (changes requested):
the actual implementation looks good, but I think we need to flesh out the etsts some more.

threepointone:
I don't mind merging this and punting on tests for later

Issue #320, opened 2025-06-13
If the MCP server capabilities initialization fails, the connectionState should be marked "failed"


laulauland:
In the `MCPClientConnection` initialization flow this `Promise.all` might reject, the error doesn't bubble up and the connectionState permanently gets stuck in `discovering`

https://github.com/cloudflare/agents/blob/398c7f5411f3a63f450007f83db7e3f29b6ed4c2/packages/agents/src/mcp/client-connection.ts#L87-L94

We should probably consider marking it `failed` or have another state to reflect that. Happy to submit a PR for this!

whoiskatrin:
@laulauland would love to see a PR for this if you are still up for it :)

laulauland:
here it is! https://github.com/cloudflare/agents/pull/374

PR #374, opened 2025-07-21
Improve MCP client connection resilience with Promise.allSettled

laulauland:
Small change kinda addressing #320 

Instead of failing the connection this would provide a fallback and log the error that registering a capability failed.

Another alternative would be:
- `try..catch` on the `Promise.all`
-  mark the `this.connectionState = failed` and exit the function

Changes:
- Replace Promise.all with Promise.allSettled in MCP client initialization to handle partial capability failures gracefully
- Log failed capabilities instead of throwing errors that break entire connections

### DIFF
--- a/.changeset/sharp-eyes-buy.md
+++ b/.changeset/sharp-eyes-buy.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
+Improve MCP client connection resilience with Promise.allSettled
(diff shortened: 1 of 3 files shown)


whoiskatrin:
@laulauland can you please take a look at failed build? could we add some tests for this? 
can i also ask you to add a changeset? thanks a ton!

laulauland:
Thanks @threepointone for the changeset! I pushed a little test for the client connection as well.

threepointone (approved):
Looks good now, thanks so much for the PR! landing

PR #382, opened 2025-07-28
only send new message

### LINKED ISSUE #377: Only send *new messages* to agent (since previous messages have already been persisted.
Since messages are persisted to the DO, I don't think it makes sense to send all messages in every request. I think it would be better to just send *new* messages and then reconstruct the history of messages from the messages already in the DO's database.

Here is an example on how this can be accomplished with AI SDK: https://ai-sdk.dev/cookbook/next/send-custom-body-from-use-chat

nicksrandall:
Fixes #377


nicksrandall:
I've tested this change using https://github.com/cloudflare/agents-starter and it seems to work! Agent has context to full history of messages but only current message is sent with the request. I believe this PR is ready for review

packages/agents/src/ai-chat-agent.ts
```diff
@@ -230,19 +237,19 @@ export class AIChatAgent<Env = unknown, State = unknown> extends Agent<
...
   ) {
-    this.sql`delete from cf_ai_chat_agent_messages`;
-    for (const message of messages) {
-      this.sql`insert into cf_ai_chat_agent_messages (id, message) values (${
+    for (const message of incomingMessages) {
+      this
+        .sql`insert or replace into cf_ai_chat_agent_messages (id, message) values (${
         message.id
       },${JSON.stringify(message)})`;
     }
```
nicksrandall: Does this need to be wrapped in a `this.ctx.storage.transactionSync`? I think that in every case this will always be an array of one message so I didn't think it was necessary but I'm not 100% sure.

nicksrandall:
@whoiskatrin @threepointone This is ready for review :)

amorriscode:
This also resolves https://github.com/cloudflare/agents/issues/396

packages/agents/src/ai-chat-agent.ts
whoiskatrin: @nicksrandall might be worth wrapping the loop in a transaction. It guarantees consistency if we ever get multiple messages or if this logic changes later. The cost for a single statement is basically negligible, so it’s a nice bit of future-proofing

victorhenrion:
This breaks human-in-the-loop flows, tool result messages aren't processed correctly
You can reproduce by running getWeatherInformation in agents-starter

whoiskatrin:
> This breaks human-in-the-loop flows, tool result messages aren't processed correctly You can reproduce by running getWeatherInformation in agents-starter

yes, this it outdated, especially with ai sdk v5 update

whoiskatrin:
@nicksrandall if you are up for updating this PR based on v5 and checking it all again, that would be much appreciated, thank you

nicksrandall:
@whoiskatrin I updated this PR to work with AI sdk v5. If you're open to it, I could use some help testing to make sure that this is working as expected.

packages/agents/src/ai-chat-agent.ts
nicksrandall: I wrapped this in a transaction.

victorhenrion:
@nicksrandall I believe this is still an issue:

> This breaks human-in-the-loop flows, tool result messages aren't processed correctly You can reproduce by running getWeatherInformation in agents-starter

You need to replace messages when adding a tool result, not just concat (e.g. https://github.com/clavierorg/agents/blob/main/packages/agents/src/ai-chat-agent.ts)

nicksrandall:
@victorhenrion I think I've fixed the issue you mentioned.

nicksrandall:
Ok, I've found some bugs with this implementation and I'm now convinced that the effort to make this work is not worth the value (at least for me). I'm going to close this PR and somebody else is welcome to tackle this if they think it would be useful for them.

PR #392, opened 2025-08-04
fix: Ensure McpAgent props stay current

Maximo-Guk:
Previously props were only updated during initialization but could become stale if ctx.props changed between requests


Maximo-Guk:
Just noticed there's already a PR open https://github.com/cloudflare/agents/pull/337 for addressing this issue!

PR #400, opened 2025-08-07
Add pkg.pr.new workflow for publishing agents package

whoiskatrin:


.github/workflows/pkg-pr-new.yml
```diff
@@ -0,0 +1,29 @@
...
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "18"
```
threepointone: 20

(And I think it can be a number? See the other workflows as a reference)

.github/workflows/pkg-pr-new.yml
```diff
@@ -0,0 +1,29 @@
...
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
```
threepointone: with:
          fetch-depth: 1

see other workflows for ref

.github/workflows/pkg-pr-new.yml
```diff
@@ -0,0 +1,29 @@
+name: Publish to pkg.pr.new
+
+on:
+  push:
+    branches:
+      - main
```
threepointone: don't need this I think, only on pull requests

.github/workflows/pkg-pr-new.yml
```diff
@@ -0,0 +1,28 @@
+name: Publish to pkg.pr.new for pre-release
+
+on:
+  pull_request: {}
+  push:
+    branches:
+      - "**"
```
threepointone: I don't think we want this on pushes to arbitrary branches? Just pull requests

.changeset/stale-ads-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+"agents": patch
+---
+
+Add pkg.pr.new workflow for publishing agents package
```
threepointone: Don't need a changeset, doesn't affect agents functionally

threepointone:
Sorry, one last thing. Can you add `pkg-pr-new` to devDependencies in the root package.json? Otherwise it installs it every time.

PR #401, opened 2025-08-08
fixes #321: Queue messages if onConnect has not completed

### LINKED ISSUE #321: Race condition: onMessage() can be called before onConnect()
There is a subtle but nasty race condition bug in the Agent base class. onConnect() is the only place where you get connection context to do things like serialize session or identity information as an attachment (so it survives hibernation) to the connection for later use in access control. However, it's possible for onMessage() to be called BEFORE onConnect(). I have worked around it for now by queuing up messages and only processing the queue at the end of my onConnect() method but it's a footgun that others will hit.

I'm prepared to investigate and submit a pull request if I can figure out how to fix it. This should be possible because at the very least I can move my message queuing approach to the Agent base class, but before I do that, I'm wondering if its a known issue and if anyone else has any thoughts.

Also, I'm wondering if I should fix it in the Agent source code and if so, where is the source of truth for that? Alternatively, maybe it's a problem with PartyKit and I should fix it there?

kchro3:
Attempt at resolving this issue: https://github.com/cloudflare/agents/issues/321

### DIFF
--- a/packages/agents/src/index.ts
+++ b/packages/agents/src/index.ts
@@ -290,6 +290,10 @@ export class Agent<Env = typeof env, State = unknown> extends Server<Env> {
   private _ParentClass: typeof Agent<Env, State> =
     Object.getPrototypeOf(this).constructor;
 
+  // Connection state management for race condition fix
+  private _connectionStates = new Map<string, 'connecting' | 'ready'>();
+  private _messageQueues = new Map<string, Array<{connection: Connection, message: WSMessage}>>();
+
   mcp: MCPClientManager = new MCPClientManager(this._ParentClass.name, "0.0.1");
 
   /**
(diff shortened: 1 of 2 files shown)


whoiskatrin:
@kchro3 do you mind taking a look at the build and fixing it?

kchro3:
I don't think I was actually running into this issue in the end. 

I was trying to fix an issue where I needed to reconnect & reauthenticate my MCP connections after the DO woke up from hibernating, but I had assumed that onConnect would happen again before onMessage. It turns out that the websocket connection is still live, so queuing the messages wouldn't have fixed that issue for me. Instead, I had to store the auth credentials using serializeAttachment on the first onConnect, and in future onMessages, if the MCPs were disconnected, it would deserializeAttachment and try again.

Since making that fix, I haven't run into the issue where onMessage happened before onConnect. Still happy to tinker with it on the weekend if folks think this is a good idea, though.

whoiskatrin:
@kchro3 up to you, i will check the original issue this weekend i think

whoiskatrin:
closing this as the bug has been fixed and fix has been merged into the last version now

PR #409, opened 2025-08-14
Refactor message types to use enum in AIChatAgent and related files

MrgSub:
- Updated message type references in ai-chat-agent.ts, ai-react.tsx, and other files to use the new MessageType enum.
- Removed unused imports and ensured consistent usage of message types across the codebase.

### DIFF
--- a/packages/agents/src/ai-types.ts
+++ b/packages/agents/src/ai-types.ts
@@ -17,7 +17,7 @@ export enum MessageType {
 
   CF_AGENT_MCP_SERVERS = "cf_agent_mcp_servers",
   CF_AGENT_STATE = "cf_agent_state",
-  RPC = "rpc",
+  RPC = "rpc"
 }
 
 /**


packages/agents/src/ai-chat-agent.ts
```diff
@@ -245,23 +259,16 @@ export class AIChatAgent<Env = unknown, State = unknown> extends Agent<
   private async _reply(id: string, response: Response) {
     // now take chunks out from dataStreamResponse and send them to the client
     return this._tryCatchChat(async () => {
-      // @ts-expect-error TODO: fix this type error
-      for await (const chunk of response.body!) {
-        const body = decoder.decode(chunk);
-
-        this._broadcastChatMessage({
```
nickfujita, line 252: why was this broadcast of the text chunks to the client removed? `_drainStream` seems to perform the chunk decode, but never broadcasts the value.

Issue #396, opened 2025-08-06
Why for persisting new messages all previous messages are deleting?


dhnchandan:
In AIChatAgent / persistMessages, before inserting new messages we are deleting all existing messages. Then inserting all messages again. Why we are doing this? It increases SQL read and write unnecessarily. We can insert broadcast the last message only.

<img width="1566" height="948" alt="Image" src="https://github.com/user-attachments/assets/7d8a0de0-6129-4247-92d1-b19ec36cbcd7" />

whoiskatrin:
Released

PR #411, opened 2025-08-15
feat(perf): reduce number of sql queries

amorriscode:
This PR resolves https://github.com/cloudflare/agents/issues/396.

Technically only the last message in the `messages` array needs to be written to the DB but for safety reasons I've adjusted the code to persist all messages that don't exist. I'm happy to update the code so it just grabs the last message in `messages` and store that instead if preferred!

Tested this by running in the playground.

### DIFF
--- a/.changeset/gold-states-kick.md
+++ b/.changeset/gold-states-kick.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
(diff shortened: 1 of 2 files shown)


amorriscode:
Ah turns out that this is not relevant once https://github.com/cloudflare/agents/pull/382 goes in. 🙂

PR #415, opened 2025-08-18
Make McpAgent extend Agent + Streaming HTTP protocol features

deathbyknowledge:
Lib changes:
- `McpAgent` now extends the base `Agent` class #235
- Refactor to only use `Agent` and Partykit `Server` primitives. Depends on https://github.com/cloudflare/partykit/pull/268 
- Clients can now GET the MCP endpoint to establish a [standalone SSE stream](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#listening-for-messages-from-the-server) for server initiated messages
- Clients can now DELETE the MCP endpoint to terminate a session #416
- Tool notifications are now also handled in the Streamable HTTP transport

Tests:
- New [tests](https://github.com/cloudflare/agents/pull/415/files#diff-1a8b07500402ffaae0b07a3d8c4fb7575952884dee1d32234be72e981c610dcfR354) to check the updated Streamable HTTP bits
- Added [E2E tests](https://github.com/cloudflare/agents/pull/415/files#diff-faab067d57deced7af1c87360cee48224e2932538890ead95af28e2580632ee9R24) that **deploy** (using [alchemy](https://github.com/sam-goodwin/alchemy)) complete MCP servers and test expected behavior with `@modelcontextprotocol/sdk`. They include authless and OAuth MCP examples.

Streamable HTTP changes adapted from the [TS SDK](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/server/streamableHttp.ts)

Run the E2E from `/packages/agents` with 
```sh
npm run test:e2e
```

PR #426, opened 2025-08-28
MCP: Support authInfo from Transport

### LINKED ISSUE #422: MCP: Support authInfo
# Issue: Request Support for authInfo from Transport in McpAgent

## Problem Description

Currently, the `McpAgent` class from `'agents/mcp'` does not provide access to
the `authInfo` from the underlying MCP transport layer. This prevents request
handlers from accessing authentication information that should be automatically
populated by the transport.

## Current Behavior

When setting request handlers in `McpAgent`, the `extra` parameter in request
handlers does not contain the `authInfo` property, even though the MCP SDK's
`RequestHandlerExtra` type defines it as:

```typescript
export type RequestHandlerExtra<
	SendRequestT extends Request,
	SendNotificationT extends Notification,
> = {
	signal: AbortSignal
	authInfo?: AuthInfo // This is undefined in our handlers
	sessionId?: string
	_meta?: RequestMeta
	requestId: RequestId
	requestInfo?: RequestInfo
	sendNotification: (notification: SendNotificationT) => Promise<void>
	sendRequest: <U extends ZodType<object>>(
		request: SendRequestT,
		resultSchema: U,
		options?: RequestOptions,
	) => Promise<z.infer<U>>
}
```

## Expected Behavior

Request handlers should be able to access `authInfo` from the `extra` parameter
when the transport has validated an access token:

```typescript
this.server.server.setRequestHandler(
	SetLevelRequestSchema,
	async (request, extra) => {
		// extra.authInfo should contain validated token information
		const authInfo = extra.authInfo
		if (authInfo) {
			console.log('Authenticated user:', authInfo.clientId)
			console.log('Token scopes:', authInfo.scopes)
		}
		// ... rest of handler logic
	},
)
```

## Impact

Without access to `authInfo` from the transport:

1. **Request handlers cannot access authentication context** - they must rely on
   agent-level props
2. **Inconsistent authentication patterns** - some handlers use
   `this.props.authInfo`, others can't access transport auth
3. **Limited OAuth integration** - can't leverage the transport's built-in token
   validation
4. **Type safety issues** - the `extra.authInfo` property exists in types but is
   always undefined

## Proposed Solution

The `McpAgent` class should be updated to allow me to populate the `authInfo` property (or populate it for me). This would probably require a callback for resolving the authorization header to `authInfo`.

## Workaround

Currently, we must access authentication information through
`this.props.authInfo` in request handlers, which works but doesn't leverage the
transport's built-in authentication capabilities.

kentcdodds:
Closes #422

## Problem Description

Currently, the `McpAgent` class from `'agents/mcp'` does not provide access to the `authInfo` from the underlying MCP transport layer. This prevents request handlers from accessing authentication information that should be automatically populated by the transport.

## Current Behavior

When setting request handlers in `McpAgent`, the `extra` parameter in request handlers does not contain the `authInfo` property, even though the MCP SDK's `RequestHandlerExtra` type defines it as:

```typescript
export type RequestHandlerExtra<
  SendRequestT extends Request,
  SendNotificationT extends Notification
> = {
  signal: AbortSignal;
  authInfo?: AuthInfo; // This is undefined in our handlers
  sessionId?: string;
  _meta?: RequestMeta;
  requestId: RequestId;
  requestInfo?: RequestInfo;
  sendNotification: (notification: SendNotificationT) => Promise<void>;
  sendRequest: <U extends ZodType<object>>(
    request: SendRequestT,
    resultSchema: U,
    options?: RequestOptions
  ) => Promise<z.infer<U>>;
};
```

## Expected Behavior

Request handlers should be able to access `authInfo` from the `extra` parameter when the transport has validated an access token:

```typescript
this.server.server.setRequestHandler(
  SetLevelRequestSchema,
  async (request, extra) => {
    // extra.authInfo should contain validated token information
    const authInfo = extra.authInfo;
    if (authInfo) {
      console.log("Authenticated user:", authInfo.clientId);
      console.log("Token scopes:", authInfo.scopes);
    }
    // ... rest of handler logic
  }
);
```

## Solution

This PR implements the following changes to support `authInfo` from the transport:

### 1. Enhanced Transport Interface

- Updated `McpSSETransport` and `McpStreamableHttpTransport` to support `MessageExtraInfo` in their `onmessage` callbacks
- Added support for passing `requestInfo` and `authInfo` through the transport layer

### 2. Request Context Tracking

- Added `_currentRequestInfo` and `_currentAuthInfo` private properties to track the current request context
- Implemented `setCurrentRequestContext()` method to capture request headers and metadata
- Added `setCurrentAuthInfo()` method to validate and store authentication information

### 3. Auth Resolver Callback

- Introduced `setAuthResolver()` method that allows users to provide a custom authentication validation function
- The resolver receives the incoming `Request` and returns `AuthInfo | undefined`
- This enables flexible authentication strategies (OAuth, JWT, API keys, etc.)

### 4. Integration Points

- Updated SSE and Streamable HTTP transport initialization to capture request context
- Modified message forwarding to include `requestInfo` and `authInfo` in transport callbacks
- Ensured both transport types receive the enhanced message context

### 5. Testing

- Added comprehensive tests for request info tracking
- Added tests for authentication info handling with and without authorization headers
- Verified that tool handlers receive the expected `extra` parameters

## Usage Example

```typescript
class MyMcpAgent extends McpAgent<Env, State, Props> {
  async init() {
    // Set up custom auth resolver
    this.setAuthResolver(async (request: Request) => {
      const authHeader = request.headers.get("authorization");
      if (!authHeader?.startsWith("Bearer ")) return undefined;

      const token = authHeader.substring(7);
      // Validate token and return auth info
      return {
        token,
        clientId: "user-123",
        scopes: ["read", "write"]
      };
    });

    // Tool handlers now have access to authInfo
    this.server.tool(
      "protectedAction",
      "A protected tool that requires authentication",
      {},
      async (params, extra) => {
        if (!extra.authInfo) {
          throw new Error("Authentication required");
        }

        console.log(`User ${extra.authInfo.clientId} performing action`);
        // ... tool logic
      }
    );
  }
}
```

## Impact

With this implementation:

1. **Request handlers can access authentication context** - they now receive `authInfo` from the transport
2. **Consistent authentication patterns** - all handlers can use the same `extra.authInfo` access pattern
3. **Enhanced OAuth integration** - leverages the transport's built-in token validation capabilities
4. **Type safety** - the `extra.authInfo` property now contains the expected data instead of being undefined
5. **Flexible auth strategies** - users can implement custom authentication logic through the resolver callback

## Breaking Changes

None. This is a purely additive change that enhances existing functionality without modifying existing APIs.

## Testing

- ✅ All existing tests pass
- ✅ New tests verify request info tracking functionality
- ✅ New tests verify authentication info handling
- ✅ Both SSE and Streamable HTTP transports are covered
- ✅ Edge cases (no auth header, invalid tokens) are handled gracefully


kentcdodds:
Playing around with this more and I just realized that I also need access to the extra info (specifically authInfo, but requestInfo too) in the McpAgent's `init` function. I need that, so I'm looking into how I can make that work as well.

kentcdodds:
I changed the implementation quite a bit and verified it works in my project.

kentcdodds:
I've also verified this works in production as well.

whoiskatrin:
@deathbyknowledge can you please take a look? thank you

whoiskatrin:
@kentcdodds do you want to add an .md in the docs or maybe mention this in the readme?

kentcdodds:
Sure thing. I added a few docs to this PR. The agent wanted to add a bunch more, but I couldn't 100% verify the accuracy of some of what it said and I don't want this PR to get tied up by those changes.

deathbyknowledge:
Hey @kentcdodds, we currently have this PR #415 that refactors `McpAgent` to natively extend the `Agent` class and might simplify some of these issues. Mind clarifying why you need headers/auth in the callback?  I might be able to address them there too

Is `this.props` insufficient to handle `init()` and your tool callbacks? What are you missing here?
I've read from a few examples that are setting feature flags through request headers to gate tools. This could be an example of how to implement that in conjunction with `props`, to make request-level info available in your props:

```ts
// In your worker fetch handler
fetch(request: Request, env: Env, ctx: ExecutionContext) {
    // Handle custom props
    const props: Props = {
      echoAvailable: !!request.headers.get("x-my-feature-flag")
    };
    
    // If we're not using OAuthProvider we can set our custom props directly
    // otherwise we'd need to make sure we don't override the already existing props
    ctx.props = props;
    return MyMCP.serve("/").fetch(request, env, ctx);
}

// In your McpAgent
async init() {
    const { echoAvailable } = this.props;
    // Simple echo tool that's gated behind a feature flag
    if (echoAvailable) {
      this.server.tool("echo", { msg: z.string() }, async ({ msg }) => ({
        content: [{ type: "text", text: msg }]
      }));
    }
}
```

If the McpAgent is behind an `OAuthProvider` we'd need to make sure we're only appending to the props in the `apiHandler` (once the provider has already set the props), otherwise they could get overridden.
Of course, you could also set any custom props for your session at the `/authorization` step.

If I'm not mistaken, if you're not using OAuthProvider, you can completely customize your prop handling and provide them to your MCP directly with `MyMCP.serve("/").fetch(request, env, ctx)`, where `ctx` includes your custom props, as the example above.

I've updated the E2E in the PR I mentioned so they include some of these examples ([authless](https://github.com/cloudflare/agents/pull/415/files#diff-7089b25ed05e440c647c49c2f28ed88e4b84db9bdeb878fa809bc2a60f973bc1R70-R80), [OAuth](https://github.com/cloudflare/agents/pull/415/files#diff-9a3ab44f2d8b2eeae9b5ffb29ae4c3261b4a7916c387f9edc54299a6a550659dR68-R82)). Let me know if I missed anything

kentcdodds:
Modifying props is actually how I handle this now as noted in #422, but I was told this is not intended to be modified so I decided to add this feature. However, if what I'm doing is fine then that's great!

However, we do still need a way to pass the `extra` argument to the transport `onmessage` callback to be spec compliant, so I think this PR still has value for that case.

kentcdodds:
Merged with main. I'd still love to get support for the `extra` argument 🙏

kentcdodds:
Anything I can do to get progress on merging this?

deathbyknowledge:
> Anything I can do to get progress on merging this?

Hey @kentcdodds, we're currently in the process of merging #415 which conflicts with these changes. Also, I believe the `extra` argument is not part of the spec but only part of the MCP TS SDK implementation.

Regardless, we're planning to re-think/improve the ergonomics around `ctx.props` which should solve the pain points you mention (afaik there are already valid workarounds), can't say when exactly that will happen though

kentcdodds:
I see. Good to know. I'm going to be recording videos showing people how to use this package and I would really love to know how things will work so I can do that in my videos (even if it doesn't work that way currently, I can make it appear that way in the videos). Can you give me an example of how I could get the auth info into my tool handler and also how I could get the `new URL(request.url).origin` into the tool handler. Right now I'm using this PR, but if this is not how that's going to work then I don't want to do that.

deathbyknowledge:
If you're using `workers-oauth-provider` to setup OAuth in your Worker, there are a few things to keep in mind. Take this example:

```typescript
// Consider this Worker with the following entrypoint
export default new OAuthProvider({
  // These API handlers ONLY run after authorization.
  // This means ctx.props has already been populated by OAuthProvider
  apiHandlers: {
    // Any other API handlers you'd want to add would go here...
    "/mcp": {
      fetch(request: Request, env: unknown, ctx: ExecutionContext) {
        // This handler is running in your Worker's context, not the DO.
        // You are safe to inject any props here as you see fit, since the
        // OAuthProvider has already set its own, meaning yours won't get overwritten.
        ctx.props = {
          ...ctx.props,
          origin: new URL(request.url).origin,
          // You could also read the headers to set custom props
          flag: !!request.headers.get("x-my-feature-flag"),
        };
        
        // The `serve` handler will always ensure that the latest props are available
        // in your MCP through `this.props`
        return MyMCP.serve("/mcp").fetch(request, env, ctx);
      }
    }
  },
  defaultHandler: myHandler, // This one runs before authorization! 
  authorizeEndpoint: "/authorize", // In fact, `/authorize` lives in the defaultHandler too!
  tokenEndpoint: "/token",
  clientRegistrationEndpoint: "/register",
});
```

This would be for "request-level" props in case you had any. If you needed session-wide props, you can set them in your `/authorize` logic.

```typescript
// Following the previous snippet, this could be its defaultHandler
const myHandler = {
  // This handles all requests for your Worker that are:
  // - non-OAuth related. Any authless routing your app might have
  // - OR the `/authorization` step for OAuth, which you get to write here
  async fetch(request: Request, env: Env) {
    const url = new URL(request.url);
    const provider = env.OAUTH_PROVIDER;
    
    // Health check for the Worker (which is authless)
    if (url.pathname === "/health") {
      return new Response("OK");
    }
    
    // Anything else you might want to expose without auth...
    
    // Your `/authorize` handler for OAuth
    if (new URL(request.url).pathname === "/authorize") {
      // Your authorization step logic
      // ...
      const { email, username } = myAuthorizedPayloadOrSomething;
      // Let's pretend you have access to `email` and `username` after
      // this step and you want to make them available in the props
      
      // Now this props will be what OAuthProvider sets in `ctx.props` and
      // makes available to every authorized request from this user. You should
      // be able to get the same data here as from `authInfo` but the specific
      // implementation depends on how you rollout your auth.
      const props = { email, username }; 
      const { redirectTo } = await provider.completeAuthorization({
        // your other auth args...
        props, // we give OAuthProvider the props to use from now on
      });
      
      return Response.redirect(redirectTo, 302);
    }
  }
};
```
Mind that the `/authorize` step is not implemented in the snippet. For the usual providers and how to rollout your own, have a look at the `remove-mcp-*` demos [here](https://github.com/deathbyknowledge/ai/tree/main/demos).

Finally, you would use the props we just set in your `McpAgent` like so:
```typescript
// The type for the props we set previously
type MyProps = {
  flag: boolean; // Let's use this flag to gate a tool
  origin: string;
  username: string;
  email: string;
}

// The type param McpAgent<..., ..., Props> will land with the PR I mentioned previously
export class MyMCP extends McpAgent<Env, State, MyProps> {
  server = new McpServer({ name: "my-mcp", version: "0.1.0" });

  async init() {
    // We gate the `whoami` tool if the request included the feature flag header.
    // This means that the `ctx.props.flag` was set at init time of the MCP
    if (this.props?.flag) {
      this.server.tool(
        "whoami",
        "Return the email and username of the user",
        {},
        async () => ({
          // This tool will read email and username from `this.props` every time it's called
          content: [{ type: "text", text: `User: ${this.props?.username} (${this.props?.email})` }]
        })
      );
    }
    
    this.server.tool(
      "getOrigin",
      "Return the request origin",
      {},
      async () => ({
        content: [{ type: "text", text: this.props?.origin }]
      })
    );
  }
}
```
It's a bit long but I think that covers every step of the request (except OAuth internals)! Let me know if that works :)

PS: If you were NOT using OAuth, you can follow the same pattern and provide your own props to `MyMCP.serve....`. You get to skip the defaultHandler/apiHandlers shenanigans.

PR #427, opened 2025-08-28
feat: rename server property to mcp in MCP-related classes for improved developer experience

### LINKED ISSUE #425: MCP: Deal with the server.server echo?
When we need to access the underlying server implementation this is what we have to do:

```typescript
import { McpAgent } from "agents/mcp";
import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";

export class MyMCP extends McpAgent {
	server = new McpServer({/* ... */});

	async init() {
		// server.server echo is annoying and confusing
		this.server.server.setRequestHandler(
			SetLevelRequestSchema,
			async (request, extra) => {
				// ... handler logic
			},
		);
	}
}
```

It's kind of annoying (and it's the MCP SKD's fault). I would rather it be `this.mcp.server.setRequestHandler`:

```typescript
import { McpAgent } from "agents/mcp";
import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";

export class MyMCP extends McpAgent {
	mcp = new McpServer({/* ... */});

	async init() {
		this.mcp.server.setRequestHandler(
			SetLevelRequestSchema,
			async (request, extra) => {
				// ... handler logic
			},
		);
	}
}
```

However, `McpAgent` requires the server to be set to the `server` property.

This is just a small quality of life thing, but I think it should be easy enough to un-document the `server` name, document the `mcp` name, and support both.

kentcdodds:
Technically there are some breaking changes here, but they're either done in a non-breaking way or they're breaking features I do not think are used (and are not documented).

Closes #425

- **BREAKING**: Renamed `server` property to `mcp` in `McpAgent` and updated related examples.
- **BREAKING**: Renamed `mcp` property to `mcpClientManager` in `Agent` class to avoid naming conflicts.
- Added backward compatibility for `server` property in `McpAgent` with a deprecation warning.
- Updated all MCP examples to reflect the new naming convention.
- Improved property naming consistency across the MCP implementation.

### DIFF
--- a/.changeset/mcp-property-rename.md
+++ b/.changeset/mcp-property-rename.md
@@ -1,0 +1,31 @@
+---
+"@cloudflare/agents": patch
+---
+
+Rename `server` property to `mcp` in MCP-related classes for better developer experience
+
+## Changes
+
+- **BREAKING**: Renamed `server` property to `mcp` in `McpAgent` class and related examples
+- **BREAKING**: Renamed `mcp` property to `mcpClientManager` in `Agent` class to avoid naming conflicts
+- Added backward compatibility support for `server` property in `McpAgent` with deprecation warning
+- Updated all MCP examples to use the new `mcp` property naming convention
+- Improved property naming consistency across the MCP implementation
+
+## Migration
+
+If you're using the `server` property in your `McpAgent` implementations, update your code:
+
+```ts
+// Before
+export class MyMcpAgent extends McpAgent {
+  server = new McpServer({...});
+}
+
+// After
+export class MyMcpAgent extends McpAgent {
+  mcp = new McpServer({...});
+}
+```
+
+The `server` property is still supported for backward compatibility but will be removed in a future version.
--- a/examples/mcp-elicitation-demo/src/server.ts
+++ b/examples/mcp-elicitation-demo/src/server.ts
@@ -17,7 +17,7 @@ type Env = {
 };
 
 export class McpServerAgent extends McpAgent<Env, { counter: number }, {}> {
-  server = new McpServer({
+  mcp = new McpServer({
     name: "Elicitation Demo Server",
     version: "1.0.0"
   });
@@ -83,7 +83,7 @@ export class McpServerAgent extends McpAgent<Env, { counter: number }, {}> {
 
   async init() {
     // Counter tool with user confirmation via elicitation
-    this.server.tool(
+    this.mcp.tool(
       "increment-counter",
       "Increment the counter with user confirmation",
       {
@@ -151,7 +151,7 @@ export class McpServerAgent extends McpAgent<Env, { counter: number }, {}> {
     );
 
     // User creation tool with form-based elicitation
-    this.server.tool(
+    this.mcp.tool(
       "create-user",
       "Create a new user with form input",
       {
@@ -225,7 +225,7 @@ export class McpServerAgent extends McpAgent<Env, { counter: number }, {}> {
     );
 
     // Counter resource
-    this.server.resource("counter", "mcp://resource/counter", (uri: URL) => {
+    this.mcp.resource("counter", "mcp://resource/counter", (uri: URL) => {
       return {
         contents: [
           {
--- a/examples/mcp/src/server.ts
+++ b/examples/mcp/src/server.ts
@@ -9,7 +9,7 @@ type Env = {
 type State = { counter: number };
 
 export class MyMCP extends McpAgent<Env, State, {}> {
-  server = new McpServer({
+  mcp = new McpServer({
     name: "Demo",
     version: "1.0.0"
   });
@@ -19,13 +19,13 @@ export class MyMCP extends McpAgent<Env, State, {}> {
   };
 
   async init() {
-    this.server.resource("counter", "mcp://resource/counter", (uri) => {
+    this.mcp.resource("counter", "mcp://resource/counter", (uri) => {
       return {
         contents: [{ text: String(this.state.counter), uri: uri.href }]
       };
     });
 
-    this.server.tool(
+    this.mcp.tool(
       "add",
       "Add to the counter, stored in the MCP",
       { a: z.number() },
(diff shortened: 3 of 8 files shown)


packages/agents/src/index.ts
```diff
@@ -291,7 +291,10 @@ export class Agent<Env = typeof env, State = unknown> extends Server<Env> {
   private _ParentClass: typeof Agent<Env, State> =
     Object.getPrototypeOf(this).constructor;
 
-  mcp: MCPClientManager = new MCPClientManager(this._ParentClass.name, "0.0.1");
+  mcpClientManager: MCPClientManager = new MCPClientManager(
+    this._ParentClass.name,
+    "0.0.1"
+  );
```
kentcdodds: In retrospect, I do not think this change is necessary and I'm happy to back it out. This is the most breaking thing about this PR. Though again I don't think anyone is intended to use this property (it's also undocumented).

packages/agents/src/mcp/index.ts
```diff
@@ -210,8 +210,24 @@ export abstract class McpAgent<
    */
   private _agent: Agent<Env, State>;
 
-  get mcp() {
-    return this._agent.mcp;
+  get mcpClientManager() {
+    return this._agent.mcpClientManager;
+  }
```
kentcdodds, line 215: This bit is technically breaking for anyone who's using the client manager, but I don't think anyone is and it's not documented.

packages/agents/src/mcp/index.ts
```diff
@@ -350,8 +366,17 @@ export abstract class McpAgent<
...
+  server?: MaybePromise<McpServer | Server>;
+  /**
+   * This is only set as optional for backward compatibility in case you're
+   * using the legacy `server` property. It's recommended to use `mcp`.
+   *
+   * In the future, we'll make this a required abstract property to implement
+   * and remove the `server` property.
    */
-  abstract server: MaybePromise<McpServer | Server>;
+  mcp?: MaybePromise<McpServer | Server>;
```
kentcdodds, line 379: This is the primary focus of this PR

examples/mcp/src/server.ts
```diff
@@ -9,7 +9,7 @@ type Env = {
 type State = { counter: number };
 
 export class MyMCP extends McpAgent<Env, State, {}> {
-  server = new McpServer({
+  mcp = new McpServer({
```
cliffhall, line 12: Maybe `mcp` could be a getter that returns `server`? That way it would be totally backward compatible.

kentcdodds: I don't think that would allow me to do this:

```ts
export class MyMCP extends McpAgent {
	mcp = new McpServer({}); // <-- this is the desired API

	async init() {
		this.mcp.server.setRequestHandler(
			SetLevelRequestSchema,
			async (request, extra) => {
				// ... handler logic
			},
		);
	}
}
```

cliffhall: Something like this?

```typescript
export class MyMCP extends McpAgent {

	get mcp(): McpServer {
		return this.server;
	}
        
	server = new McpServer({});

	async init() {
		this.mcp.server.setRequestHandler(
			SetLevelRequestSchema,
			async (request, extra) => {
				// ... handler logic
			},
		);
	}
}
```

kentcdodds: I could do that myself without having to change anything in `agents`. That's actually a reasonable workaround if they decide this is more trouble than it's worth. Thanks for the idea!

kentcdodds: Oh, actually I can't do that because `mcp` is already set to the `mcpClientManager` 🙁

.changeset/mcp-property-rename.md
```diff
@@ -0,0 +1,31 @@
+---
+"@cloudflare/agents": major
```
kentcdodds, line 2: Changed to major because there was an example of using the previous existing `mcp` property so people may actually be using that.

examples/mcp-elicitation-demo/src/server.ts
```diff
@@ -388,7 +388,7 @@ export class MyAgent extends Agent<Env, never> {
         __clientSession: this.name
       };
 
-      const result = await this.mcp.callTool({
+      const result = await this.mcpClientManager.callTool({
```
kentcdodds, line 391: Here's the breaking change.

whoiskatrin:
FYI I'm not against this, I think the change makes life better. My concern it's a breaking change, if we do so, I'd like to make a bigger release and then also prepare people for this migration.

kentcdodds:
I'm fine waiting 👍

packages/agents/src/mcp/index.ts
mattzcarey: Very happy with this change. I think this is super confusing at the moment. 

We should keep the `mcp()` and mark it as deprecated to be removed in the next version.

packages/agents/src/mcp/index.ts
```diff
@@ -210,8 +210,24 @@ export abstract class McpAgent<
...
+   */
+  get mcpServer(): MaybePromise<McpServer | Server> {
+    const server = this.mcp ?? this.server;
+    if (!server) {
+      throw new Error(
+        'Neither the `mcp` nor `server` property is set. Please set "mcp".'
+      );
+    }
+
+    return server;
```
mattzcarey, line 230: I think this adds to the confusion right now. 

IMO mcpServer should be the new name for `server()`. Thereby we can have a deprecation notice on the `mcp()` naming and keep everything clean. Much happier for more descriptive names especially as this class gets bigger.

packages/agents/src/mcp/index.ts
mattzcarey: Would you be open to this changing to mcpServer() ?

threepointone:
but that still means writing mcpServer.server

mattzcarey:
> but that still means writing mcpServer.server

i hate that much less. You also only need to write that when accessing the underlying sdk `server` which makes sense?

threepointone:
I'm going to close this, I think we'll need to revisit it in the future. Thanks for the PR Kent.

PR #435, opened 2025-09-04
add support for custom message types in AIChatAgent with generic type param

whoiskatrin:
- updated AIChatAgent<Env, State, Message> class signature
- updated messages prop and all related methods to use generic type
- added basic docs and examples


bekicot:
Hi, I came across this as my App only want to store it as ModelMessage, instead of UIMessage. This will simplify a lot of conversion in the frontend and the LLM calls.

is there any reason why this PR closed?

PR #436, opened 2025-09-04
Fix onConnect race condition

### LINKED ISSUE #321: Race condition: onMessage() can be called before onConnect()

deathbyknowledge:
Fixes #321 and adds test for message ordering.

### DIFF
--- a/packages/agents/src/tests/worker.ts
+++ b/packages/agents/src/tests/worker.ts
@@ -129,14 +129,10 @@ export default {
       return new Response("Internal Server Error", { status: 500 });
     }
 
-    if (url.pathname.startsWith("/agents/")) {
-      return (
-        (await routeAgentRequest(request, env)) ||
-        new Response("Not found", { status: 404 })
-      );
-    }
-
-    return new Response("Not found", { status: 404 });
+    return (
+      (await routeAgentRequest(request, env)) ||
+      new Response("Not found", { status: 404 })
+    );
   },
 
   async email(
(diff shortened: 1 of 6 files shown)


whoiskatrin:
tentatively looks ok to me

packages/agents/src/tests/worker.ts
```diff
@@ -105,6 +129,13 @@ export default {
       return new Response("Internal Server Error", { status: 500 });
     }
 
+    if (url.pathname.startsWith("/agents/")) {
+      return (
+        (await routeAgentRequest(request, env)) ||
+        new Response("Not found", { status: 404 })
+      );
+    }
+
```
threepointone:
```suggestion    
    return (
      (await routeAgentRequest(request, env)) ||
      new Response("Not found", { status: 404 })
    );    

```

PR #440, opened 2025-09-07
fix: pass agent._pk as id to useChat to prevent stale WebSocket instances

### LINKED ISSUE #439: Stale WebSocket Instances Used in useAgentChat
# Multiple WebSocket Instances Created in useAgent

## Description

The [useAgent](https://github.com/cloudflare/agents/blob/8dac62c6f6c513d7fd481eb3b519b533bac17f1f/packages/agents/src/react.tsx#L173) hook is creating multiple different WebSocket instances. When using useAgentChat, the Vercel AI SDK's [useChat](https://github.com/vercel/ai/blob/651860ebefb22b313ed89b3e23148d63d13226e2/packages/react/src/use-chat.ts#L61) hook uses `useRef` to store transport instances, which causes useAgentChat to use an outdated WebSocket instance.

For example, when creating a socket like this:

```js
const socket = usePartySocket({
  party: 'chat',
  host: 'ws://localhost:5173',
  room: 'jkXfNgumpoevO9njNjqlz',
  prefix: 'agents'
})

useEffect(() => {
  console.log('socket', socket._pk)
}, [socket])
```

Multiple calls to `useAgent` with the same parameters will create different WebSocket instances with different `_pk` values.

Image

## Solution

Pass the agent's unique identifier (`agent._pk`) as the `id` parameter to the [useChat](https://github.com/cloudflare/agents/blob/8dac62c6f6c513d7fd481eb3b519b533bac17f1f/packages/agents/src/ai-react.tsx#L326) hook. This ensures that when useAgent creates multiple WebSocket instances, useAgentChat will get the updated connection associated with the current agent instance.

axuj:
Pass the agent's unique identifier (`agent._pk`) as the `id` parameter to the useChat hook. This ensures that when useAgent creates multiple WebSocket instances, useAgentChat will get the updated connection associated with the current agent instance.

Fixes #439

### DIFF
--- a/packages/agents/src/ai-react.tsx
+++ b/packages/agents/src/ai-react.tsx
@@ -326,7 +326,8 @@ export function useAgentChat<
   const useChatHelpers = useChat<ChatMessage>({
     ...rest,
     messages: initialMessages,
-    transport: customTransport
+    transport: customTransport,
+    id: agent._pk
   });
 
   const processedToolCalls = useRef(new Set<string>());


threepointone (changes requested):
soft blocking so I can review

packages/agents/src/ai-react.tsx
```diff
@@ -326,7 +326,8 @@ export function useAgentChat<
   const useChatHelpers = useChat<ChatMessage>({
     ...rest,
     messages: initialMessages,
-    transport: customTransport
+    transport: customTransport,
+    id: agent._pk
```
threepointone, line 349: a problem with this is that _pk is an identifier for the connection, but "id" passed to useChat is for the conversation 

it's actually not clear to me how this actually solves the problem of stale websocket instances that are being generated by useAgent 

I'll keep looking at this, but the recommended path for now is to not use StrictMode

axuj: As explained in https://github.com/cloudflare/agents/pull/465#issuecomment-3294798782, the useChat hook of the Vercel AI SDK uses useRef to store the Chat instance. The Chat instance saved in useRef will only be updated when option.id or option.chat changes.
_pk is a random identifier generated along with the WebSocket (ws) connection. A change in _pk indicates that the WebSocket has changed. When _pk is passed as the id to useChat, it enables the reinitialization of the Chat instance, allowing the use of the new WebSocket connection.
It is unclear why StrictMode causes two WebSocket instances to be generated. However, even without Strict Mode, it is assumed that if the WebSocket connection is disconnected and reconnected, both the WebSocket instance and _pk will change—making it still necessary to have useChat update the Chat instance.

threepointone: ah ok understood

PR #447, opened 2025-09-09
add support for plain text responses alongside SSE streaming 

### LINKED ISSUE #441: using `generateText` inside `onChatMessage` returns no response [aisdkv5]
## Problem
Im trying to just use `generateText` instead of `streamText` inside `onChatMessage`, but it seems like the code for ai-chat-agent is not compatible with it?

This is my code in the agents DO:
```ts
const textResult = await generateText({
  model: defaultModel,
  system: prompt,
  messages: convertToModelMessages(this.messages),
});

await onFinish(textResult);

return new Response(textResult.text, {
  headers: { 'Content-Type': 'text/plain' },
});
```

the `fullResponseText` in `_reply` is empty, because of this line here:

```ts
const lines = chunk.split('\n');
for (const line of lines) {
  if (line.startsWith('data: ') && line !== 'data: [DONE]') {
  ...
}
```

`lines` is an array of strings in the case of `generateText` - something like this (not sure what it will look like with tool calls, as I don't have those in my codebase):

```
lines [
  '',
  '',
  "Yo! Keep that energy coming!"
]
```

## Solution
```ts
for (const line of lines) {
  if (line.startsWith('data: ') && line !== 'data: [DONE]') {
    try {
        ...
    } catch (_e) {
        ...
    }
  // in case of 'generateText`, its just a string with the text
  } else if (line.length > 0) {
    fullResponseText += line;
  }
}
```

whoiskatrin:
this_ fixes #441 

added Content-Type detection to tell SSE (text/event-stream) from plain text. For plain text responses like generateText, we now accumulate chunks directly into fullResponseText. also we synthesize text-delta events so client streaming stays compatible and persist the complete response as message parts

### DIFF
--- a/.changeset/three-lions-fail.md
+++ b/.changeset/three-lions-fail.md
@@ -2,4 +2,4 @@
 "agents": patch
 ---
 
-add support for plain text responses alongside SSE streaming 
+add support for plain text responses alongside SSE streaming


jcheese1:
@whoiskatrin do you know if theres a way to append metadata from the server, specifically `onChatMessage`?

PR #457, opened 2025-09-12
Handle CORS requests for only defined agent routes

pilcrowonpaper:
Currently `routeAgentRequest()` resolves any `OPTIONS` request it receives, regardless of the request path. This PR updates `routeAgentRequest()` to handle preflight requests only for defined agent routes (`/agents/NAMESPACE/NAME` and its subroutes).

I also improved the JSDoc a bit.

### DIFF
--- a/packages/agents/src/index.ts
+++ b/packages/agents/src/index.ts
@@ -1627,24 +1627,25 @@ export async function routeAgentRequest<Env>(
     corsHeaders = {};
   }
 
-  let response = await routePartykitRequest(request, env as any, {
+  const response = await routePartykitRequest(request, env as any, {
     prefix: "agents",
     jurisdiction: options.jurisdiction,
     locationHint: options.locationHint,
     // Preflight request with `Upgrade` header field don't exist.
     onBeforeConnect: options.onBeforeConnect as any,
     onBeforeRequest: async (req, lobby) => {
+      let resolvedRequest = req;
       if (options.onBeforeRequest !== undefined) {
         const reqOrRes = await options.onBeforeRequest(req, lobby as any);
         if (reqOrRes instanceof Response) {
           return reqOrRes;
         }
         if (reqOrRes instanceof Request) {
-          req = reqOrRes;
+          resolvedRequest = reqOrRes;
         }
       }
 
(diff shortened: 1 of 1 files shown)


packages/agents/src/index.ts
```diff
@@ -1577,65 +1577,97 @@ export type AgentContext = DurableObjectState;
...
+    // options.cors satisfies HeadersInit.
+
+    corsEnabled = true;
+    corsHeaders = options.cors;
+  } else {
+    corsEnabled = false;
+    corsHeaders = {};
+  }
+
+  const response = await routePartykitRequest(request, env as any, {
```
pilcrowonpaper, line 1630: I can replace these `any` if needed but it's something that needs to be fixed in `partyserver` imo

pilcrowonpaper: https://github.com/cloudflare/partykit/pull/278

threepointone: let's put it back the way it was for now

packages/agents/src/index.ts
```diff
@@ -1577,65 +1577,97 @@ export type AgentContext = DurableObjectState;
...
-  if (request.method === "OPTIONS") {
-    if (corsHeaders) {
-      return new Response(null, {
-        headers: corsHeaders
-      });
-    }
-    console.warn(
-      "Received an OPTIONS request, but cors was not enabled. Pass `cors: true` or `cors: { ...custom cors headers }` to routeAgentRequest to enable CORS."
-    );
+      if (resolvedRequest.method === "OPTIONS") {
```
threepointone, line 1648: don't we want to do this check before calling onBeforeRequest?

pilcrowonpaper: As a user I would assume `options.onBeforeRequest()` of `routeAgentRequest()` will be called before anything else.

packages/agents/src/index.ts
pilcrowonpaper: As in use `Record` over `any`?

threepointone:
fixed in https://github.com/cloudflare/agents/pull/871

Issue #318, opened 2025-06-10
cross-domain authentication with websocket and useAgentChat hook


pruales:
We are connecting to our chat agent from a different domain but we're unable to connect to it trying to pass an auth token. I cant find any docs on how to authenticate with the agent when using websockets and the useAgentChat hook. Is the best option to pass a query parameter with the token? if so we tried adding a query parameter to the host option on the agent hook but the hook wasnt able to format the url correctly. Whats the best approach for this?

ghost:
Check out https://github.com/cloudflare/agents/issues/150. Passing the token in the query prop works for me, although I'm constantly running into token expiration issues. Hoping the hook gets updated to allow for an async function for the query prop so the hook can handle fetching fresh tokens for us.

```typescript
const _agent = useAgent({
    host: WORKER_URL,
    protocol: "wss",
    agent: AGENT_NAME,
    name: AGENT_INSTANCE,
    query: { token },
  });
```

whoiskatrin:
cross domain auth guide can be found in docs folder in this repo, let me know if there are any questions

PR #458, opened 2025-09-12
cross-domain authentication example (docs and some polish) WIP

whoiskatrin:
Add unified async authentication support to useAgent hook
The useAgent hook now automatically detects and handles both sync and async query patterns

This also addresses [#318 ](https://github.com/cloudflare/agents/issues/318)

### DIFF
--- a/examples/cross-domain/src/client.tsx
+++ b/examples/cross-domain/src/client.tsx
@@ -174,9 +174,6 @@ function AsyncAuthApp() {
           <p>
             <strong>Room:</strong> {agent.name}
           </p>
-          <p>
-            <strong>WebSocket Ready State:</strong> {agent.readyState}
-          </p>
         </div>
       </div>
     </div>


ghost:
Thanks @whoiskatrin. Do you have a recommended pattern for handling the connection when the token expires that doesn't involve remounting the parent component? If I'm reading this right, it looks like the pattern you've proposed is:
1. Set the token on mount
2. Reload the page to trigger remounting the parent component which will fire an effect that refreshes the token

Imo, the ideal expired token experience has no page reload. What about updating `useAgent`'s `query` prop in the `options` param to work with an async function as its currently typed?

whoiskatrin:
> Thanks @whoiskatrin. Do you have a recommended pattern for handling the connection when the token expires that doesn't involve remounting the parent component? If I'm reading this right, it looks like the pattern you've proposed is:
Set the token on mount
Reload the page to trigger remounting the parent component which will fire an effect that refreshes the token
Imo, the ideal expired token experience has no page reload. What about updating useAgent's query prop in the options param to work with an async function as its currently typed?


yes, tbh, you are right, we also had TODO to implement it, I might as well just do it

yusuf-eren:
Hey, I asked this yesterday on an issue and found this PR. Thanks Katrin! That helps a lot.
I have a question:

Instead of agent based authentication, can we use `onBeforeConnect` and `onBeforeRequest` functions in `routeAgentRequest` ?

My pseudo code:
```typescript
 onBeforeConnect: async (req) => {
        const queryParams = new URLSearchParams(req.url.split('?')?.[1]);
        const token = queryParams.get('token');
        if (!token) {
          return new Response('Unauthorized', { status: 401 });
        }

        const { data, error } = await getSupabaseClient(env).auth.getUser(
          token
        );
        if (error) {
          console.log('error', error);
          return new Response('Unauthorized', { status: 401 });
        }
      },
```

If so, how can I pass that decoded data to the routed Agent?

whoiskatrin:
> I have a question:
> If so, how can I pass that decoded data to the routed Agent?


yes, you actually can (something new i learned today), Request gets returned, you can modify the request that gets passed, you can add new data to request and the context will contain the Request, so you can read it

packages/agents/src/react.tsx
```diff
@@ -288,3 +412,5 @@ export function useAgent<State>(
 
   return agent;
 }
+
+// No additional exports needed - useAgent now handles everything internally
```
threepointone:
```suggestion
```

README.md
```diff
@@ -27,7 +27,8 @@ This project is actively evolving. Current focus areas:
 - Core agent framework with state management
 - Real-time WebSocket communication
 - HTTP endpoints and routing
-- React integration hooks
+- React integration hooks (`useAgent`, `useAsyncAgent`)
+- Async authentication and dynamic query support
```
threepointone:
```suggestion
```

examples/cross-domain/src/server.ts
```diff
@@ -2,9 +2,44 @@ import { Agent, type Connection, routeAgentRequest } from "agents";
 import { env } from "cloudflare:workers";
 
 export class MyAgent extends Agent {
-  onConnect(connection: Connection) {
-    console.log("Client connected:", connection.id);
-    connection.send(`Welcome! You are connected with ID: ${connection.id}`);
+  onConnect(connection: Connection, ctx: { request: Request }) {
+    // Extract authentication from WebSocket connection
+    const url = new URL(ctx.request.url);
```
threepointone, line 14: move this to onBeforeConnect

examples/cross-domain/src/server.ts
```diff
@@ -27,21 +62,85 @@ export class MyAgent extends Agent {
     }
   }
 
-  onRequest(_request: Request): Response | Promise<Response> {
+  onRequest(request: Request): Response | Promise<Response> {
+    // Handle HTTP authentication for API requests
+    const authHeader = request.headers.get("Authorization");
```
threepointone: move this to onBeforeRequest

packages/agents/src/react.tsx
```diff
@@ -27,22 +27,33 @@ function camelCaseToKebabCase(str: string): string {
...
+  /** Query parameters - can be static object or async function */
+  query?: QueryObject | (() => Promise<QueryObject>);
+  /** Dependencies for async query caching */
+  queryDeps?: unknown[];
   /** Called when the Agent's state is updated */
   onStateUpdate?: (state: State, source: "server" | "client") => void;
   /** Called when MCP server state is updated */
   onMcpUpdate?: (mcpServers: MCPServersState) => void;
+  /** Enable debug logging */
+  debug?: boolean;
```
threepointone: remove this?

packages/agents/src/react.tsx
```diff
@@ -167,14 +205,100 @@ export function useAgent<State>(
     >()
   );
 
-  // TODO: if options.query is a function, then use
-  // "use()" to get the value and pass it
-  // as a query parameter to usePartySocket
+  // Handle both sync and async query patterns
+  const cacheKey = useMemo(() => {
+    const deps = queryDeps?.map((d) => String(d)).join("_") || "no-deps";
```
threepointone: this is dangerous because you'll get a bunch of [Object object]

packages/agents/src/react.tsx
```diff
@@ -167,14 +205,100 @@ export function useAgent<State>(
...
+    const promise = query().catch((error) => {
+      console.error(
+        `[useAgent] Query failed for agent "${options.agent}":`,
+        error
+      );
+      queryCache.delete(cacheKey); // Remove failed promise from cache
+      throw error; // Re-throw for Suspense error boundary
+    });
+
+    queryCache.set(cacheKey, promise);
```
threepointone: instead set the array over here as the key (which will work since it's a map). you'll still have to iterate over the map, and pull out instances from the array and compare (since you can't compare the array directly, since it'll probably be  new on every call)

packages/agents/src/react.tsx
```diff
@@ -27,18 +27,126 @@ function camelCaseToKebabCase(str: string): string {
...
+  }
+
+  return false;
+}
+
+function arraysEqual(a: unknown[], b: unknown[]): boolean {
+  if (a.length !== b.length) return false;
+
+  for (let i = 0; i < a.length; i++) {
+    if (!deepEqual(a[i], b[i])) return false;
```
threepointone: this should actually be a shallow equal, not a deep equal (following the same rules as react deps)

docs/authentication.md
```diff
@@ -0,0 +1,186 @@
...
+When your Agents are deployed, to keep things secure, send a token from the client, then verify it on the server. This mirrors the shape used in PartyKit’s auth guide.
+
+## WebSocket authentication
+
+WebSockets are not HTTP, so the handshake is limited.
+
+**What you cannot send**
+
+- Custom headers during the upgrade
+- `Authorization: Bearer ...` on connect
```
threepointone: need to be clear that this is only true when cross domain

docs/authentication.md
```diff
@@ -0,0 +1,186 @@
+# Authentication
```
threepointone: label this as cross domain auth

docs/authentication.md
```diff
@@ -0,0 +1,186 @@
...
+function ChatComponent() {
+  const agent = useAgent({
+    agent: "my-agent",
+    query: {
+      token: "demo-token-123",
+      userId: "demo-user"
+    }
+  });
+
+  return <div>Agent: {agent.agent}</div>;
```
threepointone: what is agent.agent and why is it output here 

(remove it?)

docs/authentication.md
```diff
@@ -0,0 +1,186 @@
...
+
+```ts
+import { useAgent } from "agents/react";
+import { useCallback } from "react";
+
+function useJWTAgent(agentName: string) {
+  const asyncQuery = useCallback(async () => {
+    let token = localStorage.getItem("jwt");
+
+    if (!token || isTokenExpired(token)) {
```
threepointone: how can you tell that a jwt token is expired? 

in practice I think the call will fail, and you have to trigger the async call only then iirc

docs/cross-domain-authentication.md
```diff
@@ -0,0 +1,186 @@
...
+
+For HTTP calls you can use headers and credentials.
+
+```ts
+import { useAgentChat } from "agents/ai-react";
+
+// Bearer token
+const chat = useAgentChat({
+  agent,
+  headers: {
```
threepointone: hmm does this work?

packages/agents/src/react.tsx
```diff
@@ -167,14 +250,82 @@ export function useAgent<State>(
...
+      }
+    };
+  }, [cacheKey, queryPromise]);
+
   const agent = usePartySocket({
     party: agentNamespace,
     prefix: "agents",
     room: options.name || "default",
-    ...options,
+    query: resolvedQuery as Record<string, string | null | undefined>,
```
threepointone: is this `as` clause necessary?

.changeset/async-auth-improvements.md
```diff
@@ -0,0 +1,6 @@
+---
+"agents": minor
```
threepointone: maybe it's fine if this is a patch release 
```suggestion
"agents": patch
```

docs/cross-domain-authentication.md
```diff
@@ -0,0 +1,190 @@
+# Cross-Domain Authentication
```
threepointone, line 1: rename the file to be cross-domain-authentication.md

docs/authentication.md
```diff
@@ -0,0 +1,190 @@
...
+
+```ts
+import { useAgent } from "agents/react";
+import { useCallback, useEffect } from "react";
+
+function useJWTAgent(agentName: string) {
+  const asyncQuery = useCallback(async () => {
+    let token = localStorage.getItem("jwt");
+
+    // In practice, you'll only know the token expired when the call fails
```
threepointone: where's the example where the token fails?

threepointone: the error will probably come on useAgent's onError. how can we check and trigger a token refresh from there?

docs/authentication.md
```diff
@@ -0,0 +1,190 @@
...
+      token,
+      userId: "demo-user"
+    };
+  }, []);
+
+  const agent = useAgent({
+    agent: agentName,
+    query: asyncQuery,
+    queryDeps: [], // Run on mount
+    debug: true
```
threepointone: this doesn't exist as a config field anymore

examples/cross-domain/src/client.tsx
```diff
@@ -85,23 +169,282 @@ function App() {
...
+      };
+      setMessages((prev) => [...prev, errorMessage]);
+    }
+  };
+
+  const updateAuthToken = () => {
+    if (tokenInputRef.current?.value) {
+      setAuthToken(tokenInputRef.current.value);
+      // Note: Changing the token will require reconnecting the WebSocket
+      window.location.reload();
```
threepointone: this seems wrong, why are we recommending a page reload when you get a new token?

deathbyknowledge: Wasn't needed either, removed

docs/authentication.md
deathbyknowledge: I didn't want to handle authentication in `onError`, so I've updated the example to also include `validateToken`, which is also async

docs/cross-domain-authentication.md
deathbyknowledge: It only works for the HTTP messages `useChatAgent` might do, like getting initial messages. I've updated the doc to make it clear

docs/cross-domain-authentication.md
```diff
@@ -0,0 +1,189 @@
...
+  }, []);
+
+  const agent = useAgent({
+    agent: agentName,
+    query: asyncQuery,
+    queryDeps: [] // Run on mount
+  });
+
+  // Handle authentication errors
+  // Token expired, refresh and reconnect
```
threepointone: can you show what this would look like in the code sample?

docs/cross-domain-authentication.md
```diff
@@ -0,0 +1,189 @@
...
+      userId: user.id,
+      timestamp: Date.now().toString()
+    };
+  }, []);
+
+  const agent = useAgent({
+    agent: "my-agent",
+    host: "http://localhost:8788",
+    query: asyncQuery,
+    debug: true
```
threepointone: not a field anymore

docs/cross-domain-authentication.md
threepointone: maybe it's not a good idea to have it all here then as an option? it means they'll get initial messages but none of the subsequent messages will go through/come back

examples/cross-domain/src/client.tsx
```diff
@@ -74,9 +121,45 @@ function App() {
...
   return (
     <div className="chat-container">
-      <div className="status-indicator">
-        <div className={`status-dot ${isConnected ? "connected" : ""}`} />
-        {isConnected ? "Connected to server" : "Disconnected"}
+      <div className="auth-section">
+        <h2>Cross-Domain Authentication Demo (Async)</h2>
+        <div className="status-indicator">
+          <div className={`status-dot ${isConnected ? "connected" : ""}`} />
+          <div className="connection-status">
```
threepointone: imo this bit about showing connection state is distracting from the auth stuff, can you remove it?

docs/cross-domain-authentication.md
deathbyknowledge: Agreed, removed

examples/cross-domain/src/client.tsx
```diff
@@ -85,23 +145,271 @@ function App() {
...
+        </div>
+      </div>
+    </div>
+  );
+}
+
+function StaticAuthApp() {
+  const [authToken, setAuthToken] = useState("demo-token-123");
+  const [messages, setMessages] = useState<Message[]>([]);
+  const [isConnected, setIsConnected] = useState(false);
```
threepointone: remove all the connected stuff, in every component

examples/cross-domain/src/client.tsx
```diff
@@ -86,22 +142,260 @@ function App() {
...
+        <div className="debug-info">
+          <p>
+            <strong>Agent:</strong> {agent.agent}
+          </p>
+          <p>
+            <strong>Room:</strong> {agent.name}
+          </p>
+          <p>
+            <strong>WebSocket Ready State:</strong> {agent.readyState}
+          </p>
```
threepointone:
```suggestion
```
thus won't rerender on updates, so let's remove it

examples/cross-domain/src/client.tsx
```diff
@@ -86,22 +142,260 @@ function App() {
...
```
threepointone:
```suggestion
```

Issue #464, opened 2025-09-15
React Router V7 sendMessage not work


BeiXiao:
I'm using React Router v7 with Hono as the backend, both in the same project deployed to the same Workers. When using useAgentChat, fetching messages works fine and I can connect to our agents, but sendMessage doesn't work at all.
I've now tried accessing our agents' Workers URL through the Chat Agent Starter project, and sendMessage works, but as soon as I use our React Router v7 to access directly, sendMessage not work.

BeiXiao:
I debugged based on the source code and found that it should be an issue with **ai-react.tsx**. The **agent.readyState** in **aiFetch is 3,** but if I directly print **agent.readyState** in **clearHistory**, it shows 1. Therefore, I suspect there is a problem with the agent instance used in aiFetch.

BeiXiao:
#465 I have submitted a PR to fix this issue. Please help review it.

jcheese1:
FYI @BeiXiao i fixed the issue by removing `<StrictMode />`. you'll have to eject it if you dont see `entry.client.tsx`

jcheese1:
cc @whoiskatrin

BeiXiao:
> FYI [@BeiXiao](https://github.com/BeiXiao) i fixed the issue by removing `<StrictMode />`. you'll have to eject it if you dont see `entry.client.tsx`

Thank you, I'll give it a try. So is it because of StrictMode causing the multiple rendering issues?

jcheese1:
> > FYI [@BeiXiao](https://github.com/BeiXiao) i fixed the issue by removing `<StrictMode />`. you'll have to eject it if you dont see `entry.client.tsx`
> 
> Thank you, I'll give it a try. So is it because of StrictMode causing the multiple rendering issues?
> 

In my case, yes

whoiskatrin:
https://github.com/cloudflare/agents/pull/465 has been merged

PR #465, opened 2025-09-15
fix(ai-react): prevent stale agent capture in aiFetch; ensure active connection is used

BeiXiao:
**Description**
- Issue: aiFetch captured the agent instance at hook init. After reconnect/replace, aiFetch still sent/received via the old (closed) socket (readyState 3) while the current agent was connected (readyState 1), breaking sendMessage and streaming.
- Fix: keep a ref to the latest agent and read it inside aiFetch; memoize the transport so it doesn’t re-capture stale references.
- Scope: internal only; no public API changes.

**Reproduction**
1) Start a chat; 
2) trigger agent reconnect/replace; 
3) send a new message → no response or broken stream due to old agent being used.

**Implementation**
- Add agentRef + effect to track the latest agent.
- Wrap aiFetch with useCallback and use agentRef.current for send/addEventListener.
- Wrap transport with useMemo, depending on agentUrlString and aiFetch.


whoiskatrin:
@BeiXiao before we review, could you please fix the build? thank you

packages/agents/src/ai-react.tsx
```diff
@@ -298,9 +305,9 @@ export function useAgentChat<
     );
 
     return new Response(stream);
-  }
+  }, [agentUrlString]);
```
threepointone: if we have this, why do we need to put agent on a ref?

BeiXiao: > if we have this, why do we need to put agent on a ref?

The **agentRef** ensures we always use the latest agent object inside the callback. Without it, the callback would be stuck with the old agent from when it was first created.

BeiXiao:
> @BeiXiao before we review, could you please fix the build? thank you

Hi, fixed it, all checks have passed 🎉
Please continue review, thanks ~

packages/agents/src/ai-react.tsx
threepointone: but doesn't agentUrlString changing also capture the new agent instance?

BeiXiao: > but doesn't agentUrlString changing also capture the new agent instance?

OK,  latest commit has removed agentUrlString from the dependency array. 
The main issue here is that the agent captured in aiFetch is unconnected, so we used agentRef to preserve the agent instance.

axuj:
Does this really fix the issue? It seems a bit strange. In the ai-sdk, the transport is used in [AbstractChat #L214C2-L214C3](https://github.com/vercel/ai/blob/3a10095ad1f6d33ed88e359891897abc2c070add/packages/ai/src/ui/chat.ts#L214C2-L214C3), and the Chat instance is created in [useChat #L61](https://github.com/vercel/ai/blob/651860ebefb22b313ed89b3e23148d63d13226e2/packages/react/src/use-chat.ts#L61). The logic for ai-sdk to determine whether to update the chat instance is in [useChat #L65](https://github.com/vercel/ai/blob/651860ebefb22b313ed89b3e23148d63d13226e2/packages/react/src/use-chat.ts#L65). It appears that this change doesn't update the transport in the ai-sdk. To update the instance, you would need to pass a new [chat](https://ai-sdk.dev/cookbook/next/use-shared-chat-context) or a new id, like in #440.

jcheese1:
FYI just straight up removing `<StrictMode />` fixed most of these issues for me

fforres:
We encountered something similar, I think this issue is related to this https://github.com/cloudflare/agents/pull/440

Removing strict-mode prevents the "change of ID" from happening
In a way, similar outcome is done on issue 440. than with this: https://github.com/cloudflare/agents/pull/465/files#diff-c4dd80b36e6832a71f0a59b28423a9acf6068806810ec96899681f0b83e774a8R122-R125

_Or if you do something like this:_
```TSX
const agent = useAgent({
  agent: 'chat',
  name: conversationId,
})
const agentChat = useAgentChat({
  agent,
  id: agent._pk,
})
```

Not sure what approach folks want to go with, but would be great to unify the conversations:
- https://github.com/cloudflare/agents/issues/439
- https://github.com/cloudflare/agents/issues/464
- https://github.com/cloudflare/agents/pull/440

whoiskatrin:
> We encountered something similar, I think this issue is related to this #440
> 
> Removing strict-mode prevents the "change of ID" from happening In a way, similar outcome is done on issue 440. than with this: https://github.com/cloudflare/agents/pull/465/files#diff-c4dd80b36e6832a71f0a59b28423a9acf6068806810ec96899681f0b83e774a8R122-R125
> 
> _Or if you do something like this:_
> 
> ```tsx
> const agent = useAgent({
>   agent: 'chat',
>   name: conversationId,
> })
> const agentChat = useAgentChat({
>   agent,
>   id: agent._pk,
> })
> ```
> 
> Not sure what approach folks want to go with, but would be great to unify the conversations:
> 
> * [Stale WebSocket Instances Used in useAgentChat #439](https://github.com/cloudflare/agents/issues/439)
> * [React Router V7 sendMessage not work #464](https://github.com/cloudflare/agents/issues/464)
> * [fix: pass agent._pk as id to useChat to prevent stale WebSocket instances #440](https://github.com/cloudflare/agents/pull/440)

super useful feedback, thanks for this, we just need to really dive deeply into whats happening and discuss the best available fix, should be not too long

threepointone:
I don't recommend using _pk as the chat identifier since it's unique per connection, I suspect .name might be better... I need to dig into that. in the meanwhile, this PR seems ok.

Issue #456, opened 2025-09-11
MCP Server regression in agents@0.1.2


kentcdodds:
Loom walkthrough: https://www.loom.com/share/0308cdfa98d04e8f839b0d0aab80eb53

Here are some logs when I try to use [this server](https://github.com/kentcdodds/cloudflare-no-init-mcp) in [Goose (by Block)](https://block.github.io/goose/):

Starting with agents@0.1.1:

```sh
~/Desktop/cloudflare-no-init-mcp (main) 👽
$ npm run dev

> remote-mcp-server-authless@0.0.0 dev
> wrangler dev


 ⛅️ wrangler 4.35.0
───────────────────
Your Worker has access to the following bindings:
Binding                     Resource            Mode
env.MCP_OBJECT (MyMCP)      Durable Object      local

╭──────────────────────────────────────────────────────────────────────╮
│  [b] open a browser [d] open devtools [c] clear console [x] to exit  │
╰──────────────────────────────────────────────────────────────────────╯
⎔ Starting local server...
[wrangler:info] Ready on http://localhost:8787
State updated
Connection established
Connection W3XLlYl387HVeVnlPhNP5 connected to _a:a8fec08aa309c5e6dad7db4c38d8b210e069829f705c9a1d131692ef66038c90
[wrangler:info] POST /mcp 200 OK (14ms)
Connection established
Connection fQZ8Ofj9KtgEyp1jMshW1 connected to _a:a8fec08aa309c5e6dad7db4c38d8b210e069829f705c9a1d131692ef66038c90
[wrangler:info] POST /mcp 202 Accepted (1ms)
[wrangler:info] GET /mcp 405 Method Not Allowed (0ms)
[wrangler:info] DELETE /mcp 405 Method Not Allowed (3ms)
State updated
Connection established
Connection AEzKG_bs1uYJWQ8z7JsG5 connected to _a:eacb8ad58f917ae5ae7930e2465f3cc0d0ec6991abcaa1f041fd524767d1cddf
[wrangler:info] POST /mcp 200 OK (8ms)
Connection established
Connection 2lpLSX3qG6oSaMRt0Nae6 connected to _a:eacb8ad58f917ae5ae7930e2465f3cc0d0ec6991abcaa1f041fd524767d1cddf
[wrangler:info] POST /mcp 202 Accepted (2ms)
[wrangler:info] GET /mcp 405 Method Not Allowed (1ms)
Connection established
[...]

PR #467, opened 2025-09-15
Silently handle writer close errors

deathbyknowledge:
Since we're closing response writers in event handlers, we can't reliably know which one will run first and might try to close the same one twice. This PR simply makes them fail silently.

We only need to update Streamable HTTP handlers since SSE was already doing this.

Addresses #456


threepointone (approved):
add a changeset

PR #469, opened 2025-09-15
feat: Include reasoning text in finalized message

umgefahren:
The finalized message will now also include the final reasoning text.

In general one should consider if this part of the agent sdk shouldn't be reworked (a lot) because the current generation of reasoning models (Claude 4, GPT-5 and Grok 4) reason a little, call a tool and then continue to reason. I think this should also be persistet so people don't get confused. The ai-sdk probably already includes some code for this kind of persistence. Maybe the agent-sdk should just use that instead of rolling it's own.

### DIFF
--- a/.changeset/plain-pigs-lead.md
+++ b/.changeset/plain-pigs-lead.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
+Include reasoning parts in finalized and persistet message.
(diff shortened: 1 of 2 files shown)


umgefahren:
Is it possible maybe to use the [onFinish](https://ai-sdk.dev/docs/reference/ai-sdk-core/stream-text#to-ui-message-stream.ui-message-stream-options.on-finish) for message persistance? Because that should include the finished message.

whoiskatrin:
> Is it possible maybe to use the [onFinish](https://ai-sdk.dev/docs/reference/ai-sdk-core/stream-text#to-ui-message-stream.ui-message-stream-options.on-finish) for message persistance? Because that should include the finished message.

We didn't do it for a reason. I will check if it's possible now, if so we will change it probably.

whoiskatrin (approved):

PR #470, opened 2025-09-15
Store initialize requests and set them in onStart

deathbyknowledge:
The MCP TS SDK's Server [sets the client information](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/server/index.ts#L295-L296) when dealing with an `InitializeRequest` in memory.
 
This makes it so that IF an `McpAgent` hibernates or is evicted when already initialized, this data (e.g. `this.server.server.getClientCapabilities()`) will no longer be available since clients won't try to initialize again.

Since we were already storing a boolean to signal whether the `McpAgent` was initialized (only used internally, so I think it's OK to update this), we can store the `InitializeRequest` instead and send it on every startup when present, so `this.server` always has the data available.

~PS:  Also updated the methods to use the new KV sync API~

### DIFF
--- a/packages/agents/src/mcp/index.ts
+++ b/packages/agents/src/mcp/index.ts
@@ -69,7 +69,7 @@ export abstract class McpAgent<
     await this.ctx.storage.put("initializeRequest", initializeRequest);
   }
 
-  getInitializeRequest() {
+  async getInitializeRequest() {
     return this.ctx.storage.get<JSONRPCMessage>("initializeRequest");
   }
 
--- a/packages/agents/src/mcp/index.ts
+++ b/packages/agents/src/mcp/index.ts
@@ -65,12 +65,12 @@ export abstract class McpAgent<
    * Helpers
    */
 
-  setInitializeRequest(initializeRequest: JSONRPCMessage) {
-    this.ctx.storage.kv.put("initializeRequest", initializeRequest);
+  async setInitializeRequest(initializeRequest: JSONRPCMessage) {
+    await this.ctx.storage.put("initializeRequest", initializeRequest);
   }
 
   getInitializeRequest() {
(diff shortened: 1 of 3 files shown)


packages/agents/src/mcp/index.ts
```diff
@@ -65,13 +65,12 @@ export abstract class McpAgent<
...
-    // TODO: move to sync api once https://github.com/cloudflare/workerd/pull/4895 lands
-    await this.ctx.storage.put("initialized", true);
+  setInitializeRequest(initializeRequest: JSONRPCMessage) {
+    this.ctx.storage.kv.put("initializeRequest", initializeRequest);
   }
 
-  async isInitialized() {
-    return (await this.ctx.storage.get("initialized")) === true;
+  getInitializeRequest() {
+    return this.ctx.storage.kv.get<JSONRPCMessage>("initializeRequest");
```
threepointone: this will require an update of runtime, might break older code 
can we use the older version for now? we can update later

deathbyknowledge: Yeah ofc

packages/agents/src/mcp/index.ts
```diff
@@ -65,13 +65,12 @@ export abstract class McpAgent<
...
-  async setInitialized() {
-    // TODO: move to sync api once https://github.com/cloudflare/workerd/pull/4895 lands
-    await this.ctx.storage.put("initialized", true);
+  async setInitializeRequest(initializeRequest: JSONRPCMessage) {
+    await this.ctx.storage.put("initializeRequest", initializeRequest);
   }
 
-  async isInitialized() {
-    return (await this.ctx.storage.get("initialized")) === true;
+  getInitializeRequest() {
```
threepointone: async

```suggestion
  async getInitializeRequest() {
```

PR #478, opened 2025-09-17
Refactor streamable HTTP transport

deathbyknowledge:
Refactor the custom Streamable HTTP transport to closely match the behavior of the [MCP SDK's implementation](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/server/streamableHttp.ts)

This also updates how we bridge from the Worker to the transport within the agent.

**PREVIOUSLY**:
- The `McpAgent.serve(...)` handler in the Worker would get a POST request, read the body and make sure it was a valid JSON-RPC message and apply validation per the spec. (this we keep)
- Then, it would establish a WebSocket to the `McpAgent`, forward the JSON-RPC payloads over WS, keep a requestId -> connectionId map, send a private `cf/standalone_sse/attach` notif to bind the “standalone SSE stream” ([this](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#listening-for-messages-from-the-server)) socket
- Close once all request IDs were answered.

**NOW**:
* Worker sets headers on the WS upgrade: `cf-mcp-method` (GET/POST) and `cf-mcp-message` (POST body JSON).
* Agent reads those headers in `onConnect` and calls `transport.handlePostRequest(...)` or `transport.handleGetRequest(...)`.
* New `StreamableHTTPServerTransport` replaces `McpStreamableHttpTransport`. It tracks `requestIds` and decides when to close.
* Outbound events go over WS as `{ type: "cf_mcp_agent_event", event, close }` and the worker writes `event` verbatim to SSE and closes when `close` is true.
* Kills the `requestId -> connectionId` map and the `cf/standalone_sse/attach` path, we track everything in the `connection.state`.

This allows us to have a transport that closely mimics the official one (except the validation, which we already do in the Worker handler), easier to maintain + more feature parity.
This also includes the `extra` parameter in tool handlers, with request information. So, the following also works now:
```typescript
    this.server.registerTool("someTool", {
     // params...
     }, async ({...}, extra) =>  {
       console.log(extra.requestInfo.headers);
       extra.sendNotification("someTool called...");
       return {
         content: [{ type: "text", text: ... }],
       }
});
```

### DIFF
--- a/packages/agents/src/mcp/transport.ts
+++ b/packages/agents/src/mcp/transport.ts
@@ -13,11 +13,7 @@ import type { AuthInfo } from "@modelcontextprotocol/sdk/server/auth/types.js";
 import { getCurrentAgent, type Connection } from "..";
 import type { McpAgent } from ".";
 import { MessageType } from "../ai-types";
-import {
-  MCP_HTTP_METHOD_HEADER,
-  MCP_MESSAGE_HEADER,
-  STANDALONE_SSE_MARKER
-} from "./utils";
+import { MCP_HTTP_METHOD_HEADER, MCP_MESSAGE_HEADER } from "./utils";
 
 export class McpSSETransport implements Transport {
   sessionId?: string;
@@ -109,7 +105,6 @@ export interface StreamableHTTPServerTransportOptions {
  * Besides these points, the implementation is the same and should be updated to match the original as new features are added.
  */
 export class StreamableHTTPServerTransport implements Transport {
-  // when sessionId is not set (undefined), it means the transport is in stateless mode
   private _started = false;
   private _eventStore?: EventStore;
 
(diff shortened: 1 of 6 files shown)


packages/agents/src/mcp/transport.ts
```diff
@@ -51,53 +62,251 @@ export class McpSSETransport implements Transport {
...
-export class McpStreamableHttpTransport implements Transport {
-  sessionId?: string;
-  // Set by the server in `server.connect(transport)`
-  onclose?: () => void;
-  onerror?: (error: Error) => void;
-  onmessage?: (message: JSONRPCMessage) => void;
+export type StreamId = string;
+export type EventId = string;
+
+// TODO: Implement this and make it opt-in?
```
threepointone, line 64: file an issue for this?

packages/agents/src/mcp/transport.ts
```diff
@@ -51,53 +62,251 @@ export class McpSSETransport implements Transport {
...
+  /**
+   * A callback for session close events
+   * This is called when the server closes a session due to a DELETE request.
+   * Useful in cases when you need to clean up resources associated with the session.
+   * Note that this is different from the transport closing, if you are handling
+   * HTTP requests from multiple nodes you might want to close each
+   * StreamableHTTPServerTransport after a request is completed while still keeping the
+   * session open/running.
+   */
+  onsessionclosed?: () => void | Promise<void>;
```
threepointone: you said we'll remove this

PR #479, opened 2025-09-17
add step count limit and auto-send control for tool confirmations

### LINKED ISSUE #477: Multiple confirmations results in APICallError [AI_APICallError]: No tool output found for function call
### repro
using `/guides/human-in-the-loop`

- start dev server
- prompt `Get weather in Boston and New York`
- 2 confirmations appear
- First one works, clicking 'Yes' to approve the second one produces the error

```js
    error: {
      message: 'No tool output found for function call call_FsdCmUUmqJVoAPpMC02IsjTC.',
      type: 'invalid_request_error',
      param: 'input',
      code: null
    }
```

### related issue with multi-step
Tool responses are not being returned to the LLM for tools which do not require confirmation

- prompt `Get news in Boston and New York and compare them`
- tool returns the same result for both function calls
- but because multi-step is not enabled, the 2 responses are not passed back to the LLM, and there is no comparison.

adding `stopWhen: stepCountIs(5)` in streamText() resolves this.

whoiskatrin:
fixes #477 

- added stopWhen: stepCountIs(5) for more than 2 steps prompts 
- for more than one confirmation added a function in ai-react to wait until there are no pending confirmations and only then send messages

### DIFF
--- a/guides/human-in-the-loop/src/server.ts
+++ b/guides/human-in-the-loop/src/server.ts
@@ -6,7 +6,8 @@ import {
   createUIMessageStream,
   createUIMessageStreamResponse,
   type StreamTextOnFinishCallback,
-  streamText
+  streamText,
+  stepCountIs
 } from "ai";
 import { tools } from "./tools";
 import {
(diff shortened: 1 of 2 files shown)


threepointone (approved):
assume you've tested this, feel free to land anytime

PR #483, opened 2025-09-19
Resumable streams support

whoiskatrin:
so far played with:
- creating new ai chat agent that uses http instead of websockets
- same for the use chat agent
- doesn't sync the streaming in tabs rn, only single tab
- using DO storage for keeping streaming ids and replaying it after 

TODO: sync tabs, more tests, docs


jcheese1:
will this be merged soon?

threepointone:
working on it

whoiskatrin:
closing for now, this will be reworked into http ai chat agent and useChatAgent

PR #485, opened 2025-09-20
update mcp agent to handle pending reqs for elicitation (fix) and sampling (new)

whoiskatrin:
we had a bug with elicitation, pending requests would just hang. my previous pooling based approach wasn't the correct one  in this case (that's on me) 

updated the demo, removed mcp-elicitation-demo (we don't need it anymore), the latest version of inspector actually supports the elicitation, i've just added new tools to test it in our existing mcp demo
also added precaution to survive hibernation


kentcdodds:
I've deployed https://cloudflare-sampling-elicitation-issues.kentcdodds.workers.dev/mcp with this and it's working great.

image

Note the regular MCP SDK elicitation still does not work with on Cloudflare infra because the use of `eval` ([read more](https://github.com/modelcontextprotocol/typescript-sdk/issues/689). Not sure what to do about that. But the specially elicitation API you've made works ok.

image

I would love to be able to just rely on the standard SDK for everything so hopefully https://github.com/modelcontextprotocol/typescript-sdk/issues/689 gets resolved.

But I'm happy with this!

whoiskatrin:
> I've deployed https://cloudflare-sampling-elicitation-issues.kentcdodds.workers.dev/mcp with this and it's working great.
> 
> image
> Note the regular MCP SDK elicitation still does not work with on Cloudflare infra because the use of `eval` ([read more](https://github.com/modelcontextprotocol/typescript-sdk/issues/689). Not sure what to do about that. But the specially elicitation API you've made works ok.
> 
> image
> I would love to be able to just rely on the standard SDK for everything so hopefully [modelcontextprotocol/typescript-sdk#689](https://github.com/modelcontextprotocol/typescript-sdk/issues/689) gets resolved.
> 
> But I'm happy with this!

we are genuinely considering just waiting for the main MCP SDK issue to resolve, instead of adding all this complexity

kentcdodds:
From what I can tell, it does look like they're working on it.

whoiskatrin:
Closing to wait for the fix in MCP SDK.

PR #495, opened 2025-09-22
Fix OAuth callback handling and add HOST auto-detection

ghostwriternr:
## Summary

This PR fixes OAuth callback handling for MCP servers and adds automatic HOST detection to simplify configuration.

### OAuth Fixes
- Remove MCPClientManager override in MyAgent example that was preventing OAuth callbacks from working
- Add callback URL persistence across Durable Object hibernation 
- Fix connection reuse during OAuth reconnect to prevent state loss
- Add transport type detection from URL patterns to prevent OAuth code consumption

### Enhancement
- Make `callbackHost` parameter optional in `addMcpServer()` - automatically derives from request when not provided
- Update examples to reflect HOST as optional environment variable

## Test Plan
- [x] OAuth callbacks work with Linear MCP server
- [x] OAuth state persists across DO hibernation
- [x] SSE transport connections work with OAuth
- [x] HOST auto-detection works when variable is not set
- [x] Backward compatibility maintained for existing code
- [x] All checks pass (formatting, linting, typechecking)

### DIFF
--- a/packages/agents/src/index.ts
+++ b/packages/agents/src/index.ts
@@ -1514,12 +1514,21 @@ export class Agent<
       };
     }
 
+    // Detect transport type from URL pattern
+    let transportType: "sse" | "streamable-http" | "auto" = "auto";
+    if (url.endsWith("/sse")) {
+      transportType = "sse";
+    } else if (url.endsWith("/mcp")) {
+      transportType = "streamable-http";
+    }
+
     const { id, authUrl, clientId } = await this.mcp.connect(url, {
       client: options?.client,
       reconnect,
       transport: {
         ...headerTransportOpts,
-        authProvider
+        authProvider,
+        type: transportType
       }
     });
 
--- a/packages/agents/src/index.ts
+++ b/packages/agents/src/index.ts
@@ -1406,8 +1406,9 @@ export class Agent<
   /**
    * Connect to a new MCP Server
    *
+   * @param serverName Name of the MCP server
    * @param url MCP Server SSE URL
-   * @param callbackHost Base host for the agent, used for the redirect URI.
+   * @param callbackHost Base host for the agent, used for the redirect URI. If not provided, will be derived from the current request.
    * @param agentsPrefix agents routing prefix if not using `agents`
    * @param options MCP client and transport (header) options
    * @returns authUrl
@@ -1415,7 +1416,7 @@ export class Agent<
   async addMcpServer(
     serverName: string,
     url: string,
-    callbackHost: string,
+    callbackHost?: string,
     agentsPrefix = "agents",
     options?: {
       client?: ConstructorParameters<typeof Client>[1];
@@ -1424,7 +1425,22 @@ export class Agent<
       };
     }
   ): Promise<{ id: string; authUrl: string | undefined }> {
-    const callbackUrl = `${callbackHost}/${agentsPrefix}/${camelCaseToKebabCase(this._ParentClass.name)}/${this.name}/callback`;
+    // If callbackHost is not provided, derive it from the current request
+    let resolvedCallbackHost = callbackHost;
+    if (!resolvedCallbackHost) {
+      const { request } = getCurrentAgent();
+      if (!request) {
+        throw new Error(
+          "callbackHost is required when not called within a request context"
+        );
+      }
+
+      // Extract the origin from the request
+      const requestUrl = new URL(request.url);
+      resolvedCallbackHost = `${requestUrl.protocol}//${requestUrl.host}`;
+    }
+
+    const callbackUrl = `${resolvedCallbackHost}/${agentsPrefix}/${camelCaseToKebabCase(this._ParentClass.name)}/${this.name}/callback`;
 
     const result = await this._connectToMcpServerInternal(
       serverName,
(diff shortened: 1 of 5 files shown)


ghostwriternr:
> tentatively looks good, think we can add a test?

Yep, working on tests! There were quite a lot of minute issues that had to be fixed, which took a lot of time to even discover manually. Need tests.

threepointone:
no rush then!

packages/agents/src/tests/mcp/transports/auto.test.ts
```diff
@@ -71,14 +71,18 @@ describe("Auto Transport Mode", () => {
...
+        "sse" // but we explicitly request SSE
       );
 
-      await expect(connection.init()).rejects.toThrow();
-      expect(connection.connectionState).toBe("failed");
+      await connection.init();
+
+      // Should succeed because URL normalization strips /mcp and tries /sse
+      expect(connection.connectionState).toBe("ready");
+      expect(connection.tools).toBeDefined();
```
ghostwriternr: This is the main behavioural change in this PR that I'm reconsidering. Everything else is net positive.

ghostwriternr: And this is not a _necessary_ change, but keeping it this way makes it harder to shoot yourself in the foot as a user. Our auto mode actually works pretty well at this point, and we can just try both transports one after the other.

But I'm wondering if there are valid reasons to trust that explicit > implicit:
1. Explicit gives more control to devs over the upgrade path for spec changes in future
2. More control over error scenarios too

packages/agents/src/index.ts
```diff
@@ -1507,12 +1530,17 @@ export class Agent<
       };
     }
 
+    // Always use auto transport to try /mcp first, then fallback to /sse
+    // This provides consistent behavior regardless of input URL format
+    const transportType = "auto";
```
threepointone: so now I can't override to specifically use a transport type?

packages/agents/src/mcp/client-connection.ts
```diff
@@ -70,6 +73,29 @@ export class MCPClientConnection {
     this.client = new Client(info, clientOptions);
   }
 
+  /**
+   * Normalize URL to extract base URL only
+   * @param url The URL to normalize (may include /mcp or /sse endpoints)
+   * @returns Base server URL without transport-specific endpoints
+   */
+  private _normalizeUrl(url: URL): URL {
```
threepointone: this /sse /mcp thing is only a convention so I'm not sure about this...

packages/agents/src/mcp/client-connection.ts
```diff
@@ -287,46 +313,93 @@ export class MCPClientConnection {
...
+        if (
+          !code &&
+          error.message.includes("Unauthorized") &&
+          this.options.transport.authProvider &&
+          currentTransportType
+        ) {
+          await this.options.transport.authProvider.saveOAuthTransport(
+            currentTransportType
+          );
+          console.log(
```
threepointone: remove this log? or use a debug logger (do we have one?)

packages/agents/src/mcp/do-oauth-client-provider.ts
```diff
@@ -136,7 +140,16 @@ export class DurableObjectOAuthClientProvider implements AgentsOAuthProvider {
   }
 
   async saveCodeVerifier(verifier: string): Promise<void> {
-    await this.storage.put(this.codeVerifierKey(this.clientId), verifier);
+    const key = this.codeVerifierKey(this.clientId);
+
+    // Don't overwrite existing verifier to preserve first PKCE verifier
+    const existing = await this.storage.get<string>(key);
+    if (existing) {
+      console.log(`Preserving existing PKCE verifier for ${this.clientId}`);
```
threepointone: stray log?

packages/agents/src/tests/mcp/client-manager.test.ts
```diff
@@ -0,0 +1,251 @@
...
+      );
+
+      await expect(
+        manager.handleCallbackRequest(callbackRequest)
+      ).rejects.toThrow(
+        "Failed to authenticate: the client isn't in the `authenticating` state"
+      );
+    });
+  });
+});
```
threepointone, line 251: good tests!

packages/agents/src/tests/mcp/transports/auto.test.ts
threepointone: auto is good, but I question the url rewriting
also, explicitly choosing a transport type should be allowed to error if not available? 
let's talk about it

packages/agents/src/index.ts
ghostwriternr: Yea that was the behaviour, and upon sitting with it, doesn't make sense. I was briefly inspired by https://github.com/evalstate/fast-agent, but being able to override is better.

packages/agents/src/mcp/client-connection.ts
ghostwriternr: Agreed - I missed this initially. The spec gives these as an example and everyone follows it, but this is not enforced.

ghostwriternr:
Fixing CI

packages/agents/src/tests/mcp/transports/auto.test.ts
ghostwriternr: I've reverted this behaviour to be exactly how it already is - auto tries HTTP streamable, then SSE, then fails. If explicitly chosen, that transport is tried else fail.

RatStar811 (approved):

PR #500, opened 2025-09-23
Fix OAuth with auto-fallback transport discovery

ghostwriternr:
## Summary

Fixes OAuth authentication failures when using auto-fallback transport with MCP servers.

## Problem
- Auto-fallback was saving the wrong transport during OAuth, causing 404 errors on callback
- Multiple transport attempts generated mismatched PKCE verifiers

## Solution
- Defer transport saving until after OAuth is initiated
- Preserve auth URL and PKCE verifier across transport attempts
- Track correct transport during discovery phase

## Test Plan
- [x] All tests pass (20 tests)
- [x] OAuth works with Sentry MCP server
- [x] No regression for non-OAuth servers
- [x] Auto-fallback and explicit transport modes both work

### DIFF
--- a/.changeset/good-cars-happen.md
+++ b/.changeset/good-cars-happen.md
@@ -1,0 +1,9 @@
+---
+"agents": patch
+---
+
+Fix OAuth with auto-fallback transport discovery
+
+- Defer transport saving until OAuth initiation to prevent wrong transport selection
+- Preserve auth URL and PKCE verifier across transport attempts
+- Fix 404 errors on OAuth callback when using auto-fallback mode


packages/agents/src/mcp/client-connection.ts
```diff
@@ -282,10 +283,19 @@ export class MCPClientConnection {
       "Elicitation handler must be implemented for your platform. Override handleElicitationRequest method."
     );
   }
+
+  /**
+   * Get the transport that was last attempted (for OAuth tracking)
+   */
+  getLastAttemptedTransport(): BaseTransportType | undefined {
```
threepointone, line 290: why have this getter at all?

(I don't feel strongly about this, but figure you can just read lastAttemptedTransport)

ghostwriternr: Was directly reading at first, but then needed a getter for tests

ghostwriternr:
Closed this PR because it was patching over a bunch of hacky solutions instead of a cleaner fix (creating a separate PR)

PR #509, opened 2025-09-25
Decouple OAuth authorization from transport selection

ghostwriternr:
OAuth and transport selection were tightly coupled, causing authorization codes to be consumed during transport fallback attempts. This led to "authorization code already used" errors and made OAuth-protected MCP servers unusable.

Split the OAuth flow into two distinct phases:
1. **OAuth completion** - Finishes auth and probes for working transport
2. **Connection establishment** - Connects using the appropriate transport

## Changes

- Remove transport tracking from OAuth provider (no longer needed)
- Add `finishAuthProbe()` to complete OAuth independently
- Allow transport type configuration in `addMcpServer()` options
- Add configurable OAuth callback handling (redirects, custom handlers)
- Fix callback URL persistence across Durable Object hibernation
- Add error helpers and logging for debugging

## Testing

- All existing tests pass
- OAuth flow tested with Linear MCP server
- Transport fallback tested with both SSE and streamable-http
- Callback URL persistence tested across DO hibernation

## Breaking Changes

None - all changes are backward compatible.

### DIFF
--- a/examples/mcp-client/src/server.ts
+++ b/examples/mcp-client/src/server.ts
@@ -8,7 +8,7 @@ type Env = {
 
 export class MyAgent extends Agent<Env, never> {
   onStart() {
-    // Configure OAuth callback to use popup-closing behavior
+    // Optionally configure OAuth callback. Here we use popup-closing behavior since we're opening a window on the client 
     this.mcp.configureOAuthCallback({
       customHandler: (result: MCPClientOAuthResult) => {
         if (result.authSuccess) {
--- a/examples/mcp-client/src/server.ts
+++ b/examples/mcp-client/src/server.ts
@@ -1,4 +1,5 @@
 import { Agent, type AgentNamespace, routeAgentRequest } from "agents";
+import type { MCPClientOAuthResult } from "agents/mcp";
 
 type Env = {
   MyAgent: AgentNamespace<MyAgent>;
@@ -6,6 +7,28 @@ type Env = {
 };
 
 export class MyAgent extends Agent<Env, never> {
+  onStart() {
+    // Configure OAuth callback to use popup-closing behavior
+    this.mcp.configureOAuthCallback({
+      customHandler: (result: MCPClientOAuthResult) => {
+        if (result.authSuccess) {
+          return new Response("<script>window.close();</script>", {
+            headers: { "content-type": "text/html" },
+            status: 200
+          });
+        } else {
+          return new Response(
+            `<script>alert('Authentication failed: ${result.authError}'); window.close();</script>`,
+            {
+              headers: { "content-type": "text/html" },
+              status: 200
+            }
+          );
+        }
+      }
+    });
+  }
+
   async onRequest(request: Request): Promise<Response> {
     const reqUrl = new URL(request.url);
     if (reqUrl.pathname.endsWith("add-mcp") && request.method === "POST") {
--- a/examples/mcp-client/src/server.ts
+++ b/examples/mcp-client/src/server.ts
@@ -8,7 +8,7 @@ type Env = {
 
 export class MyAgent extends Agent<Env, never> {
   onStart() {
-    // Optionally configure OAuth callback. Here we use popup-closing behavior since we're opening a window on the client 
+    // Optionally configure OAuth callback. Here we use popup-closing behavior since we're opening a window on the client
     this.mcp.configureOAuthCallback({
       customHandler: (result: MCPClientOAuthResult) => {
         if (result.authSuccess) {
--- a/packages/agents/src/core/events.ts
+++ b/packages/agents/src/core/events.ts
@@ -1,0 +1,52 @@
+export interface Disposable {
+  dispose(): void;
+}
+
+export function toDisposable(fn: () => void): Disposable {
+  return { dispose: fn };
+}
+
+export class DisposableStore implements Disposable {
+  private readonly _items: Disposable[] = [];
+
+  add<T extends Disposable>(d: T): T {
+    this._items.push(d);
+    return d;
+  }
+
+  dispose(): void {
+    while (this._items.length) {
+      try {
+        this._items.pop()!.dispose();
+      } catch {
+        // best-effort cleanup
+      }
+    }
+  }
+}
+
+export type Event<T> = (listener: (e: T) => void) => Disposable;
+
+export class Emitter<T> implements Disposable {
+  private _listeners: Set<(e: T) => void> = new Set();
+
+  readonly event: Event<T> = (listener) => {
+    this._listeners.add(listener);
+    return toDisposable(() => this._listeners.delete(listener));
+  };
+
+  fire(data: T): void {
+    for (const listener of [...this._listeners]) {
+      try {
+        listener(data);
+      } catch (err) {
+        // do not let one bad listener break others
+        console.error("Emitter listener error:", err);
+      }
+    }
+  }
+
+  dispose(): void {
+    this._listeners.clear();
+  }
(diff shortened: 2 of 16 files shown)


ghostwriternr:
Deployed the [sample mcp-client here](https://mcp-client.ghostwriternr.workers.dev/) for quick testing

image

packages/agents/src/mcp/client-connection.ts
```diff
@@ -74,24 +79,120 @@ export class MCPClientConnection {
...
     } catch (e: any) {
-      if (e.toString().includes("Unauthorized")) {
+      if (isUnauthorized(e)) {
         // unauthorized, we should wait for the user to authenticate
         this.connectionState = "authenticating";
         return;
       }
+      // For explicit transport mismatches or other errors, mark as failed
+      // and do not throw to avoid bubbling errors to the client runtime.
+      console.error("Connection initialization failed:", e);
```
threepointone: add details about mcp server url/name that it failed with

packages/agents/src/mcp/client-connection.ts
```diff
@@ -74,24 +79,120 @@ export class MCPClientConnection {
...
       this.connectionState = "failed";
+      return;
+    }
+
+    await this.discoverAndRegister();
```
threepointone, line 127: check for idempotency of this call

packages/agents/src/mcp/client-connection.ts
```diff
@@ -74,24 +79,120 @@ export class MCPClientConnection {
...
+  }
+
+  /**
+   * Finish OAuth by probing transports based on configured type.
+   * - Explicit: finish on that transport
+   * - Auto: try streamable-http, then sse on 404/405/Not Implemented
+   */
+  private async finishAuthProbe(code: string): Promise<void> {
+    if (!this.options.transport.authProvider) {
+      throw new Error("No auth provider configured");
```
threepointone, line 137: add info here about name/url of mcp server

packages/agents/src/mcp/client-connection.ts
```diff
@@ -74,24 +79,120 @@ export class MCPClientConnection {
...
+    }
+
+    const configuredType = this.options.transport.type;
+    if (!configuredType) {
+      throw new Error("Transport type must be specified");
```
threepointone, line 142: add info here about name/url of mcp server

packages/agents/src/mcp/client-connection.ts
```diff
@@ -74,24 +79,120 @@ export class MCPClientConnection {
...
+    }
+
+    const tryFinish = async (base: BaseTransportType) => {
```
threepointone: nit: rename this to something else? tryFInishAuth?

packages/agents/src/mcp/client-connection.ts
```diff
@@ -304,67 +405,40 @@ export class MCPClientConnection {
...
       try {
         await this.client.connect(transport);
-
-        // Clear saved transport after successful OAuth completion
-        if (code && this.options.transport.authProvider) {
-          await this.options.transport.authProvider.clearOAuthTransport();
-        }
-
+        this.lastConnectedTransport = currentTransportType;
+        console.log(
```
threepointone: stray log?

threepointone: at least add url/server info

packages/agents/src/mcp/client.ts
```diff
@@ -36,6 +51,13 @@ export class MCPClientManager {
     private _version: string
   ) {}
 
+  /**
+   * Register a callback invoked after a connection is successfully established
+   */
+  onConnected(handler: (serverId: string) => void | Promise<void>): void {
+    this._onConnected = handler;
```
threepointone: if this is a public api, then we have to maintain a list of listeners and call them and clean up if required 

probably just make McpClientManager an event emitter

packages/agents/src/mcp/client.ts
```diff
@@ -158,25 +208,56 @@ export class MCPClientManager {
...
+  }
 
-    if (this.mcpConnections[serverId].connectionState === "authenticating") {
-      throw new Error("Failed to authenticate: client failed to initialize");
+  /**
+   * Establish connection in the background after OAuth completion
+   * This method is called asynchronously and doesn't block the OAuth callback response
+   * @param serverId The server ID to establish connection for
+   */
+  async establishConnectionInBackground(serverId: string): Promise<void> {
```
threepointone: establishConnection

packages/agents/src/tests/mcp/transports/auto.test.ts
```diff
@@ -57,7 +57,7 @@ describe("Auto Transport Mode", () => {
         "auto"
       );
 
-      await expect(connection.init()).rejects.toThrow();
+      await connection.init();
```
threepointone, line 60: think a bit more here 

is this a breaking change?

packages/agents/src/mcp/client-connection.ts
ghostwriternr: I completely forgot about our massively under-utilised observability module. I've wired up the logging to instead happen via that, and I quite like how it has shaped up.

examples/mcp-client/src/server.ts
```diff
@@ -1,11 +1,34 @@
...
+import type { MCPClientOAuthResult } from "agents/mcp";
 
 type Env = {
   MyAgent: AgentNamespace<MyAgent>;
   HOST?: string; // Optional - will be derived from request if not provided
 };
 
 export class MyAgent extends Agent<Env, never> {
+  onStart() {
+    // Configure OAuth callback to use popup-closing behavior
```
threepointone:
```suggestion
    // Optionally configure OAuth callback. Here we use popup-closing behavior since we're opening a window on the client 
```

packages/agents/src/mcp/client.ts
ghostwriternr: This was a great point and I completely missed this on the first go. `onConnected` is now an emitter that any number of listeners can subscribe to and they will all get the event. So even if our end users want to hook into this, that should be okay.

packages/agents/src/core/events.ts
```diff
@@ -0,0 +1,52 @@
+export interface Disposable {
```
threepointone, line 1: Doesn't workers have EventTarget? Or we know nodejs_compat is available, maybe one of the builtins?

packages/agents/src/tests/mcp/transports/auto.test.ts
ghostwriternr: Its not a breaking change in practice (but you could argue that throwing an exception before to not throwing one now is breaking). We could technically just wrap the caller with a try/catch, but at that point, I'm just moving the knowledge of how this method behaves from the method itself over to the caller, which is not an improvement over this design. I think we can keep it this way.

packages/agents/src/core/events.ts
threepointone: I don't mind if we do this, just curious whether it was intentional.

ghostwriternr: I frankly couldn't find any docs online on EventTarget + workers, so I didn't want to risk it (same reason I figured bringing in pino is a bad idea, because they seem to explicitly state that they don't intend on testing with workers).

I checked EventEmitter has been available under nodejs_compat since early 2024, so felt like a safe bet. Furthermore, I quite like this pattern of being able to define well-typed, purpose-specific event emitters - this pattern is borrowed 1:1 from the VSCode codebase.

Issue #488, opened 2025-09-22
MCP Client: Adding an MCP server (Agents.addMcpServer()) with OAuth2 authentication malfunctions.


marcbordessoule:
Description:
When deploying the MCP client example from the [Cloudflare Agents repository](https://github.com/cloudflare/agents/tree/main/examples/mcp-client),
After registering an MCP server using OAuth2,
After clicking the 'Authorize' button,
The message "**Not Found**" is displayed in the browser.

Explanation:
The reception of the callback sent by the authorization server is not always detected in MCPClientManager.isCallbackRequest():
When isCallbackRequest() is executed, the reloading of the registered MCP servers into memory is not yet complete (_connectToMcpServerInternal) and the server currently being authenticated is not yet available in _callbackUrls[]

marcbordessoule:
I propose a temporary workaround:  PR #489

marcbordessoule:
PRs #495, #509, #500 have improved things a lot but there is a new issue when authenticating MCP client with MCP server.

Registering the first mcp server works fine, but the second fails systematically with the message:
Error in MyAgent:nrZVJvY3 fetch: Error: Could not find serverId: callback

Explanation:
The message comes from handleCallbackRequest()

Retrieving the serverId does not work properly. In some cases, callbackUrls are registered (in the addMcpServer phase) without the last part corresponding to the serverId.

An example of callbackUrls:
```ts
this._callbackUrls: [
'http://localhost:5173/agents/my-agent/nrZVJvY3/callback',
'http://localhost:5173/agents/my-agent/nrZVJvY3/callback/7l1uA0VL'
]

// The error is generated here (Agent)
async handleCallbackRequest(req: Request) { async handleCallbackRequest(req: Request) {
    const url = new URL(req.url);
    const urlMatch = this._callbackUrls.find((url) => {
      return req.url.startsWith(url);
    });
    ...
    const serverId = urlParams[urlParams.length - 1];

    ...
    if (this.mcpConnections[serverId] === undefined) {
        throw new Error(`Could not find serverId: ${serverId}`);
    }
```
A idea:
It seems to me that the server id should be created during addMcpServer and not in MCPClientManager.connect()


Furthermore, there seems to be a deeper problem.

For example, during a handleCallbackRequest(), two processes start simultaneously:
- on Agent.onStart(), the MCP servers are reloaded from the database and reconnected.
- on Agent.onRequest(), the callback processing.
Agent.onStart() must be completed before Agent.onRequest().
[...]

PR #515, opened 2025-09-26
addMcpServer: added the serverId further upstream 

marcbordessoule:
MCP client Adding a new MCP server with Oauth2 :

Resolved the problem on the _callbackUrls (#488) and created the serverId upstream in the addMcpServer

### DIFF
--- a/packages/agents/src/mcp/client.ts
+++ b/packages/agents/src/mcp/client.ts
@@ -45,6 +45,7 @@ export class MCPClientManager {
    */
   async connect(
     url: string,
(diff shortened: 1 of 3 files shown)


packages/agents/src/mcp/client.ts
```diff
@@ -45,6 +45,7 @@ export class MCPClientManager {
    */
   async connect(
     url: string,
+    serverId: string,
```
threepointone, line 48: this would be a breaking change, can you move it into options instead?

PR #521, opened 2025-09-26
Fix OAuth state parameter security vulnerability

ghostwriternr:
Replace OAuth client_id values in state parameters with secure random tokens to prevent CSRF attacks and state prediction vulnerabilities.

The current implementation exposes the OAuth client_id directly in the state parameter, making it predictable and potentially enabling cross-site request forgery attacks.

## Solution

Replace state parameter with secure random tokens while storing clientId separately:

```typescript
// Before: Predictable state value
authUrl.searchParams.append("state", client_id);

// After: Secure random token
const stateToken = nanoid();
authUrl.searchParams.set("state", stateToken);
this._clientId_ = client_id; // Store separately
```

Update callback handler to read clientId from stored property instead of state parameter, with fallback to legacy behavior for backward compatibility.

## References

- **OAuth 2.0 Security BCP**: [RFC 6819](https://tools.ietf.org/html/rfc6819) - State parameter should be unguessable random values to prevent CSRF attacks
- **OWASP OAuth Guidelines**: State parameters should be cryptographically random

Supports both static and dynamic client registration flows while maintaining backward compatibility.

### DIFF
--- a/packages/agents/src/mcp/do-oauth-client-provider.ts
+++ b/packages/agents/src/mcp/do-oauth-client-provider.ts
@@ -5,6 +5,7 @@ import type {
   OAuthClientMetadata,
   OAuthTokens
 } from "@modelcontextprotocol/sdk/shared/auth.js";
+import { nanoid } from "nanoid";
 
 // A slight extension to the standard OAuthClientProvider interface because `redirectToAuthorization` doesn't give us the interface we need
 // This allows us to track authentication for a specific server and associated dynamic client registration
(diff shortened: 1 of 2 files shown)


threepointone (approved):
Changeset

packages/agents/src/mcp/do-oauth-client-provider.ts
```diff
@@ -122,12 +123,16 @@ export class DurableObjectOAuthClientProvider implements AgentsOAuthProvider {
...
    */
   async redirectToAuthorization(authUrl: URL): Promise<void> {
-    // We want to track the client ID in state here because the typescript SSE client sometimes does
-    // a dynamic client registration AFTER generating this redirect URL.
     const client_id = authUrl.searchParams.get("client_id");
+
+    // Store clientId internally for later use by clientInformation()
     if (client_id) {
-      authUrl.searchParams.append("state", client_id);
+      this._clientId_ = client_id;
```
threepointone: will this work across hibernation?

ghostwriternr: It doesn't, but I realised it doesn't need to. The client id is persisted into `cf_agents_mcp_servers` and gets restored after hibernation and the OAuth flow works perfectly even without this (this was not being stored here earlier anyway).

PR #525, opened 2025-09-29
add new messages instead of replacing them, add tests

whoiskatrin:
the fix is also considered tool calling, works as expected
Screenshot 2025-10-01 at 14 01 13


whoiskatrin:
tested with double and triple tools calls, works as expected

Issue #420, opened 2025-08-26
`useAgentChat`, when re-render with a different agent instance, does not update messages to the new instance


superxiao:
```typescript
  const agent = useAgent({
    agent: "Chat",
    name: threadId
  });

  const chat = useAgentChat({
    agent
  });
```
When switching to a new `threadId`, this code sometimes does not update the `chat.messages` to fetch for the new agent instance named with the new `threadId`. Specifically, it seems to work when the previous agent instance has not received new messages after load, but once a message is appended, switching to a new instance will retain the previous agent instance's messages. This makes implementing chat threads non-straightforward.

My current workaround is to unmount the chat UI that contains these hooks, and then remount with the new `threadId` as a prop, but that causes some unpleasant UI flashes. (I'm integrating `assistant-ui`, so has limited visual control there)

If this is an intentional optimization, would still be useful to have an option to refetch messages on name change.

threepointone:
not intentional, it's definitely a bug

whoiskatrin:
@superxiao do you mind giving it a go using this version of the package? `npm i https://pkg.pr.new/cloudflare/agents@531`

whoiskatrin:
@superxiao just bumping this up! thanks

superxiao:
I already built up a system around remounting, coupled with routers - but I managed to go back to a version without router and removed the remounting behavior. Yeah, I confirm that the agent and chat instances while staying mounted, does reload message history for different agent instance names (threads), using the version you provided. @whoiskatrin

PR #531, opened 2025-10-01
update our cache key in useAgentChat to include agent name (fix for #420)

whoiskatrin:
this fixes https://github.com/cloudflare/agents/issues/420

so previously when we switch from threadId-1 to threadId-2, we would use agentUrlString as the cache key, which in this cause would be identical for both agents and the cache would return the msgs for thread1, instead of thread2

wrote tests for the initial bug, failed as expected, added agent.name to the cacheKey and now all is passing

### DIFF
--- a/examples/resumable-stream-chat/src/app.tsx
+++ b/examples/resumable-stream-chat/src/app.tsx
@@ -1,0 +1,112 @@
+import type { UIMessage as Message } from "ai";
+import "./styles.css";
+import { useAgentChatHttp } from "agents/use-agent-chat-http";
+import { useCallback, useEffect, useRef, useState } from "react";
+
+export default function Chat() {
+  const [theme, setTheme] = useState<"dark" | "light">("dark");
+  const [input, setInput] = useState("");
+  const messagesEndRef = useRef<HTMLDivElement>(null);
+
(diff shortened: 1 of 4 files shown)


threepointone:
That's odd. Doesn't agentUrlString already include the name and id?

whoiskatrin:
> That's odd. Doesn't agentUrlString already include the name and id?

so here is my theory: when you call useAgent with a new name, it just mutates the same socket. It updates agent.name = newName, but the socket itself is still hanging on to the old _pkurl until it reconnects and PartySocket runs setWSProperties again. During that gap, if you check the URL, you’ll still see the old one (including the old room ID), so it doesn’t line up with the new thread. That stale URL is what the cache picks up, which is why you end up serving the previous conversation.

On the other hand, if you unmount and remount, the socket gets rebuilt from scratch, so _pkurl is set with the new room right away, that's why it's not always happening, I think.

I re-wrote the tests (initial ones had wrong URLs), tests these one again and I think this is the fix

That's my main theory, correct me if I'm wrong

whoiskatrin:
switching this to draft for now, gonna go dig into partysocket to see what's happening in there

Issue #528, opened 2025-09-30
Significant increase in cpu startup time


dnyg:
After adding the Cloudflare Agent `AIChatAgent` class and exporting it in my main file, running `wrangler deploy` often results in:

>[ERROR] Your Worker failed validation because it exceeded startup limits

I have used the Cloudflare tool `npx wrangler check startup` to generate a `worker-startup.cpuprofile` and I use this bash script to output the duration

```bash
jq -r '(.endTime - .startTime) / 1000' worker-startup.cpuprofile
```

My current worker has a startup time of 31ms without the agent, and over 200ms with the agent. I tried to use dynamic imports to defer the imports to runtime, and it improved the situration a bit, but it was still very high. Unfortunately, I can't defer the `AIChatAgent` class that my class extends.

I then created this very very minimal test repo: https://github.com/dnyg/cf-agent-test

Here I have an example of a worker with only "agents" as a dependency and a hardcoded fetch response along with an implementation of `AIChatAgent` where everything is default.

Without the DO this worker has a startup time of 13-14ms. With the worker it is 34ms (a big increase - 250% - but not as breaking as what I'm experiencing on my actual worker).

As I understand, it is not possible to point to the location of your Durable Object within your `wrangler.jsonc` file - it has to be exported in the main file.

I am not sure what exactly is causing my cpu startup time to skyrocket, but I assume it might be differences in versions of the `ai` and `zod` packages within my project, and agents.

Any ideas to what I can do to again be able to deploy? For now I will probably just move it to a seperate worker, and suffer the cpu startup time there.

threepointone:
Can you try setting build.minify as true in your vite config?

dnyg:
The `build.minify` is currently set to `esbuild`. Changing it to `true` didn't help a lot in my project. The minimal test repo I made doesn't use vite at all

threepointone:
Thanks for the repro, I'll have a look

threepointone:
investigating this here https://github.com/cloudflare/workers-sdk/issues/10723

dnyg:
I saw this PR that attempts to fix the issue, and actually does quite a lot of good things. But I have a comment on what is causing the remaining problems https://github.com/cloudflare/agents/pull/544#issuecomment-3384416863

The problem is definitely the `ajv` package, and it is used by `@modelcontextprotocol/sdk`

whoiskatrin:
@dnyg the changes should be in the latest version, let me know if there are any issues

PR #537, opened 2025-10-06
Replace imports of Zod with wildcard import

irvinebroque:
refs https://github.com/cloudflare/workers-sdk/issues/10723#issuecomment-3369017859, https://github.com/cloudflare/agents/issues/528, https://github.com/colinhacks/zod/issues/4637#issuecomment-2950064668

Wildcard import like this is tree-shakeable:

`import * as zod from "zod";`

This is not tress-shakeable:

`import { z } from "zod";`

Marking PR as draft then looking at examples to validate any impact on startup time and bundle size, make sure nothing breaks.


whoiskatrin:
@irvinebroque @threepointone I don't think this solves it. I've tested this and it doesn't do anything to startup time. It's still ~30-33ms using the package with this change, unless I'm missing something very obvious, but CPU profile didn't change.

threepointone (changes requested):
soft blocking

irvinebroque:
yep found the same — fine with not merging, wanted to try figuring out. I still don't get why this doesn't change it

PR #544, opened 2025-10-08
Startup time opt

whoiskatrin:


dnyg:
I think these are some good solid improvements, and the remaining problems all lie in the import of `@modelcontextprotocol/sdk` because it utilizes the `ajv` library that is taking up most of the startup time at this point
Screenshot 2025-10-09 at 09 01 50

packages/agents/src/mcp/client.ts
```diff
@@ -339,6 +339,17 @@ export class MCPClientManager {
    * @returns a set of tools that you can use with the AI SDK
    */
   getAITools(): ToolSet {
+    // Lazy-load jsonSchema from ai package to avoid pulling in the entire package at module load time
+    // This is loaded synchronously here (when getAITools is called), not at module import time
+    let jsonSchemaFn: ((schema: unknown) => unknown) | undefined;
```
threepointone:
```suggestion
    let jsonSchemaFn: typeof require("ai").jsonSchema | undefined;
```

maybe this i nstead?

threepointone: also we shouldn't be calling this everytime we call the function. should probably be a top level var/fn.

packages/agents/src/observability/index.ts
```diff
@@ -33,18 +32,24 @@ export const genericObservability: Observability = {
...
+    const indexModule = require("../index");
+    if (indexModule && typeof indexModule.getCurrentAgent === "function") {
+      const { request } = indexModule.getCurrentAgent();
+      if (request) {
+        const url = new URL(request.url);
+        localMode = url.hostname === "localhost";
+        return localMode;
+      }
+    }
+  } catch (_e) {}
```
threepointone: I would be surprised if this lazy import actually affects startup time, especially because it'll be the index module that imports observability

also the try catch makes it slower, and shouldn't be needed

packages/agents/src/ai-chat-agent.ts
```diff
@@ -22,6 +19,25 @@ import { autoTransformMessages } from "./ai-chat-v5-migration";
 
 const decoder = new TextDecoder();
 
+// Lazy-load AI SDK utilities
+type AIUtils = {
+  getToolName: (part: unknown) => string | undefined;
+  isToolUIPart: (part: unknown) => part is ToolUIPart;
+  parsePartialJson: (json: string) => Record<string, unknown>;
```
threepointone: give these the types of the actual functions

packages/agents/src/ai-chat-agent.ts
```diff
@@ -22,6 +19,25 @@ import { autoTransformMessages } from "./ai-chat-v5-migration";
 
 const decoder = new TextDecoder();
 
+// Lazy-load AI SDK utilities
+type AIUtils = {
```
threepointone:
```suggestion
type AISDK = {
```

packages/agents/src/ai-chat-agent.ts
```diff
@@ -22,6 +19,25 @@ import { autoTransformMessages } from "./ai-chat-v5-migration";
...
+  getToolName: (part: unknown) => string | undefined;
+  isToolUIPart: (part: unknown) => part is ToolUIPart;
+  parsePartialJson: (json: string) => Record<string, unknown>;
+};
+let aiUtils: AIUtils | undefined;
+function getAIUtils() {
```
threepointone:
```suggestion
function getAISDK() {
```

packages/agents/src/ai-chat-agent.ts
```diff
@@ -22,6 +19,25 @@ import { autoTransformMessages } from "./ai-chat-v5-migration";
...
+  if (!aiUtils) {
+    const ai = require("ai");
```
threepointone: btw, this will break DCE and import all of ai even if we don't use it. can you tell me how this affects bundle size?

packages/agents/src/index.ts
```diff
@@ -323,10 +324,14 @@ export class Agent<
...
-    "0.0.1"
-  );
+  private _mcp?: MCPClientManager;
+
+  get mcp(): MCPClientManager {
+    if (!this._mcp) {
+      this._mcp = new MCPClientManager(this._ParentClass.name, "0.0.1");
+    }
+    return this._mcp!; // Non-null assertion: we just initialized it above
+  }
```
threepointone: how does this affect startup time? I wouldn't think it would

packages/agents/src/lazy-loader.ts
```diff
@@ -0,0 +1,41 @@
+import type { MCPClientManager } from "./mcp/client";
```
threepointone: this file isn't referenced anywhere, what is it for

packages/agents/src/mcp/client.ts
whoiskatrin: require doesn't work in esbuild

threepointone: ? how so

whoiskatrin: i found a way, it's ok

packages/agents/src/index.ts
whoiskatrin: reverted

packages/agents/src/mcp/client.ts
```diff
@@ -357,17 +367,14 @@ export class MCPClientManager {
...
             outputSchema: tool.outputSchema
-              ? // @ts-expect-error drift between ai and mcp types
-                jsonSchema(tool.outputSchema)
+              ? getJsonSchema()!(tool.outputSchema as JSONSchema7)
               : undefined
           }
         ];
       })
-    );
+    ) as ToolSet;
```
threepointone: is this necessary now?

packages/agents/src/ai-chat-agent.ts
threepointone: please answer this question

packages/agents/src/ai-chat-agent.ts
```diff
@@ -22,6 +19,26 @@ import { autoTransformMessages } from "./ai-chat-v5-migration";
...
+type AISDK = {
+  getToolName: typeof import("ai").getToolName;
+  isToolUIPart: typeof import("ai").isToolUIPart;
+  parsePartialJson: typeof import("ai").parsePartialJson;
+};
+let aiSDK: AISDK | undefined;
+function getAISDK() {
+  if (!aiSDK) {
+    // Use destructuring to help with tree-shaking
+    const { getToolName, isToolUIPart, parsePartialJson } = require("ai");
```
threepointone, line 32: please measure and tell me whether ans how this affects bundle size

packages/agents/src/index.ts
```diff
@@ -2,6 +2,8 @@ import type { env } from "cloudflare:workers";
 import { AsyncLocalStorage } from "node:async_hooks";
 import type { Client } from "@modelcontextprotocol/sdk/client/index.js";
 import type { SSEClientTransportOptions } from "@modelcontextprotocol/sdk/client/sse.js";
+import { MCPClientManager } from "./mcp/client";
```
threepointone: revert the changes in this file? doesn't see, to have anything to do with load time anymore

packages/agents/src/mcp/client.ts
whoiskatrin: reverted, not needed

packages/agents/src/ai-chat-agent.ts
whoiskatrin: +43 bytes, it's not too bad, but it's critical for startup time

packages/agents/src/index.ts
whoiskatrin: done

packages/agents/src/ai-chat-agent.ts
```diff
@@ -729,7 +756,7 @@ export class AIChatAgent<Env = unknown, State = unknown> extends Agent<
 
                         updateToolPart({
                           toolCallId: data.toolCallId,
-                          toolName: getToolName(toolInvocation),
+                          toolName: getToolName(toolInvocation) || "",
```
threepointone: this makes me nervous, using a blank string as a tool name

whoiskatrin: this has been resolved, removed

packages/agents/src/ai-chat-agent.ts
```diff
@@ -775,7 +804,7 @@ export class AIChatAgent<Env = unknown, State = unknown> extends Agent<
                           throw new Error("Tool invocation not found");
                         updateToolPart({
                           toolCallId: data.toolCallId,
-                          toolName: getToolName(toolInvocation),
+                          toolName: getToolName(toolInvocation) || "",
```
threepointone: here as well

whoiskatrin: done

PR #550, opened 2025-10-09
encode MCP message headers with Base64

### LINKED ISSUE #549: json-rpc request stringified and placed into cf-mcp-message HTTP header
**Describe the bug**

agents/mcp places JSON-RPC request payloads into the cf-mcp-message HTTP header when bridging WebSocket connections to Durable Objects. This causes failures when:
- the payload contains Unicode characters > 0xFF (violates Fetch API specification)
- [ossibly when payload size approaches certain thresholds in Cloudflare Workers runtime (even with valid ASCII)

I came across this issue when building an mcp server with workers and durable objects for a ChatGPT app. 

**Exact error message** (location is not personal in case of a privacy worry - I'm using Cloudflare WARP <3) 
✘ [ERROR] Invalid header name or value: "[{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"_meta":{"openai/userAgent":"ChatGPT/1.2025.273 (iOS 18.6.2; iPhone17,3; build 18210387024)","openai/locale":"en-GB","openai/userLocation":{"city":"Plymouth","region":"England","country":"GB","timezone":"Europe/London","latitude":"50.37153","longitude":"-4.14305"},"openai/subject":"v1/wden1N8mNaYHMiKUgYbJGh2oJynnyAQ1y8iEIlmElVK93MGvGv6dsw"},"name":"todo.create_task","arguments":{"columnId":"in-progress","title":"Cancel gym membership","note":"Steps: 1) Find contract & cancellation policy. 2) Get gym’s phone/email and member ID. 3) Submit written notice or call to cancel. 4) Ask for confirmation in writing and any prorated refund. 5) Return key fob if required.","statusId":"in-progress"}}}]". Per the Fetch specification, the Headers class may only accept header names and values which contain 8-bit characters. That is, they must not contain any Unicode code points greater than 0xFF. As a quirk, we are encoding this string in UTF-8 in the header, but in a browser this would result in a TypeError exception. Consider encoding this string in ASCII for compatibility with browser implementations of the Fetch specification.

**Root Cause(as far as I can see, I just started using the library very recently)**:
packages/agents/src/mcp/utils.ts:231
createStreamingHttpHandler (utils.ts:227-234) creates a Request to forward to the Durable Object
It places the entire message array in cf-mcp-message header
Any char > 0xFF causes immediate failure in Workers runtime while it passes in Node runtime. 

**To Reproduce**
Long way: Creating an MCP agent using McpAgent.serve() with streamable-http transport, then creating the connection in ChatGPT and asking it to send a very long message to server . https://github.com/openai/openai-apps-sdk-examples/issues - replacing node server with agents/mcp should be the quickest way. I'm going to open source my own repo later today/tomorrow and will update here.

for now minimal reproduction can be possible in this way:

```
// In Cloudflare Workers runtime (not Node.js):
const payload = JSON.stringify([{
  jsonrpc: "2.0",
  id: 1,
  method: "tools/call",
  params: {
    name: "test",
    arguments: { note: "Unicode bullet: • and em dash: —" }
  }
}]);

// This fails in Workers but succeeds in Node.js:
const req = new Request('http://cloudflare.com', {
  headers: { 'cf-mcp-message': payload }
});
```

**Expected behavior**
Handling of long and arbitrary JSON-RPC payloads:
Unicode characters (emoji, international characters, typographic symbols)
Large payloads (within reasonable limits)

**Version:**
0.2.9

ainergiz:
encode MCP message headers with Base64 to support unicode characters

[**Fixes #549**](https://github.com/cloudflare/agents/issues/549)

brief description - detailed explanation is in the issue: 
JSON-RPC payloads with Unicode characters (emoji, em dashes, smart quotes) fail in HTTP headers. This happened when creating a ChatGPT app with agents library because ChatGPT sends requests with Unicode characters.

**Solution:**
- Base64 encode `cf-mcp-message` header content
- Add `cf-mcp-message-encoding` header to indicate encoding type
- Maintain backward compatibility for unencoded messages

**Changes:**
- `src/mcp/utils.ts`: encodes outgoing messages
- `src/mcp/index.ts`: decodes incoming messages with fallback
- `src/mcp/transport.ts`: clears encoding header
- `src/tests/mcp/transports/streamable-http.test.ts`: added Unicode test

**Trade-offs:**
- (small one with solution) Size increase: ~35% (e.g., 730 bytes → 971 bytes), still well within 8KB header limits
- future optimisation related to this increase: could use [Node.js's brand new `zlib`](https://developers.cloudflare.com/workers/runtime-apis/nodejs/zlib/) module for compression even if header size is not a concern 

End to end test: tested the solution by tarballing my fork into existing mcp server and it worked with previous messages that has unicode characters

### DIFF
--- a/packages/agents/src/mcp/index.ts
+++ b/packages/agents/src/mcp/index.ts
@@ -151,7 +151,7 @@ export abstract class McpAgent<
         if (this._transport instanceof StreamableHTTPServerTransport) {
           switch (req.headers.get(MCP_HTTP_METHOD_HEADER)) {
             case "POST": {
-              // This returns the repsonse directly to the client
+              // This returns the response directly to the client
               const payloadHeader = req.headers.get(MCP_MESSAGE_HEADER);
               let rawPayload: string;
 


packages/agents/src/mcp/index.ts
```diff
@@ -17,7 +17,9 @@ import {
   handleCORS,
   isDurableObjectNamespace,
   MCP_HTTP_METHOD_HEADER,
-  MCP_MESSAGE_HEADER
+  MCP_MESSAGE_ENCODING_HEADER,
```
deathbyknowledge: We can remove this, backwards compatibility should never be an issue here. If you upgrade your version of `agents` both the handler that sets the header and the handler that reads the header update at the same time, nothing else is involved.

packages/agents/src/mcp/index.ts
```diff
@@ -153,7 +155,43 @@ export abstract class McpAgent<
...
-              const parsedBody = await JSON.parse(payloadHeader ?? "{}");
+              const encodingHeader = req.headers.get(
+                MCP_MESSAGE_ENCODING_HEADER
+              );
+              let rawPayload: string;
+
+              // fallback to empty object if no payload header
+              if (!payloadHeader) {
+                rawPayload = "{}";
+              } else if (!encodingHeader) {
```
deathbyknowledge: Which means we could remove most of this too

packages/agents/src/mcp/transport.ts
```diff
@@ -230,6 +234,7 @@ export class StreamableHTTPServerTransport implements Transport {
     // Remove headers that are not part of the original request
     delete requestInfo.headers[MCP_HTTP_METHOD_HEADER];
     delete requestInfo.headers[MCP_MESSAGE_HEADER];
+    delete requestInfo.headers[MCP_MESSAGE_ENCODING_HEADER];
```
deathbyknowledge: Same here

packages/agents/src/mcp/utils.ts
```diff
@@ -25,7 +25,14 @@ export const MCP_HTTP_METHOD_HEADER = "cf-mcp-method";
  */
 export const MCP_MESSAGE_HEADER = "cf-mcp-message";
 
-const MAXIMUM_MESSAGE_SIZE_BYTES = 4 * 1024 * 1024; // 4MB
+/**
+ * Indicates how the MCP message was encoded before being stuffed into a header.
+ */
+export const MCP_MESSAGE_ENCODING_HEADER = "cf-mcp-message-encoding";
```
deathbyknowledge: Same here

packages/agents/src/mcp/utils.ts
```diff
@@ -224,11 +231,23 @@ export const createStreamingHttpHandler = (
         request.headers.forEach((value, key) => {
           existingHeaders[key] = value;
         });
+
+        // Remove MCP-specific headers to prevent stale/corrupted data from
```
deathbyknowledge: Was this ever an issue? When does it happen? The header is not exposed to the public, is just for internal routing. Plus, it would get overwritten in line 248, should be safe to remove

packages/agents/src/mcp/index.ts
ainergiz: Thank you for the comments! I'm still in REST API world and had a bit confusion about internal, external exposure. extra encoding header naming and unnecessary checks removed now

packages/agents/src/mcp/index.ts
ainergiz: yep removed, keeping empty object check

packages/agents/src/mcp/transport.ts
ainergiz: removed

packages/agents/src/mcp/utils.ts
ainergiz: cleaned up

packages/agents/src/mcp/utils.ts
ainergiz: agreed, spread is going to get rid of it even if it was an issue

packages/agents/src/mcp/index.ts
```diff
@@ -153,7 +154,32 @@ export abstract class McpAgent<
...
+                    "utf-8"
+                  );
+                } catch (_error) {
+                  throw new Error(
+                    "Internal Server Error: Failed to decode MCP message header"
+                  );
+                }
+              }
+
+              if (
```
deathbyknowledge: This is not needed since we're already doing this check in the Worker handler, so we can just keep the `if/else` above

packages/agents/src/mcp/utils.ts
```diff
@@ -25,7 +25,7 @@ export const MCP_HTTP_METHOD_HEADER = "cf-mcp-method";
  */
 export const MCP_MESSAGE_HEADER = "cf-mcp-message";
 
-const MAXIMUM_MESSAGE_SIZE_BYTES = 4 * 1024 * 1024; // 4MB
+export const MAXIMUM_MESSAGE_SIZE_BYTES = 4 * 1024 * 1024; // 4MB
```
deathbyknowledge: Not needed

ainergiz:
thanks! I've just removed the message size check, hopefully should be good to go @deathbyknowledge

packages/agents/src/mcp/index.ts
```diff
@@ -153,7 +153,23 @@ export abstract class McpAgent<
             case "POST": {
               // This returns the repsonse directly to the client
```
```suggestion
              // This returns the response directly to the client
```

PR #554, opened 2025-10-10
use tsdown, update deps, slim bundles

threepointone:
we were bundling external modules into the main release, fixed that 
also tsdown is just faster (and maintained)


threepointone:
In CI 

`npm run build` goes from ~20 seconds to ~10 seconds

threepointone:
package size whent from 922kb to 217 kb (excluding sourcemaps)

packages/agents/scripts/build.ts
```diff
@@ -15,19 +15,10 @@ async function main() {
       "src/observability/index.ts",
       "src/codemode/ai.ts"
     ],
-    external: [
-      "cloudflare:workers",
-      "cloudflare:email",
-      "@ai-sdk/react",
-      "ai",
```
ericclemmons, line 22: Looking into this.

I believe this change is causing a Vite bundling issue that wasn't present prior.


Image

Clearing `node_modules/.vite` and such to figure out if it's related to this or something in the project.

(I _did_ run into my own set of issues with `tsdown` that caused me to go back to https://github.com/colinhacks/zshy)

ericclemmons: Yes, I had to go back to `0.2.9` for my app to continue working.

Image

I'll see if I can narrow down the difference in code between the two bundles...

threepointone: fixed in 0.2.12

PR #565, opened 2025-10-13
RPC transport

### LINKED ISSUE #548: feat: define Agent and McpAgent in the same worker
We want to be able to make a single worker that defines an Agent and an McpAgent, and the agent connects to the server (with `this.mcp.connect()`)

Problems: 

1. We actually want it to use a service binding to connect, but the mcp.connect api expects as URL (and there might some mcp transport constraints where it assumes fetch?)
2. Internal "fetch" won't work without the `global_fetch_strictly_public` flag (also it's annoying to setup the domains/host to connect to when the binding is right there)

we should solve for 1.

mattzcarey:
## RPC Transport for MCP

Enables an Agent to connect to an McpAgent via Durable Object RPC bindings — no HTTP, no network overhead. Both can live in the same Worker.

Closes #548

### User-facing API

```typescript
// Pass the DO namespace directly — TypeScript discriminates string vs namespace
await this.addMcpServer("counter", env.MY_MCP);

// With props for user context
await this.addMcpServer("counter", env.MY_MCP, {
  props: { userId: "user-123", role: "admin" }
});

// HTTP still works exactly as before
await this.addMcpServer("github", "https://mcp.github.com");
```

`addMcpServer` now has proper TypeScript overloads so the RPC and HTTP paths are type-safe. `AddRpcMcpServerOptions` (with `props`) is separate from `AddMcpServerOptions` (with `callbackHost`, `transport`, etc.) — you cannot accidentally pass HTTP options to an RPC binding or vice versa.

### Architecture

```
Agent (MCP client)                    McpAgent (MCP server)
  │                                      │
  │ this.addMcpServer("name", env.MCP)   │
  │                                      │
  ├─► RPCClientTransport                 │
  │     │                                │
  │     │ getServerByName(namespace,      │
  │     │   "rpc:name", { props })       │
  │     │                                │
  │     │ stub.handleMcpMessage(msg) ──► handleMcpMessage()
  │     │                            │   │
  │     │                            │   ├─► RPCServerTransport
  │     │                            │   │     .handle(msg)
  │     │                            │   │     → onmessage → MCP Server
  │     │                            │   │     ← send() ← MCP Server
  │     │                            │   │
  │     │ ◄── response ─────────────┘   │
  │     │                                │
  │     ▼ onmessage(response)            │
  │                                      │
  ├─► MCPClientManager                   │
  │     .connect() / .discoverIfConnected()
  │                                      │
  └─► tools available via getAITools()   │
```

### Design decisions

**RPCClientTransport accepts a namespace, not a stub.** It creates the stub internally during `start()` via `getServerByName` from partyserver, which handles `setName` + props in one call. The transport is a thin wrapper: `start()` creates the stub, `send()` calls `stub.handleMcpMessage(msg)` and routes responses to `onmessage`.

**RPCServerTransport is minimal.** Dropped session management (unnecessary — RPC within a DO is inherently session-scoped). Dropped 170 lines of hand-written JSON-RPC validation in favor of `JSONRPCMessageSchema.parse()` from the MCP SDK. What remains: `handle()` receives a message, calls `onmessage`, waits for `send()` via a microtask-batched promise, returns the response.

**Hibernation works automatically.** When `addMcpServer` is called with an RPC binding, the binding name (looked up by scanning `env`) and props are persisted to the `cf_agents_mcp_servers` table. On wake-up, `_restoreRpcMcpServers()` reads these rows, looks up `env[bindingName]`, and reconnects with the same ID. This matches the existing HTTP restore path. Depends on partyserver 0.3.0 which persists `this.name` to DO storage so it survives RPC-triggered wake-ups.

**Deduplication by server name.** Calling `addMcpServer` with the same name returns the existing connection instead of creating duplicates. This applies to both RPC and HTTP. Connection IDs are stable across hibernation (passed via `reconnect: { id }` to `mcp.connect()`).

**No changes to the Agent base class beyond `addMcpServer`.** The original PR added `_resolveRpcBinding`, `_buildRpcTransportOptions`, `_buildHttpTransportOptions`, and `_connectToMcpServerInternal` to the ~4300-line `index.ts`. All removed. The RPC path in `addMcpServer` is ~30 lines that create transport options and delegate to `mcp.connect()`.

### Files changed

| File | What changed |
|------|-------------|
| `packages/agents/src/mcp/rpc.ts` | Rewritten. 609→245 lines. New `RPCClientTransport` (namespace-based), simplified `RPCServerTransport`, `RPC_DO_PREFIX` constant |
| `packages/agents/src/index.ts` | `addMcpServer` overloads for RPC/HTTP, `AddRpcMcpServerOptions` type, name-based dedup, `_restoreRpcMcpServers()`, `_findBindingNameForNamespace()` |
| `packages/agents/src/mcp/client.ts` | `getRpcServersFromStorage()`, `saveRpcServerToStorage()`, `restoreConnectionsFromStorage` skips `rpc:` URLs, fixed `callTool` stripping `serverId` |
| `packages/agents/src/mcp/client-connection.ts` | RPC transport in `MCPTransportOptions` union, `getTransport("rpc")`, auth probe rejects RPC |
| `packages/agents/src/mcp/types.ts` | 108→26 lines. Removed RPC connection config types (now internal to transport) |
| `packages/agents/src/mcp/index.ts` | Updated exports, `RPC_DO_PREFIX` |
| `examples/mcp-rpc-transport/` | Full rewrite: Workers AI (no API keys), Kumo/agents-ui, Tailwind |
| `docs/mcp-transports.md` | New documentation for RPC transport |

### Test coverage

**Unit tests** (`tests/mcp/transports/rpc.test.ts` — 21 tests):
- RPCClientTransport lifecycle, messaging, error handling, validation
- RPCServerTransport request/response, batching, notifications, timeout
- End-to-end via McpAgent (tool listing, tool calling, sequential requests)

**E2e tests** (`tests/mcp/add-rpc-mcp-server.test.ts` — 7 tests):
- Tool discovery via RPC binding
- Tool execution with correct response content
- Storage persistence (binding name + props saved)
- Hibernation restore with stable connection ID
- Deduplication (same name → same connection)
- Props passing end-to-end
- Server removal (connection + storage cleanup)

### Notes for reviewers

- The `as unknown as DurableObjectNamespace<McpAgent>` cast in `addMcpServer` is needed because `DurableObjectNamespace<T>` where `T extends McpAgent` is not directly assignable to `DurableObjectNamespace<McpAgent>` due to generic variance. The cast is safe since `T` is a subclass.
- `RPCClientTransportOptions` has a generic `<T extends McpAgent>` for the namespace type, but the class itself stores `DurableObjectNamespace<McpAgent>` (the base type) to avoid the infinite type recursion that `DurableObjectNamespace<any>` causes with `getServerByName`.
- The deprecated `mcp.connect()` method is used intentionally for RPC connections because `registerServer()` saves transport options to storage via `JSON.stringify`, and `DurableObjectNamespace` is not serializable. RPC persistence is handled separately via `saveRpcServerToStorage`.
- Requires partyserver 0.3.0+ which persists `this.name` to DO storage, fixing the "Attempting to read .name before it was set" error on RPC-triggered wake-ups.

### DIFF
--- a/examples/mcp-rpc-transport/src/server.ts
+++ b/examples/mcp-rpc-transport/src/server.ts
@@ -49,15 +49,9 @@ export class MyMCP extends McpAgent<Env, State, Props> {
 
 export class Chat extends AIChatAgent<Env> {
   async onStart() {
-    console.log(
-      `[Chat.onStart] called, name="${this.name}", connections=${Object.keys(this.mcp.mcpConnections).length}`
-    );
     await this.addMcpServer("test-server", this.env.MyMCP, {
       props: { userId: "demo-user-123", role: "admin" }
     });
-    console.log(
-      `[Chat.onStart] after addMcpServer, connections=${Object.keys(this.mcp.mcpConnections).length}, tools=${this.mcp.listTools().length}`
-    );
   }
 
   async onChatMessage() {
--- a/packages/agents/src/index.ts
+++ b/packages/agents/src/index.ts
@@ -1131,15 +1131,8 @@ export class Agent<
         },
         async () => {
           await this._tryCatch(async () => {
-            console.log(
-              `[Agent.onStart] restoring connections, name="${this.name}"`
-            );
             await this.mcp.restoreConnectionsFromStorage(this.name);
-            console.log(`[Agent.onStart] restoring RPC MCP servers`);
             await this._restoreRpcMcpServers();
-            console.log(
-              `[Agent.onStart] restore complete, connections=${Object.keys(this.mcp.mcpConnections).length}`
-            );
             this.broadcastMcpServers();
 
             this._checkOrphanedWorkflows();
@@ -3555,13 +3548,7 @@ export class Agent<
 
   private async _restoreRpcMcpServers(): Promise<void> {
     const rpcServers = this.mcp.getRpcServersFromStorage();
-    console.log(
-      `[_restoreRpcMcpServers] found ${rpcServers.length} RPC server(s) in storage`
-    );
     for (const server of rpcServers) {
-      console.log(
-        `[_restoreRpcMcpServers] server="${server.name}" id="${server.id}" alreadyConnected=${!!this.mcp.mcpConnections[server.id]}`
-      );
       if (this.mcp.mcpConnections[server.id]) {
         continue;
       }
@@ -3579,7 +3566,7 @@ export class Agent<
         continue;
       }
 
-      const normalizedName = server.server_url.replace("rpc:", "");
+      const normalizedName = server.server_url.replace(RPC_DO_PREFIX, "");
 
       try {
         await this.mcp.connect(`${RPC_DO_PREFIX}${normalizedName}`, {
@@ -3854,9 +3841,6 @@ export class Agent<
     const existingServer = this.mcp
       .listServers()
       .find((s) => s.name === serverName);
-    console.log(
-      `[addMcpServer] name="${serverName}", existingInStorage=${!!existingServer}, existingInMemory=${!!(existingServer && this.mcp.mcpConnections[existingServer?.id ?? ""])}`
-    );
     if (existingServer && this.mcp.mcpConnections[existingServer.id]) {
       const conn = this.mcp.mcpConnections[existingServer.id];
       if (
@@ -3868,6 +3852,11 @@ export class Agent<
           state: MCPConnectionState.AUTHENTICATING,
           authUrl: conn.options.transport.authProvider.authUrl
         };
+      }
+      if (conn.connectionState === MCPConnectionState.FAILED) {
+        throw new Error(
+          `MCP server "${serverName}" is in failed state: ${conn.connectionError}`
+        );
       }
       return { id: existingServer.id, state: MCPConnectionState.READY };
     }
--- a/packages/agents/src/mcp/rpc.ts
+++ b/packages/agents/src/mcp/rpc.ts
@@ -60,13 +60,9 @@ export class RPCClientTransport implements Transport {
     }
 
     const doName = `${RPC_DO_PREFIX}${this._name}`;
-    console.log(
-      `[RPCClientTransport] start: calling getServerByName("${doName}")`
-    );
     this._stub = await getServerByName(this._namespace, doName, {
       props: this._props
     });
-    console.log(`[RPCClientTransport] start: stub obtained for "${doName}"`);
 
     this._started = true;
   }
@@ -86,15 +82,8 @@ export class RPCClientTransport implements Transport {
     }
 
     try {
-      const msgSummary = Array.isArray(message)
-        ? `batch[${message.length}]`
-        : ((message as { method?: string }).method ?? "response");
-      console.log(
-        `[RPCClientTransport] send: calling handleMcpMessage("${msgSummary}")`
-      );
       const result: JSONRPCMessage | JSONRPCMessage[] | undefined =
         await this._stub.handleMcpMessage(message);
-      console.log(`[RPCClientTransport] send: got result for "${msgSummary}"`);
 
       if (!result) {
         return;
--- a/.changeset/rpc-transport-rewrite.md
+++ b/.changeset/rpc-transport-rewrite.md
@@ -1,0 +1,44 @@
+---
+"agents": minor
+---
+
+Add RPC transport for MCP: connect an Agent to an McpAgent via Durable Object bindings
+
+**New feature: `addMcpServer` with DO binding**
+
+Agents can now connect to McpAgent instances in the same Worker using RPC transport — no HTTP, no network overhead. Pass the Durable Object namespace directly:
+
+```typescript
+// In your Agent
+await this.addMcpServer("counter", env.MY_MCP);
+
+// With props
+await this.addMcpServer("counter", env.MY_MCP, {
+  props: { userId: "user-123", role: "admin" }
+});
+```
+
+The `addMcpServer` method now accepts `string | DurableObjectNamespace` as the second parameter with proper TypeScript overloads, so HTTP and RPC paths are type-safe and cannot be mixed.
+
+**Hibernation support**
+
+RPC connections survive Durable Object hibernation automatically. The binding name and props are persisted to storage and restored on wake-up, matching the behavior of HTTP MCP connections. No need to manually re-establish connections in `onStart()`.
+
+**Deduplication**
+
+Calling `addMcpServer` with the same server name multiple times (e.g., across hibernation cycles) now returns the existing connection instead of creating duplicates. This applies to both RPC and HTTP connections. Connection IDs are stable across hibernation restore.
+
+**Other changes**
+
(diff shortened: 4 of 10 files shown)


examples/rpc-transport/README.md
```diff
@@ -0,0 +1 @@
+# RPC Transport demo
```
whoiskatrin: can we have a proper readme here?

whoiskatrin: also need .md for RPC transport in the docs folder

examples/mcp-rpc-transport/src/server.ts
```diff
@@ -0,0 +1,136 @@
...
+  }
+}
+
+const model = openai("gpt-4o-2024-11-20");
+
+/**
+ * Chat Agent implementation that handles real-time AI chat interactions
+ */
+export class Chat extends AIChatAgent<Env> {
+  async onStart(): Promise<void> {
```
whoiskatrin: If MCP connection fails during startup, will the agent will be in an undefined state with no error handling?

packages/agents/src/mcp/rpc.ts
```diff
@@ -0,0 +1,143 @@
...
+    if (!this._pendingResponse) {
+      this._pendingResponse = message;
+    } else if (Array.isArray(this._pendingResponse)) {
+      this._pendingResponse.push(message);
+    } else {
+      this._pendingResponse = [this._pendingResponse, message];
+    }
+  }
+
+  async handle(
```
whoiskatrin, line 173: If onmessage handler never calls send(), this will hang forever? Should we add a timeout?

packages/agents/src/mcp/rpc.ts
```diff
@@ -0,0 +1,534 @@
...
+      this._pendingResponse.push(message);
+    } else {
+      this._pendingResponse = [this._pendingResponse, message];
+    }
+
+    // Resolve the promise on the next tick to allow multiple sends to accumulate
+    if (this._responseResolver) {
+      const resolver = this._responseResolver;
+      this._responseResolver = null;
+      // Use queueMicrotask to allow additional send() calls to accumulate
```
whoiskatrin: if this _responseResolver = null doesn't it mean that all subsequent send() calls lose their resolver?

examples/mcp-rpc-transport/src/server.ts
mattzcarey: fair. in practise it would probably be a `@callable` which registers the server from the frontend. I did this for brevity. adding a note in the doc

packages/agents/src/mcp/rpc.ts
mattzcarey: I dont think so. first send() schedules the resolver, subsequent calls just batch messages before the microtask fires and they are all resolved.

packages/agents/src/mcp/rpc.ts
mattzcarey: added a 60 second timeout configurable in the transport.

mattzcarey:
there is a bug with restarts of the `McpAgent` and loosing the tool registrations.

jpm-halfspace:
How far is this from being merged?

whoiskatrin:
> How far is this from being merged?

not far, it will go though another round of reviews today

packages/agents/src/mcp/rpc.ts
```diff
@@ -0,0 +1,608 @@
...
+      validateJSONRPCMessage(batch[i]);
+    } catch (error) {
+      throw new Error(
+        `Invalid JSON-RPC batch: message at index ${i} is invalid: ${error instanceof Error ? error.message : String(error)}`
+      );
+    }
+  }
+
+  return true;
+}
```
mattzcarey: this should probs be zod schema as MCP sdk does.

packages/agents/src/mcp/rpc.ts
```diff
@@ -0,0 +1,608 @@
...
+    if (Array.isArray(message)) {
+      validateJSONRPCBatch(message);
+    } else {
+      validateJSONRPCMessage(message);
+    }
+
+    // Set the name if the stub is a DO
+    if (this._doName && this._stub.setName) {
+      await this._stub.setName(this._doName);
+    }
```
mattzcarey: I dont want to do this in the transport since it could be a worker to worker binding...

packages/agents/src/mcp/rpc.ts
```diff
@@ -0,0 +1,608 @@
...
+
+    // Initialize props on first send
+    if (this._props && !this._propsInitialized && this._stub.updateProps) {
+      await this._stub.updateProps(this._props);
+      this._propsInitialized = true;
+    }
```
mattzcarey: same here. we need a better/first class way to pass props through the transport.

packages/agents/src/index.ts
```diff
@@ -665,18 +678,19 @@ export class Agent<
...
-                  {
+                  reconnect: {
                     id: server.id,
                     oauthClientId: server.client_id ?? undefined
                   }
-                )
+                })
                   .then(() => {
                     // Broadcast updated MCP servers state after each server connects
                     this.broadcastMcpServers();
```
mattzcarey: this is the biggest change here. I'm not sure I like it. 

Why is the main agent class caring about this. it should all be added to the client manager.

packages/agents/src/index.ts
```diff
@@ -1405,29 +1419,148 @@ export class Agent<
   }
 
   /**
-   * Connect to a new MCP Server
+   * Validates and resolves a Durable Object binding from env
+   * @returns The namespace and binding name for storage
+   */
+  private _resolveRpcBinding<T extends McpAgent>(
```
mattzcarey: also grim to have this at the top level agent class. this is an implementation detail of the rpc transport. And actually a bad one at that.

packages/agents/src/index.ts
```diff
@@ -1405,29 +1419,148 @@ export class Agent<
...
+   * await agent.addMcpServer("my-server", env.MyMCP, {
+   *   transport: { type: "rpc" },
+   *   props: { userId: "user-123", role: "admin" }
+   * });
+   *
+   * // HTTP/SSE transport
+   * await agent.addMcpServer("my-server", "https://example.com/mcp", "https://my-app.com");
+   * ```
    */
+  async addMcpServer<
```
mattzcarey: the lifecycle of calling this needs rethinking. I initially thought ahh we add this to onStart and then all is good. But that means you have v little control on when the mcp server is connected and when the auth happens etc. Also there is almost no error handling.

Need to think about this a bit more but maybe actually we deprecate this `addMcpServer` and make the lower level `McpClientManager` easier to use. If we do that then this change is actually not necessary as the example can directly use the client manager.

threepointone:
I took a crack at this and rewrote it all, with tests and such. I also managed to fix some other gnarly stuff along the way (of note, this.name is now available inside alarms, because of enhancements to partyserver). 

I (and by I I mean cursor) also rewrote the PR description. 

Quite happy, I think this is ready to land. I might just land it tonight.

PR #566, opened 2025-10-14
fix: oauth2 client flow

mattzcarey:
DO shuts down during the flow and loses mcpConnections map and the callback url array. Shows in this error..

Best server to test this with is `https://observability.mcp.cloudflare.com/mcp`

Screenshot 2025-10-14 at 18 02 44

### DIFF
--- a/.changeset/light-readers-jam.md
+++ b/.changeset/light-readers-jam.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
+fix: Oauth2 client flow


packages/agents/src/index.ts
```diff
@@ -1458,32 +1461,107 @@ export class Agent<
...
+          server.server_url,
+          server.callback_url,
+          server.server_options ? JSON.parse(server.server_options) : undefined,
+          {
+            id: server.id,
+            oauthClientId: server.client_id ?? undefined
+          }
+        );
+      } catch (error) {
+        console.error(`Failed to restore connection for ${serverId}:`, error);
```
threepointone: hmm should it be a silent error? what can we do better here?

mattzcarey: done better

whoiskatrin (approved):

Issue #570, opened 2025-10-16
Agents example scheduleTask tool doesn't work. `getCurrentAgent` returns undefined


eastlondoner:
**Describe the bug**
The schedule task tool in agents-starter doesn't work because `getCurrentAgent` returns undefined

**To Reproduce**

Run `npm create cloudflare@latest agents-starter -- --template=cloudflare/agents-starter`
- say no when it asks you to deploy
- update agents to `0.2.13`
- run `npm start`
- ask it to schedule a task

<img width="1426" height="955" alt="Image" src="https://github.com/user-attachments/assets/385d6f9f-9c5f-487b-ba5c-ca4377d93fca" />

**Expected behavior**
tool works, getCurrentAgent is defined

**Screenshots**
see above

**Version:**
latest 0.2.13

whoiskatrin:
Looking into this

eastlondoner:
Deleted node_modules re ran `npm install` and it's fixed

eastlondoner:
So it turns out that vite stores caches in `node_modules/.vite` and vite can decide for its own reasons to bundle the module multiple times, which results in multiple copies of the `agentContext` and depending on which vite is using you can see the issue originally described.

<img width="720" height="499" alt="Image" src="https://github.com/user-attachments/assets/80559a03-0584-421b-bd0d-e34ed30bb59e" />

<img width="720" height="369" alt="Image" src="https://github.com/user-attachments/assets/056b4b8e-388d-47b9-b83d-02b2ff765b43" />

threepointone:
I gotta close this. there's not much we specifically can do, it's a bundler problem. sorry :(

PR #574, opened 2025-10-19
Add resolve.dedupe: ['agents'] to all vite.config.ts files

eastlondoner:
This fixes weird and annoying issues that can arise if vite decides to bundle the agents import multiple times.

FYI I'd also recommend adding `react` to this array as well but keeping the change very specific.

This addresses this issue https://github.com/cloudflare/agents/issues/570


threepointone:
There needs to be a better solution here, we can't expect our users to do this in all their apps. Seems like something funny going on with vite

eastlondoner:
yeah you're totally right.

I really don't know the ins and outs of vite bundling tbh. For sure you can wait and see if other people run into this (seems to me it's things like monorepos, packages declaring `agents` as a dependency that will cause it).

I think if you want to do something about it the nice thing to do is to handle it inside of the cloudflare vite plugin

eastlondoner:
or see if there's an alternative to using that module-singleton asyncLocalStorage - although I don't know what that would be

eastlondoner:
For example If the async local storage instance was attached to the cloudflare:workers virtual module or something inside it like Cloudflare env then it would be a true singleton regardless of module resolution.

I think you can also do stuff with global which is fine but kinda yucky and leaks more

whoiskatrin:
Closing for now, we will get back to this soon, thank you

PR #604, opened 2025-10-28
update chess mcp to use stateless worker

deathbyknowledge:
Replace the Chess App's `McpAgent` with `createMcpHandler` (plus a few nits).

Works exactly the same but it's less LoC and doesn't require an extra DO export.


openai-sdk/chess-app/src/index.ts
```diff
@@ -1,73 +1,52 @@
...
-            {
-              type: "text",
-              text: "Successfully rendered chess game menu"
-            }
-          ],
-          structuredContent: {}
-        };
+        uri: "ui://widget/index.html",
+        mimeType: "text/html+skybridge",
+        text: await (await env.ASSETS.fetch("http://localhost/")).text()
```
threepointone: can we inject the domain here into the html, so it can be picked up by the useAgent up there, so it doesn't have to be hardcoded?

openai-sdk/chess-app/src/index.ts
```diff
@@ -1,73 +1,63 @@
...
+import { env } from "cloudflare:workers";
 
-// Adapted from https://developers.openai.com/apps-sdk/build/examples
-export class McpWidgetAgent extends McpAgent<Env> {
-  server = new McpServer({ name: "Pizzaz", version: "v1.0.0" });
+const getWidgetHtml = async () => {
+  let html = await (await env.ASSETS.fetch("http://localhost/")).text();
+  html = html.replace(
+    "<!--RUNTIME_CONFIG-->",
+    `<script>window.HOST = \`${env.HOST}\`;</script>`
```
threepointone: no no, I was thinking we use const {request} = getCurrentAgent() and pull the domain out from there!

threepointone (approved):
stamping to unblock, but look at my comment! 

this is dope. oh man we're going to use this new mcp handler everywhere

openai-sdk/chess-app/wrangler.jsonc
```diff
@@ -18,15 +14,14 @@
...
       "new_sqlite_classes": ["ChessGame"]
     }
   ],
   "assets": {
     "directory": "dist",
     "binding": "ASSETS"
+  },
+  "vars": {
+    "HOST": "http://localhost:5174/" // replace with the hostname of your deployed worker
   }
```
threepointone: and we won't even need this!

openai-sdk/chess-app/src/index.ts
gching: hmm dont think this works since it's not an agent anymore (correct me if im wrong)

only way to get the request would potentially be passing in the request into the `McpAuthContext` or any kind of  als, which is what im doing atm

@mattzcarey thoughts on adding it in by default in `createMcpHandler` 🤔?

mattzcarey: You will be able to get it from the `extra` param in the resource once I fix something :)

mattzcarey: okay so you can actually get this from `extra.headers.host`.  

reverting url in #619

PR #614, opened 2025-10-30
feat: CSP and proper state management

mattzcarey:


docs/securing-mcp-servers.md
```diff
@@ -141,21 +141,65 @@ const clientName = sanitizeText(client.clientName);
...
+        }, 2000);
+      </script>
+    </body>
+  </html>
+`;
+
+return new Response(htmlContent, {
+  headers: buildSecurityHeaders(setCookie, nonce) // Pass nonce to include in CSP
 });
 ```
```
threepointone, line 198: can you add a bit here that If you're using something like vite, it can automatically set these things for you https://vite.dev/guide/features.html#content-security-policy-csp

PR #617, opened 2025-10-30
AI Playground improvements

deathbyknowledge:
- Since we're streaming as soon as we received the SSE response we'd stop showing the loading indicator. Fixed to show it properly.
- When the assistant called a tool we'd display tool call + result information but show an empty assistant message. Fixed to hide that.
- Tool descriptions can get long so they're collapsable now.
image


whoiskatrin:
Fab! love this

PR #626, opened 2025-11-03
fix: remove url from RequestInfo

mattzcarey:
This broke MCP standard spec and there is no reason to add it upstream.

Use `headers.host` instead for hostname. Helpful for building MCP for the Apps SDK.

### DIFF
--- a/.changeset/kind-hotels-open.md
+++ b/.changeset/kind-hotels-open.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
+Remove url field from RequestExtra in WorkerTransport. It is non standard and goes against the MCP spec types.


packages/agents/src/mcp/worker-transport.ts
```diff
@@ -392,10 +392,8 @@ export class WorkerTransport implements Transport {
       );
     }
 
-    // TODO: push this upstream to the mcp sdk asap
-    const requestInfo: RequestInfo & { url?: string } = {
-      headers: Object.fromEntries(request.headers.entries()),
-      url: request.url
+    const requestInfo: RequestInfo = {
+      headers: Object.fromEntries(request.headers.entries())
```
deathbyknowledge, line 396: Can we add `Host` ourselves from the request URL in case the header is missing? I _think_ OpenAI reads the UI resource from the server-side not from the user's browser, which doesn't automatically add the `Host` header.

mattzcarey: That's very annoying. 

I guess the question is do we hack this then or try and push this `url` param upstream. 

In the official MCP transport getting req.url only gives you paths not the hostname so adding it there felt like we were diverging.

deathbyknowledge: Yeah :/

I think we're okay adding this though, we'd be setting `Host` to the expected value, plus any non-valid value would most likely fail CORS validation anyway. I'm not a fan of the idea of diverging from the API just for this either.

PR #630, opened 2025-11-04
Fix OAuth redirect handling in MCP clients

ghostwriternr:
OAuth callbacks with successRedirect/errorRedirect returned 200 OK instead of 302 redirects because CORS response reconstruction lost status codes and headers. Relative URLs also failed since Response.redirect() requires absolute URLs.

Preserve status/statusText when adding CORS headers, use Headers constructor to properly copy all headers, and leverage URL constructor to normalize relative URLs to absolute.

Tested with Cloudflare Observability MCP server - OAuth flow now completes successfully with proper redirects.

### DIFF
--- a/packages/agents/src/index.ts
+++ b/packages/agents/src/index.ts
@@ -1757,7 +1757,8 @@ export async function routeAgentRequest<Env>(
   if (
     response &&
     corsHeaders &&
-    request.headers.get("upgrade")?.toLowerCase() !== "websocket"
+    request.headers.get("upgrade")?.toLowerCase() !== "websocket" &&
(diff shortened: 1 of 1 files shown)


packages/agents/src/index.ts
```diff
@@ -1752,14 +1757,19 @@ export async function routeAgentRequest<Env>(
   if (
     response &&
     corsHeaders &&
-    request.headers.get("upgrade")?.toLowerCase() !== "websocket" &&
```
threepointone, line 1755: can we guarantee this header doesn't come as Upgrade?

PR #631, opened 2025-11-04
Handle OAuth errors and validate redirect URLs

ghostwriternr:
OAuth providers send error parameters when authorization fails, but the client was throwing exceptions instead of handling them gracefully. This caused OAuth callbacks to crash when users denied access or when providers returned errors like access_denied.

Additionally, URL constructor calls for redirect configuration were unprotected and could crash on malformed URLs.

Return structured error results for OAuth failures and add try-catch protection for URL construction with fallback to base origin.

### DIFF
--- a/.changeset/happy-students-prove.md
+++ b/.changeset/happy-students-prove.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
+Handle OAuth errors and validate redirect URLs


packages/agents/src/tests/worker.ts
```diff
@@ -191,6 +191,14 @@ export class TestOAuthAgent extends Agent<Env> {
     return new Response("Test OAuth Agent");
   }
 
+  // Allow tests to configure OAuth callback behavior
+  configureOAuthForTest(config: {
+    successRedirect?: string;
+    errorRedirect?: string;
+  }): void {
+    this.mcp.configureOAuthCallback(config);
```
threepointone, line 199: hmm should this be wrapped with a process.env.NODE_ENV check? other wise feels like a security loophole.

ghostwriternr: This code is test-only, so should be okay? Not exposed to users.

ghostwriternr: Did it this way to avoid having to do any unnecessary changes to the core. Need this only for tests.

threepointone: Oh right it's inside tests, missed it

threepointone:
image
test's failing

threepointone:
seems like a flake. I reran and it passed. might be an issue later, but ok for now

PR #632, opened 2025-11-04
Docs sync

threepointone:


.github/workflows/sync-docs.yml
```diff
@@ -0,0 +1,122 @@
...
+
+on:
+  pull_request:
+    types: [opened, synchronize, edited]
+
+jobs:
+  sync-docs:
+    # Only run if PR title contains [docs], docs, or Docs (case-insensitive)
+    if: |
+      startsWith(github.event.pull_request.title, '[docs]') ||
```
threepointone: this only runs on docs PRs?

whoiskatrin: fixed

PR #641, opened 2025-11-07
update dependencies

threepointone:


patches/@modelcontextprotocol+sdk+1.21.0.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/@modelcontextprotocol/sdk/dist/esm/validation/ajv-provider.js b/node_modules/@modelcontextprotocol/sdk/dist/esm/validation/ajv-provider.js
```
threepointone, line 1: hate having to do this, but this will be fixed after we land https://github.com/modelcontextprotocol/typescript-sdk/pull/1065

PR #653, opened 2025-11-17
Schedule fixes

### LINKED ISSUE #600: Schedules can't process long AI requests
**Describe the bug**
When using the schedule feature to "debounce" processing some async workflow, the execution is limited to only 30s, or it'll fail with:

```
IoContext timed out due to inactivity; waitUntil tasks were canceled without completing.
```

It happens as we have to `void` a promise instead of `await` it in the scheduled function, as otherwise it fails because of the limit of `blockConcurrencyWhile`:

```
A call to blockConcurrencyWhile() in a Durable Object waited for too long. The call was canceled and the Durable Object was reset.
```

**Expected behavior**
I'm not sure how to work around this problem; it looks like removing `blockConcurrencyWhile` might help, but I'm not sure.

**Version:**
We are using a old version 0.0.76, but it also happens when upgrading to latest version `0.2.17`.

### LINKED ISSUE #616: Impossible to fully delete agent via a schedule
**Describe the bug**
I'm having difficulty deleting an agent after a period of inactivity. The root cause of this appears to be that  await this.ctx.storage.deleteAlarm(); is not always (maybe usually) respected when the object is subsequently immediately evicted with an untrappable exeption (due to .abort()).

What I'm seeing is that in my schedule handler I call .destroy(), which as advertised, deletes the DO and evicts. However, 2 seconds later (as per the retry docs for DO), the alarm is retried by the DO runtime, even though .destroy() clearly calls deleteAlarm().

This causes the agent to come back into existence and re-create its schema, which then will consume some (small amount) of storage. Although small, over time and many agents, these zombies will add up.

To work around this, rather than calling ".destroy()" I have my own version of the destroy method that omits the "abort()" call, so that the alarm handler won't throw and won't be retried. However this doesn't work because the "alarm()" handler uses the database to remove the executed schedule, but of course the database schema is gone once my handler deletes storage, so it throws anyway.

Possible solutions, which all I think require that the alarm handler does not throw when the object is destroyed because .deleteAlarm() appears unreliable, include:
* making the "alarm" handler in the base class not readonly, so we can wrap it with a try/catch in our implementation classes and silently ignore the "table not found" errors when trying to remove the stored schedule from a now non-existent table
* Swallowing this particular error in the base class, but returning early from the alarm handler (so that _scheduleNextAlarm isn't called which will have more errors)
* Setting a member variable "_destroyed" in the destroy call, then using that to exit the alarm handler early after the user code callback. This would also require omitting the ".abort()" call or making it optional with a parameter to .destroy().

**To Reproduce**
* Create an agent and use the api to create a schedule with a handler that calls and awaits destroy()
* Observe in the Cloudflare logs that the alarm is being retried, and thus re-constructs the DO and the schema

**Expected behavior**
Schedule callbacks that call "destroy" are not retried by the DO infrastructure

**Screenshots**

Image

**Version:**
0.2.17

**Additional context**
none

deathbyknowledge:
## `this.destroy()` in handlers

Currently, calling `this.destroy` inside a schedule won't clean up the Agent properly. There are 2 issues at play:
- `this.destroy` deletes all tables from SQLITE (and all other storage). This causes an error in the alarm handler since the `Agent` alarm tries to update the `cf_agent_schedules` table after the schedule callback is run.
- `this.ctx.abort` throws an uncatchable error to force DO eviction. When this happens inside the alarm handler the runtime attempts a re-try.

To address both, this PR:
1. Sets an in-memory flag `this.destroyed = true` from within `this.destroy` to skip SQLITE updates in the alarm handler whenever it's set.
2. Now yields `this.ctx.abort` to the event loop so we don't throw the error in the alarm handler's context. Since we're always in a single-threaded runtime we ensure that the handler completes before the `this.ctx.abort` task in the event loop is executed.
(IO ops after `this.destroy` could see `this.ctx.abort` being called before completion but adding anything after `this.destroy` is an anti-pattern anyway so I think it's fine)

Fixes #616.

## `this.ctx.blockConcurrencyWhile(...)` in constructor
Inside the `Agent` there's a manual call to `await this.alarm()` (for which we use the bCW) to run due schedules. Since we're using `this.ctx.storage.setAlarm` internally, the alarm method is called by the runtime after the constructor anyway, so we can get rid of our manual call.

So basically:
**BEFORE**: `constructor()` -> `await this.alarm()` (called manually) -> `alarm()` (called by the runtime) 
**NOW**: `constructor()` -> `alarm()` (called by the runtime) 

Fixes #600 

PS: There are 2 seemingly unrelated changes included:
1. `this._scheduleNextAlarm` wasn't taking into account schedules for **now** (e.g. `this.schedule(0, "destroy")`).
2. Set `observability = undefined` in the test DOs to reduce stdout noise.

### DIFF
--- a/.changeset/neat-beds-cheat.md
+++ b/.changeset/neat-beds-cheat.md
@@ -2,4 +2,6 @@
 "agents": patch
 ---
 
-add destroyed flag and yield ctx.abort
+Allow `this.destroy` inside a schedule by including a `destroyed` flag and yielding `ctx.abort` instead of calling it directly
+Fix issue where schedules would not be able to run for more 30 seconds due to `blockConccurencyWhile`. `alarm()` isn't manually called anymore, getting rid of the bCW.
+Fix an issue where immediate schedules (e.g. `this.schedule(0, "foo"))`) would not get immediately scheduled.
--- a/packages/agents/src/tests/alarms.test.ts
+++ b/packages/agents/src/tests/alarms.test.ts
@@ -1,6 +1,7 @@
 import { env } from "cloudflare:test";
 import { describe, expect, it } from "vitest";
 import type { Env } from "./worker";
+import { getAgentByName } from "..";
 
 declare module "cloudflare:test" {
   interface ProvidedEnv extends Env {}
@@ -8,10 +9,10 @@ declare module "cloudflare:test" {
 
 describe("scheduled destroys", () => {
   it("should not throw when a scheduled callback nukes storage", async () => {
-    const agentId = env.TestDestroyScheduleAgent.idFromName(
+    let agentStub = await getAgentByName(
+      env.TestDestroyScheduleAgent,
       "alarm-destroy-repro"
     );
-    let agentStub = env.TestDestroyScheduleAgent.get(agentId);
 
     // Alarm should fire immediately
     await agentStub.scheduleSelfDestructingAlarm();
@@ -20,7 +21,11 @@ describe("scheduled destroys", () => {
     // Let the alarm run
     await new Promise((resolve) => setTimeout(resolve, 50));
 
-    agentStub = env.TestDestroyScheduleAgent.get(agentId);
+    agentStub = await getAgentByName(
+      env.TestDestroyScheduleAgent,
+      "alarm-destroy-repro"
+    );
+
     await expect(agentStub.getStatus()).resolves.toBe("unscheduled");
   });
 });


threepointone:
📚 Documentation has been updated to reflect these changes.

**Documentation PR:** https://github.com/cloudflare/cloudflare-docs/pull/26558

The following documentation pages have been updated:
- `/agents/concepts/agent-class/` - Updated `destroy()` method documentation to note it can be safely called within scheduled tasks
- `/agents/api-reference/schedule-tasks/` - Added note about calling `destroy()` in scheduled task callbacks

The documentation now explains that when `destroy()` is called from within a schedule callback:
- An internal flag is set to skip remaining database updates
- `ctx.abort()` is yielded to the event loop to ensure clean alarm handler completion

threepointone:
Documentation has been updated to reflect this change in [cloudflare-docs#26558](https://github.com/cloudflare/cloudflare-docs/pull/26558).

threepointone:
## 📚 Documentation Updated

The documentation has been updated to reflect the changes in this PR.

**Updated documentation:**
- [Schedule tasks](/agents/api-reference/schedule-tasks/) - Added section on lifecycle management with scheduled tasks and immediate scheduling examples

**Cloudflare Docs PR:** https://github.com/cloudflare/cloudflare-docs/pull/26558

**Changes include:**
- Documentation of safe `destroy()` usage within scheduled callbacks
- Example showing immediate task scheduling (delay = 0)
- Explanation of how the Agent SDK defers destruction to ensure scheduled tasks complete successfully

---
*This comment was automatically generated by the documentation sync workflow.*

threepointone:
## Documentation Sync

📚 Documentation has been synced to cloudflare-docs

**Docs PR:** https://github.com/cloudflare/cloudflare-docs/pull/26558

### Changes synced:
- Updated `destroy()` method documentation to clarify behavior when called within scheduled tasks
- Added example showing `destroy()` working correctly in scheduled callbacks  
- Documented support for immediate schedules (0 second delay)
- Added lifecycle management section with self-destructing agent example

The documentation now reflects the fixes for calling `destroy()` in scheduled contexts and the support for immediate schedule execution.

threepointone:
Documentation for this PR has been updated in cloudflare/cloudflare-docs#26558

threepointone:
📚 Documentation update: https://github.com/cloudflare/cloudflare-docs/pull/26558

This PR documents the schedule bug fixes, including:
- Support for immediate schedule execution (delay=0)
- Safe `destroy()` calls within scheduled callbacks
- Internal implementation details (destroyed flag and event loop yielding)

.changeset/neat-beds-cheat.md
```diff
@@ -0,0 +1,5 @@
+---
+"agents": patch
+---
+
+add destroyed flag and yield ctx.abort
```
threepointone: this isn't clear, can you make it just a bit clearer what it does?

packages/agents/src/tests/alarms.test.ts
```diff
@@ -0,0 +1,26 @@
...
+declare module "cloudflare:test" {
+  interface ProvidedEnv extends Env {}
+}
+
+describe("scheduled destroys", () => {
+  it("should not throw when a scheduled callback nukes storage", async () => {
+    const agentId = env.TestDestroyScheduleAgent.idFromName(
+      "alarm-destroy-repro"
+    );
+    let agentStub = env.TestDestroyScheduleAgent.get(agentId);
```
threepointone: use getAgentByName here please? (in general we can never use a raw stab since it skips initialisation)

packages/agents/src/tests/alarms.test.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    // Alarm should fire immediately
+    await agentStub.scheduleSelfDestructingAlarm();
+    await expect(agentStub.getStatus()).resolves.toBe("scheduled");
+
+    // Let the alarm run
+    await new Promise((resolve) => setTimeout(resolve, 50));
+
+    agentStub = env.TestDestroyScheduleAgent.get(agentId);
```
threepointone: here too

packages/agents/src/tests/worker.ts
```diff
@@ -144,6 +147,7 @@ export class TestEmailAgent extends Agent<Env> {
 }
 
 export class TestCaseSensitiveAgent extends Agent<Env> {
+  observability = undefined;
```
threepointone, line 150: why removing observability on all these agents?

packages/agents/src/index.ts
```diff
@@ -457,26 +458,18 @@ export class Agent<
...
-          type TEXT NOT NULL CHECK(type IN ('scheduled', 'delayed', 'cron')),
-          time INTEGER,
-          delayInSeconds INTEGER,
-          cron TEXT,
-          created_at INTEGER DEFAULT (unixepoch())
-        )
-      `;
-
-        // execute any pending alarms and schedule the next alarm
-        await this.alarm();
```
threepointone, line 477: really wish I could remember why we did this. but I'll keep an.eye out for it.

threepointone (commented):
some notes, will stamp after clarification.

.changeset/neat-beds-cheat.md
deathbyknowledge: Updated

packages/agents/src/tests/worker.ts
deathbyknowledge: They flood stdout with logs about schedules, connections etc. I've never found them particularly useful (at least in the test runs).

There's still a lot of workerd internal logs but I appreciate less noise unless others find them useful

threepointone: fair. tbh I'm very unhappy with our observability stuff anyway, I want to rip it all out.

Issue #660, opened 2025-11-18
serveSSE fails to propagate headers and session ID


naji247:
**Describe the bug**
The SSE transport in MCP does not pass session ID, authentication info, or request headers through to MCP message handlers, while the StreamableHTTP transport does. This creates feature parity issues between the two transport methods and prevents SSE-based MCP connections from accessing critical request context like authentication tokens and custom headers.

**To Reproduce**

  1. Set up an MCP Agent using SSE transport with McpAgent.serveSSE() or McpAgent.serve() with transport: "sse"
  2. Add authentication middleware that sets request.auth with user credentials
  3. Connect to the agent using an MCP client (e.g., MCP Inspector) over SSE
  4. Observe that MessageExtraInfo received in the MCP server handlers is undefined, lacking:
    - Session ID
    - Authentication info (authInfo)
    - Request headers (requestInfo.headers)
   
**Expected behavior**
SSE transport should have feature parity with StreamableHTTP transport. When MCP messages are received over SSE, the handlers should receive:
  - Session ID for tracking connections across requests
  - Authentication info from middleware (if present)
  - Request headers for accessing custom headers, authorization tokens, etc.

This data should be passed via the MessageExtraInfo parameter, matching the StreamableHTTP implementation.

**Screenshots**
EXPECTED (via StreamableHTTP)

<img width="560" height="366" alt="Image" src="https://github.com/user-attachments/assets/39c4fb0a-57e2-483a-b0b7-5012676aa5f6" />

ACTUAL (via SSE)

<img width="589" height="322" alt="Image" src="https://github.com/user-attachments/assets/8e67cc65-f11e-4f31-8b04-3b8ca084dc6e" />

**Version:**
agents@0.2.23

**Additional context**
This issue is important to our team at [MCPcat](https://mcpcat.io) as we're helping Cloudflare Agents users maintain session tracking via Durable objects in Cloudflare workers.

PR #661, opened 2025-11-18
Fix: Add Session ID and Header Support to SSE Transport (#660)

naji247:
# Overview
This is a PR to fix issue #660. The SSE transport in MCP was missing critical request context that StreamableHTTP transport already provided. When using SSE, MCP servers couldn't access:
- Session IDs for connection tracking
- Authentication information from middleware
- Request headers for custom metadata or authorization

This created a feature gap between the two transports, limiting SSE-based implementations. Even though SSE is technically deprecated, our analysis of major companies with MCP servers shows that many are still dependent on SSE. To disable SSE support for their users would require all of them to update their MCP JSON configs (which is quite a hassle). 

This PR brings full feature parity to the SSE transport in the following ways:

1. Session ID Support

- Updated McpSSETransport to accept and store session IDs
- Modified transport initialization to pass session ID from the agent name

2. Authentication & Header Forwarding

- Created buildMessageExtraInfo() helper to construct MessageExtraInfo with auth and headers
- Updated onSSEMcpMessage() to accept and forward authentication context
- Modified SSE handler in utils.ts to extract and pass request metadata

# Key Changes
- McpSSETransport: Now accepts optional sessionId in constructor
- onSSEMcpMessage: Accepts extraInfo parameter with headers and auth
- buildMessageExtraInfo: New helper that filters internal headers and builds proper context
- Type definitions: Added RequestWithAuth type for cleaner code

# Testing
I added a few tests to the existing SSE based test file. The build passes successfully and the changes have been validated to ensure:
- Session IDs are properly transmitted
- Authentication info from middleware is accessible
- Request headers are forwarded correctly
- Internal transport headers are filtered out

### DIFF
--- a/packages/agents/src/tests/mcp/transports/streamable-http.test.ts
+++ b/packages/agents/src/tests/mcp/transports/streamable-http.test.ts
@@ -636,7 +636,9 @@ describe("Streamable HTTP Transport", () => {
 
       // Extract the echoed request info
       const result = parsed.result as CallToolResult;
-      const contentText = result.content?.[0]?.text;
+      const firstContent = result.content?.[0];
+      const contentText =
+        firstContent?.type === "text" ? firstContent.text : undefined;
       const echoedData = JSON.parse(
         typeof contentText === "string" ? contentText : "{}"
       );


naji247:
> this looks good to me, but I'll let @deathbyknowledge have a look before we land it since he'd touched this part of the code

Thank you @threepointone ! Please let us know if you want any modifications. I tried to keep it similar to how Streamable HTTP was implemented.

packages/agents/src/tests/mcp/transports/streamable-http.test.ts
```diff
@@ -583,4 +583,104 @@ describe("Streamable HTTP Transport", () => {
...
+          "Content-Type": "application/json",
+          "mcp-session-id": sessionId,
+          "x-user-id": "test-user-123",
+          "x-request-id": "req-456",
+          "x-custom-header": "custom-value"
+        },
+        method: "POST"
+      });
+
+      // Spy on the transport to capture the extraInfo
```
deathbyknowledge: Doesn't seem like this is testing anything new here. What am I missing?

packages/agents/src/tests/mcp/transports/streamable-http.test.ts
```diff
@@ -583,4 +583,104 @@ describe("Streamable HTTP Transport", () => {
...
+      const response = await worker.fetch(request, env, ctx);
+
+      expect(response.status).toBe(200);
+
+      // The test verifies that internal headers (cf-mcp-method, cf-mcp-message, upgrade)
+      // are filtered out before being passed to the MCP server
+      // Only custom headers like x-custom-header should be preserved in requestInfo.headers
+      const sseText = await readSSEEvent(response);
+      const result = parseSSEData(sseText);
+      expectValidToolsList(result);
```
deathbyknowledge: On a closer look it seems like the 3 tests are doing the same. They are either adding custom headers to the request or setting an `auth` property.
Then the test checks that the request was successful and that the tool list is correct but there are no assertions relevant to either the custom headers or the `auth` field.

One thing that can be done is update the MCP behind the worker to do/return differently in these scenarios.

packages/agents/src/mcp/index.ts
```diff
@@ -188,14 +215,20 @@ export abstract class McpAgent<
...
+    extraInfo?: { headers: Record<string, string>; auth?: AuthInfo }
   ): Promise<Error | null> {
     // Since we address the DO via both the protocol and the session id,
     // this should never happen, but let's enforce it just in case
     if (this.getTransportType() !== "sse") {
       return new Error("Internal Server Error: Expected SSE transport");
     }
 
+    // Ensure transport is initialized before processing messages
+    if (!this._transport) {
```
deathbyknowledge: This should not be needed, `onStart` is called every time a request is routed and the DO is initialized. Were you facing issues with `this._transport` not being initialized?

packages/agents/src/mcp/index.ts
```diff
@@ -85,11 +89,34 @@ export abstract class McpAgent<
...
+  private buildMessageExtraInfo(extraInfo?: {
+    headers: Record<string, string>;
+    auth?: AuthInfo;
+  }): MessageExtraInfo | undefined {
+    if (!extraInfo) return;
+
+    const headers = { ...extraInfo.headers };
+
+    // Remove internal headers that are not part of the original request
+    delete headers["content-type"];
```
deathbyknowledge: These headers _are_ part of the original request, they should not be removed

packages/agents/src/mcp/index.ts
```diff
@@ -85,11 +89,34 @@ export abstract class McpAgent<
...
+  }
+
   /** Returns a new transport matching the type of the Agent. */
   private initTransport() {
     switch (this.getTransportType()) {
       case "sse": {
-        return new McpSSETransport(() => this.getWebSocket());
+        return new McpSSETransport(
+          () => this.getWebSocket(),
+          this.getSessionId()
```
deathbyknowledge: In the StreamableHTTP we use the pattern [here](https://github.com/cloudflare/agents/blob/main/packages/agents/src/mcp/transport.ts#L122-L128) to get this, so we can avoid drilling fields through the constructor.

packages/agents/src/mcp/utils.ts
```diff
@@ -682,7 +682,10 @@ export const createLegacySseHandler = (
       });
 
       const messageBody = await request.json();
-      const error = await agent.onSSEMcpMessage(sessionId, messageBody);
+      const error = await agent.onSSEMcpMessage(sessionId, messageBody, {
+        headers: Object.fromEntries(request.headers.entries()),
+        auth: (request as RequestWithAuth).auth
```
deathbyknowledge: Hmm this normally runs inside your Worker `fetch` handler, how did the request get the `auth` property?

packages/agents/src/mcp/index.ts
naji247: Sorry, I did this just to make sure it was properly initialized. I will remove this :)

packages/agents/src/mcp/index.ts
naji247: Sounds good. Will do 🫡 

Originally I was worried about breaking existing uses of this class, but it seems to be internal only and limited to this part of the codebase.

packages/agents/src/tests/mcp/transports/streamable-http.test.ts
naji247: I decided to create a tool in the `workers.ts` test utility that actually returns the requestInfo it receives to make it more thorough of a test.

packages/agents/src/mcp/utils.ts
naji247: You're right, for the purposes of the Cloudflare users we're aiding, we actually only needed the headers (and not `auth` explicitly) so our tests weren't covering the auth header. I added it in with the hopes of meeting parity with StreamableHTTP but I decided to instead reduce the scope of this PR to just include header and session management information.

naji247:
@deathbyknowledge I went ahead and updated the tests, reverted some of the changes I made to include auth information, and refactored the SSE constructor to use an `options` object paradigm.

packages/agents/src/mcp/index.ts
```diff
@@ -85,11 +88,32 @@ export abstract class McpAgent<
     return websockets[0];
   }
 
+  /** Build MessageExtraInfo from extracted headers and auth info */
+  private buildMessageExtraInfo(extraInfo?: {
```
deathbyknowledge: I think we can get rid of this and just pass the extra object to `onSSEMcpMessage` directly

packages/agents/src/mcp/utils.ts
```diff
@@ -682,7 +682,9 @@ export const createLegacySseHandler = (
       });
 
       const messageBody = await request.json();
-      const error = await agent.onSSEMcpMessage(sessionId, messageBody);
+      const error = await agent.onSSEMcpMessage(sessionId, messageBody, {
```
deathbyknowledge: If we remove `buildMessageExtraInfo` this can be:
```ts
      const headers =  Object.fromEntries(request.headers.entries());
      const error = await agent.onSSEMcpMessage(sessionId, messageBody, {
        requestInfo: { headers }
      });
```

packages/agents/src/mcp/transport.ts
```diff
@@ -15,17 +15,23 @@ import type { McpAgent } from ".";
...
   onclose?: () => void;
   onerror?: (error: Error) => void;
-  onmessage?: (message: JSONRPCMessage) => void;
+  onmessage?: (message: JSONRPCMessage, extra?: MessageExtraInfo) => void;
 
   private _getWebSocket: () => WebSocket | null;
   private _started = false;
-  constructor(getWebSocket: () => WebSocket | null) {
-    this._getWebSocket = getWebSocket;
+  constructor(options: McpSSETransportOptions) {
```
deathbyknowledge: Sorry I wasn't clear before, this is the pattern I was referring to.
```ts
constructor() {
    const { agent } = getCurrentAgent<McpAgent>();
    if (!agent)
      throw new Error("McpAgent was not found in Transport constructor");

    // Initialization is handled in `McpAgent.serveSSE()` and agents are addressed by sessionId,
    // so we'll always have this available.
    this.sessionId = agent.getSessionId();
    this._getWebSocket = () => agent.getWebSocket();
}
```

packages/agents/src/mcp/index.ts
```diff
@@ -85,11 +88,32 @@ export abstract class McpAgent<
...
+    };
+  }
+
   /** Returns a new transport matching the type of the Agent. */
   private initTransport() {
     switch (this.getTransportType()) {
       case "sse": {
-        return new McpSSETransport(() => this.getWebSocket());
+        return new McpSSETransport({
+          getWebSocket: () => this.getWebSocket(),
```
deathbyknowledge: None of the parameters here are needed after the changes mentioned in the transport constructor

deathbyknowledge (changes requested):
Looks good! There are a few nits and one type error breaking the CI but this should be ready to merge after those are addressed 👍

packages/agents/src/mcp/transport.ts
naji247: I thought you might have meant this :) Wasn't 100% sure. Let me know if it was fine to make the getWebSocket method public, as the example you gave sort of requires that.

naji247:
@deathbyknowledge refactored to use the `getCurrentAgent` utility and removed the less necessary buildExtra method like you asked! Also, I think all type checks are passing locally now too.

naji247:
Ah looks like the TS issue still exists... fixing now!

naji247:
@deathbyknowledge sorry to trouble you again, but do you think we can merge this in sometime this week? :)

packages/agents/src/mcp/utils.ts
```diff
@@ -682,7 +683,24 @@ export const createLegacySseHandler = (
       });
 
       const messageBody = await request.json();
-      const error = await agent.onSSEMcpMessage(sessionId, messageBody);
+
+      // Build MessageExtraInfo with filtered headers
+      const headers = Object.fromEntries(request.headers.entries());
+
+      // Remove internal headers that are not part of the original request
+      delete headers[MCP_HTTP_METHOD_HEADER];
```
deathbyknowledge: These `delete`s can be removed, they are only relevant in the `StreamableHTTP` transport.

naji247: Removed! 👍

naji247:
@deathbyknowledge should be good to go :)

deathbyknowledge:
CI is failing, seems like a silly type error. Can you check with `npm run check` please?

whoiskatrin:
> @deathbyknowledge should be good to go :)

@naji247, sorry to bug you, but type checks are failing

naji247:
Thanks @whoiskatrin I swear they passed locally :( fixed now I think

naji247:
Woo 🥳 they passed.

deathbyknowledge:
Merged O7

PR #662, opened 2025-11-19
assorted playground fixes

threepointone:
- use streamdown for text
- move the input down
- scroll to the bottom when messages change
- fix the bug where it would stop streaming


site/ai-playground/src/app.tsx
```diff
@@ -57,7 +58,6 @@ function generateNewSessionId(): string {
 }
 
 const App = () => {
-  const [error, setError] = useState("");
```
threepointone, line 60: this wasn't used anywhere

PR #673, opened 2025-11-23
added resumable streaming with minimal setup

whoiskatrin:
something to play with https://resumable-stream-chat.katjareznikova.workers.dev


examples/resumable-stream-chat/src/server.ts
```diff
@@ -0,0 +1,113 @@
...
+> {
+  private streams = new ResumableStreamManager(this, this.ctx);
+
+  initialState: ResumableStreamingChatState = {
+    messages: [],
+    activeStreamId: null
+  };
+
+  async onStart() {
+    await super.onStart();
```
deathbyknowledge: What is this for?

examples/resumable-stream-chat/wrangler.jsonc
```diff
@@ -0,0 +1,28 @@
+{
+  "$schema": "../../node_modules/wrangler/config-schema.json",
+  "compatibility_date": "2025-03-14",
+  "compatibility_flags": [
+    "nodejs_compat",
+    "nodejs_compat_populate_process_env"
```
threepointone, line 6:
```suggestion
    "nodejs_compat"
```

examples/resumable-stream-chat/wrangler.jsonc
```diff
@@ -0,0 +1,28 @@
+{
+  "$schema": "../../node_modules/wrangler/config-schema.json",
+  "compatibility_date": "2025-03-14",
```
threepointone, line 3:
```suggestion
  "compatibility_date": "2025-11-07",
```

examples/resumable-stream-chat/env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="vite/client" />
```
deathbyknowledge: 🧐

packages/agents/src/ai-chat-agent.ts
```diff
@@ -42,13 +121,284 @@ export class AIChatAgent<Env = unknown, State = unknown> extends Agent<
...
+    `;
+
+    if (activeStreams && activeStreams.length > 0) {
+      const stream = activeStreams[0];
+      const streamAge = Date.now() - stream.created_at;
+
+      // Check if stream is stale (started too long ago)
+      if (streamAge > STREAM_STALE_THRESHOLD_MS) {
+        // Mark stale stream as error and don't restore
+        this.sql`
```
deathbyknowledge: Just a thought, do we rather delete the entry here instead? As a user I'm not sure I want to keep taking storage with older stream metadata

packages/agents/src/ai-chat-agent.ts
```diff
@@ -42,13 +121,284 @@ export class AIChatAgent<Env = unknown, State = unknown> extends Agent<
...
+          : 0;
+    }
+  }
+
+  /**
+   * Override onConnect to notify clients about active streams that can be resumed.
+   * Stream chunks are sent only after the client acknowledges with CF_AGENT_STREAM_RESUME_ACK.
+   */
+  override async onConnect(connection: Connection, ctx: ConnectionContext) {
+    await super.onConnect(connection, ctx);
```
deathbyknowledge: Is the `super.onConnect` needed?

deathbyknowledge (commented):
Looks good, only had a few thoughts. This needs a test though.

packages/agents/src/ai-chat-agent.ts
```diff
@@ -42,13 +121,284 @@ export class AIChatAgent<Env = unknown, State = unknown> extends Agent<
...
+          ? lastChunk[0].max_index + 1
+          : 0;
+    }
+  }
+
+  /**
+   * Override onConnect to notify clients about active streams that can be resumed.
+   * Stream chunks are sent only after the client acknowledges with CF_AGENT_STREAM_RESUME_ACK.
+   */
+  override async onConnect(connection: Connection, ctx: ConnectionContext) {
```
threepointone: instead of overriding directly here, can we do it like we dp in the base agent class? so that consumers won't have to call super.onConnect

whoiskatrin: good call, yes, we totally should do that

packages/agents/src/ai-chat-agent.ts
whoiskatrin: fixed

packages/agents/src/ai-chat-agent.ts
whoiskatrin: fixed

examples/resumable-stream-chat/env.d.ts
whoiskatrin: tbh, we don't really need it, removed

examples/resumable-stream-chat/src/server.ts
whoiskatrin: not needed anymore with Sunil's suggestion

PR #683, opened 2025-11-27
add task management and workflow integration

whoiskatrin:
Addded task-runner example here

Shows how to run long-running background tasks from an Agent using Cloudflare Workflows.

Two patterns demonstrated:
- **Quick analysis** - runs in the Agent, good for fast tasks (<30s)
- **Deep analysis** - runs in a Workflow, durable with automatic retries
The workflow sends progress updates back to the Agent, which syncs to connected clients via WebSocket.

**Files:**
- examples/task-runner/src/server.ts - Agent that dispatches to workflows
- examples/task-runner/src/workflows/analysis.ts - Durable workflow with step-based checkpointing
- examples/task-runner/src/App.tsx - React UI with real-time progress

### DIFF
--- a/examples/task-runner/src/workflows/analysis.ts
+++ b/examples/task-runner/src/workflows/analysis.ts
@@ -1,8 +1,17 @@
 /**
- * Analysis Workflow
+ * Deep Analysis Workflow
  *
- * A durable workflow that analyzes GitHub repositories using AI.
- * Demonstrates Cloudflare Workflows with step-based checkpointing.
+ * A durable workflow demonstrating Cloudflare Workflows capabilities that are
+ * IMPOSSIBLE in a regular Durable Object:
+ *
+ * 1. step.waitForEvent() - Pause for human approval (can wait hours/days)
+ * 2. step.sleep() - Schedule future work (workflow hibernates, no compute cost)
+ * 3. Multi-phase execution - Survives DO eviction, restarts, deploys
+ * 4. Automatic retries - Exponential backoff on transient failures
+ * 5. Long-running analysis - Can take minutes/hours without timeout
+ *
+ * Quick Analysis (in Agent): Single pass, ~30 seconds, no durability
+ * Deep Analysis (this Workflow): Multi-phase, can pause for approval, scheduled follow-ups
  */
 
 import {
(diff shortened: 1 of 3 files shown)


examples/task-runner/src/workflows/analysis.ts
```diff
@@ -0,0 +1,318 @@
...
+        ] as DurableObjectNamespace;
+        if (!agentNS) return;
+
+        const agentId = agentNS.idFromName(_agentName);
+        const agent = agentId ? agentNS.get(agentId) : null;
+        if (!agent) return;
+
+        // Call the handleWorkflowUpdate method via RPC-style fetch
+        await agent.fetch(
+          new Request("http://internal/rpc", {
```
deathbyknowledge: This is not RPC and the agent does not implement `onRequest`, so it just fails

examples/task-runner/src/workflows/analysis.ts
```diff
@@ -0,0 +1,318 @@
...
+            method: "POST",
+            headers: { "Content-Type": "application/json" },
+            body: JSON.stringify({
+              type: "rpc",
+              id: crypto.randomUUID(),
+              method: "handleWorkflowUpdate",
+              args: [{ taskId, ...update }]
+            })
+          })
+        );
```
deathbyknowledge:
```suggestion
      await agent.handleWorkflowUpdate({ taskId, ...update });
```

examples/task-runner/src/workflows/analysis.ts
whoiskatrin: Fixed

examples/task-runner/src/workflows/analysis.ts
whoiskatrin: Fixed

whoiskatrin:
> The quick and deep analysis are effectively doing the same... Imo we should try to do more with the workflow so the benefits are clearer

Added some more workflow specific features liek human in the loop if finds critical issues and sleep steps.

examples/task-runner/vite.config.ts
```diff
@@ -0,0 +1,6 @@
+import { cloudflare } from "@cloudflare/vite-plugin";
+import { defineConfig } from "vite";
+
+export default defineConfig({
+  plugins: [cloudflare()]
```
deathbyknowledge, line 5: How come it works without the react plugin? I thought it was required

deathbyknowledge (approved):
Looks good now, covers most features of workflows

threepointone:
this looks good, let's chat before landing this

deathbyknowledge:
We can close this since #799 got merged right?

Issue #24, opened 2025-02-24
ai chat agent should handle multiple inputs coming in at the same time


threepointone:
queue them up in the backend and make sure all in-flight responses have been sent before responding to it

threepointone:
this is fixed with messageConcurrency

PR #685, opened 2025-11-27
chat processing with typing indicators and batching support for msgs

whoiskatrin:
the idea is to type more than one msg at a time
want to give users 2 options:
- blast the chat and let him reply after all my messages (that's where typing indicator comes to mind)
- or say a few things, but it gets queued and replied to separately
inspired by this https://github.com/cloudflare/agents/issues/24


packages/agents/src/ai-chat-agent.ts
```diff
@@ -187,70 +327,38 @@ export class AIChatAgent<Env = unknown, State = unknown> extends Agent<
...
-                  id: data.id,
-                  type: MessageType.CF_AGENT_USE_CHAT_RESPONSE
-                },
-                [connection.id]
-              );
-            }
-          });
+          if (this.chatProcessingMode === "sequential") {
+            // Sequential mode: queue all requests and process one by one
+            this._chatQueue.push(request);
```
deathbyknowledge: Do we not want to use the built-in `queue`?

packages/agents/src/ai-chat-agent.ts
```diff
@@ -34,6 +34,23 @@ const STREAM_STALE_THRESHOLD_MS = 5 * 60 * 1000;
 const CLEANUP_INTERVAL_MS = 10 * 60 * 1000;
 /** Default age threshold for cleaning up completed streams (ms) - 24 hours */
 const CLEANUP_AGE_THRESHOLD_MS = 24 * 60 * 60 * 1000;
+/** Default idle timeout - how long to wait for user to start typing after sending a message (ms) */
+const DEFAULT_CHAT_IDLE_TIMEOUT_MS = 5000;
```
deathbyknowledge: So by default we wait 5 seconds to start processing a message if the user is **not** typing? That doesn't sound right

packages/agents/src/ai-chat-agent.ts
```diff
@@ -34,6 +34,23 @@ const STREAM_STALE_THRESHOLD_MS = 5 * 60 * 1000;
...
+/** Default typing timeout - how long to wait after user stops typing (ms) */
+const DEFAULT_CHAT_TYPING_TIMEOUT_MS = 1500;
```
deathbyknowledge: I'd expect this one to be higher than the one above, since the user possibly will send another message.

packages/agents/src/ai-chat-agent.ts
```diff
@@ -112,6 +129,129 @@ export class AIChatAgent<Env = unknown, State = unknown> extends Agent<
...
+   * }
+   *
+   * // Batch - combine messages into one response
+   * class MyAgent extends AIChatAgent {
+   *   chatProcessingMode = "batch";
+   *   chatIdleTimeout = 3000; // 3 second batching window
+   * }
+   * ```
+   */
+  chatProcessingMode: ChatProcessingMode = "sequential";
```
deathbyknowledge: The base `Agent` class has the static `options` which only host `hibernate: true/false`. Feels like this is a class-level config as well, maybe we want to have something similar instead of just adding class properties for developers to configure.

deathbyknowledge (commented):
Left some comments, I might've gotten the logic wrong with the timeouts

packages/agents/src/ai-chat-agent.ts
whoiskatrin: if only applies to batch mode, when chatProcessingMode = "batch", in a normal setup, no delays

PR #696, opened 2025-12-03
fix callback validation in mcp client

mattzcarey:
- validate the nonce and server id from saved params. 
- give them a ttl
- delete when validated
- callback urls are now the same across all servers on this client, difference is the state param which is very oauth-y

add some tests for this.

update the playground to test this. 

Screenshot 2025-12-03 at 12 56 36


packages/agents/src/mcp/client.ts
```diff
@@ -635,46 +635,81 @@ export class MCPClientManager {
...
+    const url = new URL(req.url);
+    const state = url.searchParams.get("state");
+    const serverId = this.extractServerIdFromState(state);
+    if (!serverId) {
+      return false;
+    }
+
     const servers = this.getServersFromStorage();
-    return servers.some(
-      (server) => server.callback_url && req.url.startsWith(server.callback_url)
```
ghostwriternr, line 652: no-op: I can't believe this logic has lasted here as long as it has. Love the move away from this!

packages/agents/src/mcp/client.ts
```diff
@@ -720,32 +744,20 @@ export class MCPClientManager {
...
-    const clientId = conn.options.transport.authProvider.clientId || state;
-
-    // Set the OAuth credentials
-    conn.options.transport.authProvider.clientId = clientId;
-    conn.options.transport.authProvider.serverId = serverId;
-
     try {
+      await authProvider.consumeState(state);
       await conn.completeAuthorization(code);
+      await authProvider.deleteCodeVerifier();
```
ghostwriternr, line 750: no-op: I like this pattern

packages/agents/src/mcp/client.ts
```diff
@@ -635,46 +635,81 @@ export class MCPClientManager {
...
   }
 
   async handleCallbackRequest(req: Request) {
     const url = new URL(req.url);
+    const code = url.searchParams.get("code");
+    const state = url.searchParams.get("state");
+    const error = url.searchParams.get("error");
+    const errorDescription = url.searchParams.get("error_description");
+
+    if (!state) {
```
ghostwriternr, line 671: Can be handled in a separate PR (in fact, I'd prefer them to be separate): Do these errors get surfaced properly to the browser? I vaguely remember feeling constantly stumped by the error messages that surface up to the DO runtime from the MCP client and they were not helpful for 1. me as a user and 2. me as a developer building a client. Maybe these do get surfaced properly - but just checking anyway.

Might need to look at `handleMcpOAuthCallback`. Also test using various combinations of `customHandler` and `errorRedirect`.

ghostwriternr: And if we do handle these errors, are we broadcasting the state with `this.broadcastMcpServers` to our React hooks even in error scenarios?

packages/agents/src/tests/worker.ts
```diff
@@ -391,7 +421,44 @@ export class TestOAuthAgent extends Agent<Env> {
         transport: {
           authProvider: {
```
ghostwriternr, line 422: nit, for a later pr: not at all related to this change but I got reminded when I noticed the ability to pass in a custom authProvider here

We've deprecated all the places where a custom authProvider is publicly exposed afaik. Maybe we should update examples/mcp-client too so no one uses the  pattern anymore.

.changeset/nine-bottles-heal.md
```diff
@@ -0,0 +1,5 @@
+---
+"agents": patch
+---
+
+Enables connecting to multiple MCP servers simultaneously and hardens OAuth state handling against replay/DoS attacks.
```
ghostwriternr, line 5: Optional: maybe we should add a note that authentication flows that were initiated when using an older version would not complete after switching to the new version, due to change of callback url? But I don't think it's too significant an issue (super edge case imo - mainly just need to keep devs informed?), so the semver bump can still be a patch.

mattzcarey:
@claude this is not a race condition. We extend the auth provider which is used in the MCP SDK.

packages/agents/src/mcp/do-oauth-client-provider.ts
```diff
@@ -124,17 +137,90 @@ export class DurableObjectOAuthClientProvider implements AgentsOAuthProvider {
...
+
+  async state(): Promise<string> {
+    const nonce = nanoid();
+    const state = `${nonce}.${this.serverId}`;
+    const storedState: StoredState = {
+      nonce,
+      serverId: this.serverId,
+      createdAt: Date.now()
+    };
+    await this.storage.put(this.stateKey(nonce), storedState);
```
ghostwriternr, line 152: Minor, optional: I don't think this will happen, but wondering if we should clean these up lazily or via alarms in case many users end up leaving OAuth flows incomplete.

ghostwriternr (approved):
This is a solid change, lfg! Added a few super minor comments that I'd prefer we look at in a separate PR (need not even act on all of them). This PR, in its current state, is pretty good and complete.

packages/agents/src/mcp/client.ts
mattzcarey: fair 100%. Another PR agreed

packages/agents/src/tests/worker.ts
mattzcarey: lol yes

packages/agents/src/mcp/do-oauth-client-provider.ts
mattzcarey: imo you would have one DO per user so I dont see it as a massive issue. We could do a lazy cleanup tho

ghostwriternr: Aha, right. no-op then tbh. can deal with this if it becomes a stated issue.

PR #715, opened 2025-12-05
intercept elicitation responses

whoiskatrin:


packages/agents/src/mcp/index.ts
```diff
@@ -95,7 +95,11 @@ export abstract class McpAgent<
         return new McpSSETransport();
       }
       case "streamable-http": {
-        return new StreamableHTTPServerTransport({});
+        const transport = new StreamableHTTPServerTransport({});
+        transport.messageInterceptor = async (message) => {
```
mattzcarey: We need a better way to do this. Stamping not to block but this leaks transport specific logic

PR #722, opened 2025-12-09
fix: move AI SDK packages to peer dependencies

agcty:
## Summary

- Move `ai` and `@ai-sdk/openai` from direct dependencies to peer dependencies
- Add `@ai-sdk/react` as optional peer dependency (was previously undeclared despite being imported)
- Allow consumers to control their own AI SDK versions

## Problem

Currently, `ai` and `@ai-sdk/openai` are pinned as direct dependencies (`"ai": "5.0.103"`). This causes issues for consumers using different AI SDK versions (e.g., AI SDK 6):

- Duplicate packages get installed (one for the consumer's code, one for agents)
- Bundle bloat from including two versions
- Potential type mismatches when passing messages between consumer code and `AIChatAgent`

Additionally, `@ai-sdk/react` is imported in `ai-react.tsx` but was never declared in package.json, relying on consumers having it installed transitively.

## Solution

- `ai`: required peer dep (`>=5.0.0`) - core AI SDK types and functions
- `@ai-sdk/openai`: optional peer dep (`>=2.0.0`) - only needed for codemode feature
- `@ai-sdk/react`: optional peer dep (`>=1.0.0`) - only needed for React hooks

All three packages are also added to `devDependencies` for local development and testing.

## Breaking Changes

Consumers who don't already have `ai` installed will need to add it explicitly. In practice, anyone using the AI features is already installing `ai` separately anyway.

Closes #42

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -685,6 +685,7 @@
       "version": "2.0.16",
       "resolved": "https://registry.npmjs.org/@ai-sdk/gateway/-/gateway-2.0.16.tgz",
       "integrity": "sha512-qiIaVs1w1XcNiFG6cjhOwolPuMFSvy6ZxDeLaPlEK/kSmNGfd+gUA2CTpBPWWT3qN6Zxfdrwq+ti4BfkdmLIJQ==",
+      "dev": true,
       "license": "Apache-2.0",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


whoiskatrin:
@agcty don't merge it just yet, please

PR #723, opened 2025-12-09
fix: require React 19 in peer dependencies

agcty:
## Summary

Update the React peer dependency from `"*"` (any version) to `"^19.0.0"`.

## Problem

The React hooks (`useAgent`, `useAgentChat`) use the `use()` hook:

```tsx
// react.tsx:296
const queryResult = use(queryPromise!);

// ai-react.tsx:202
const initialMessages = initialMessagesPromise
  ? use(initialMessagesPromise)
  : (optionsInitialMessages ?? []);
```

The `use()` hook was introduced in React 19 and doesn't exist in React 18 or earlier. The current `"react": "*"` constraint is misleading since consumers on React 18 will get a confusing runtime error:

```
TypeError: use is not a function
```

## Solution

Change `"react": "*"` to `"react": "^19.0.0"` in both `peerDependencies` and `devDependencies`. This makes the requirement explicit so consumers get a clear error during installation rather than at runtime.

## Breaking Changes

Consumers on React 18 will now see a peer dependency warning/error during installation. However, the code was already broken for them at runtime, so this just surfaces the issue earlier.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19274,7 +19274,6 @@
       "os": [
         "android"
       ],
-      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -19315,7 +19314,6 @@
       "os": [
         "darwin"
       ],
-      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -19336,7 +19334,6 @@
       "os": [
         "freebsd"
       ],
-      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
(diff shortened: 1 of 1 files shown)


whoiskatrin:
@agcty don't merge it just yet, please. Also, can you please fix types in this PR?

agcty:
@whoiskatrin Fixed and tested locally, CI should pass now!

I also added @types/react assuming that's what you meant by "fix types", but the build passes without it too. Let me know if you'd like me to revert that part or if you meant something else!

PR #724, opened 2025-12-09
fix: move MCP SDK to peer dependencies

agcty:
## Summary

Move `@modelcontextprotocol/sdk` from direct dependencies to peer dependencies, allowing consumers to control their own MCP SDK version.

## Problem

Currently, `@modelcontextprotocol/sdk` is pinned as a direct dependency (`"^1.23.0"`). This causes issues for consumers using different MCP SDK versions:

- Duplicate packages get installed
- Type mismatches when consumers reference MCP types (`Tool`, `Prompt`, `Resource`, `ServerCapabilities`) from their own code
- The agents package exposes MCP SDK types in its public API (e.g., `MCPServersState`, `McpAgent.server`)

## Solution

- `@modelcontextprotocol/sdk`: required peer dep (`>=1.0.0`)

Unlike the AI SDK packages where some are optional, MCP is a required peer dependency because:
1. MCP functionality is core to the package (every `Agent` has `this.mcp` built-in)
2. Types like `MCPServersState` and `McpAgent` are fundamental exports
3. The `McpAgent` abstract class directly exposes `McpServer | Server` types

## Breaking Changes

Consumers who don't already have `@modelcontextprotocol/sdk` installed will need to add it explicitly. However, anyone using MCP features is likely already interacting with MCP types.

Related to #42

### DIFF
--- a/packages/agents/package.json
+++ b/packages/agents/package.json
@@ -26,7 +26,6 @@
   "dependencies": {
     "@ai-sdk/openai": "2.0.73",
     "@cfworker/json-schema": "^4.1.1",
-    "@modelcontextprotocol/sdk": "^1.23.0",
     "ai": "5.0.103",
     "cron-schedule": "^6.0.0",
     "json-schema": "^0.4.0",
@@ -41,12 +40,14 @@
   },
   "devDependencies": {
     "@cloudflare/workers-oauth-provider": "^0.1.0",
+    "@modelcontextprotocol/sdk": "^1.23.0",
     "@types/yargs": "^17.0.35",
     "react": "*",
     "vitest-browser-react": "^1.0.1",
     "x402": "^0.7.3"
   },
   "peerDependencies": {
+    "@modelcontextprotocol/sdk": ">=1.0.0",
     "react": "*",
     "viem": ">=2.0.0",
     "x402": "^0.7.1"


whoiskatrin:
@agcty don't merge it just yet, please

packages/agents/package.json
```diff
@@ -41,12 +40,14 @@
...
   "devDependencies": {
     "@cloudflare/workers-oauth-provider": "^0.1.0",
+    "@modelcontextprotocol/sdk": "^1.23.0",
     "@types/yargs": "^17.0.35",
     "react": "*",
     "vitest-browser-react": "^1.0.1",
     "x402": "^0.7.3"
   },
   "peerDependencies": {
+    "@modelcontextprotocol/sdk": ">=1.0.0",
```
threepointone, line 50: because the mcp sdk does breaking changes in minors right now, I don't think we can actually guarantee this. not a great situation for us to be in unfortunately. we need to think through this. 

tagging @mattzcarey for visibility

threepointone (changes requested):
blocking this while we figure out what to do here

mattzcarey:
When MCP v2 lands we should be able to do this but for now this is gonna be super tough to support as their are breaking changes in every minor version atm

kamranayub:
Just wanna chime in to say that even with it built in, pulling down and trying any of the current ai templates:

https://github.com/cloudflare/ai/tree/main/demos/remote-mcp-authless

I get a TypeScript error because the MCP SDK isn't found:

```
❯ npm run type-check

> remote-mcp-server-authless@0.0.0 type-check
> tsc --noEmit

src/index.ts:1:27 - error TS2307: Cannot find module '@modelcontextprotocol/sdk/server/mcp.js' or its corresponding type declarations.

1 import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/index.ts:15:14 - error TS7031: Binding element 'name' implicitly has an 'any' type.

15     async ({ name }) => {
                ~~~~


Found 2 errors in the same file, starting at: src/index.ts:1
```

I realized that `agents` was using MCP as a direct dep, which is also why it failed when installing the _latest_ SDK as a direct dep in my project.

If I install the exact version agents depends on (as a devDep), I can get it to work

PR #726, opened 2025-12-09
fix cache ttl

whoiskatrin:


packages/agents/src/react-tests/cache-ttl.test.ts
```diff
@@ -8,63 +8,54 @@ describe("Cache TTL", () => {
   });
 
   it("should respect cacheTtl of 0 (immediate expiration)", async () => {
-    const key = ["test", "default", "dep1"];
+    const key = "test-key-1";
```
mattzcarey, line 11: maybe silly question but is this breaking?

mattzcarey (approved):

PR #729, opened 2025-12-10
add client-defined tools and prepareSendMessagesRequest options

whoiskatrin:
### Fix client-side tool execution and message duplication
- Add support for client-defined tools via useAgentChat with automatic and manual (human-in-the-loop) execution modes
- Fix duplicate assistant messages when client-side tools execute by sending CF_AGENT_TOOL_RESULT to server as source of truth
- Improve message ID synchronization between client and server by matching on toolCallId when IDs differ
- Add toolsRequiringConfirmation option for human-in-the-loop workflows
- Exclude originating connection from CF_AGENT_CHAT_MESSAGES broadcasts to prevent message conflicts

### DIFF
--- a/.changeset/slow-jobs-boil.md
+++ b/.changeset/slow-jobs-boil.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
+add client-defined tools and prepareSendMessagesRequest options


packages/agents/src/ai-chat-agent.ts
```diff
@@ -25,6 +26,72 @@ import {
...
+ * Schema for a client-defined tool sent from the browser.
+ * These tools are executed on the client, not the server.
+ */
+export type ClientToolSchema = {
+  /** Unique name for the tool */
+  name: string;
+  /** Human-readable description of what the tool does */
+  description?: string;
+  /** JSON Schema defining the tool's input parameters */
+  parameters?: Record<string, unknown>;
```
mattzcarey: nit: naming here. Can we not build this type from ai sdk tools omitting `execute`

packages/agents/src/ai-react.tsx
```diff
@@ -8,17 +8,127 @@ import type {
...
 import type { OutgoingMessage } from "./ai-types";
 import { MessageType } from "./ai-types";
 import type { useAgent } from "./react";
 
+/**
+ * JSON Schema type for tool parameters.
+ * Supports common JSON Schema properties with an index signature
+ * for extension properties like $ref, format, minimum, maximum, etc.
+ */
+export type JSONSchemaType = {
```
mattzcarey: there are so many different specs for json schema. MCP standardise on 2020-12. you can use https://www.npmjs.com/package/json-schema-typed as a type lib for this.

packages/agents/src/ai-react.tsx
```diff
@@ -8,17 +8,127 @@ import type {
...
+ */
 export type AITool<Input = unknown, Output = unknown> = {
+  /** Human-readable description of what the tool does */
   description?: string;
+  /** JSON Schema defining the tool's input parameters */
+  parameters?: JSONSchemaType;
+  /**
+   * @deprecated Use `parameters` instead. Will be removed in a future version.
+   */
   inputSchema?: unknown;
```
mattzcarey: intereste on this change and why we are vendoring this type ourselves not using ai sdk.

packages/agents/src/ai-react.tsx
```diff
@@ -8,17 +8,127 @@ import type {
...
+ * Must match the server-side ClientToolSchema type in ai-chat-agent.ts.
+ */
+export type ClientToolSchema = {
+  /** Unique name for the tool */
+  name: string;
+  /** Human-readable description of what the tool does */
+  description?: string;
+  /** JSON Schema defining the tool's input parameters */
+  parameters?: Record<string, unknown>;
+};
```
mattzcarey, line 61: looks like a type not a schema? and same as above...

packages/agents/src/ai-react.tsx
```diff
@@ -8,17 +8,127 @@ import type {
...
+  if (params === undefined || params === null) {
+    return undefined;
+  }
+  // JSONSchemaType and plain objects are compatible with Record<string, unknown>
+  if (typeof params === "object") {
+    return params as Record<string, unknown>;
+  }
+  // Primitive values shouldn't be used as parameters, but handle gracefully
+  return undefined;
+}
```
mattzcarey: technically a boolean is a valid json schema. You want to use the Interface object on the JSONSchema type here to restrict the type to something which will go over wire. this shouldnt be necessary.

packages/agents/src/ai-react.tsx
whoiskatrin: yes yes, fixing this, my oversight

packages/agents/src/ai-react.tsx
whoiskatrin: fixed this

PR #732, opened 2025-12-16
Setup proper peer deps for zod v4

Scalahansolo:
Following the guide at https://zod.dev/library-authors to let users "bring their own version of zod"


.idea/workspace.xml
```diff
@@ -0,0 +1,113 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
threepointone: can you delete this file?

Scalahansolo: Yeah. my bad on this.

PR #750, opened 2025-12-23
dont override with undefined

deathbyknowledge:


wjvantilburg:
@deathbyknowledge seeing as this didn't affect my issue, not sure if needed. I did check it out. I noticed there were some new exposed undefined attributes, such as `title` `errorText`. Don't know if this is new with V6 just a FYI.

PR #752, opened 2025-12-24
Upgrade MCP SDK to v1.25.1

mattzcarey:


examples/mcp-worker/README.md
```diff
@@ -2,8 +2,6 @@
 
 This example demonstrates how to use `createMcpHandler` to create an unauthenticated stateless MCP server.
 
-This is THE simplest way to get started with MCP on Cloudflare.
```
deathbyknowledge, line 5: Dethroned. Maybe worth a mention here that this is now supported directly in the mcp library and a link to the new example?

Just a nit, don't have to do this now

deathbyknowledge (approved):
Very positive changes :D

PR #761, opened 2025-12-29
fix: allow returning a non-streaming reponse from `onChatMessage()`

iTrooz:
Currently, doing
```ts
  async onChatMessage(
    onFinish: StreamTextOnFinishCallback<ToolSet>,
    _options?: { abortSignal?: AbortSignal }
  ) {
    return new Response("Hello !");
  }
```
from an AIChatAgent creates an error on the frontend, because doing so is only generating a `text-delta` event, and not the `text-start` and `text-end` events that should go in-between, according to https://ai-sdk.dev/docs/ai-sdk-ui/stream-protocol#text-start-part (from my understanding ?)

This PR fixes it by adding these 2 events.

### DIFF
--- a/packages/ai-chat/src/tests/chat-persistence.test.ts
+++ b/packages/ai-chat/src/tests/chat-persistence.test.ts
@@ -5,13 +5,11 @@ import { MessageType } from "../types";
 import type { UIMessage as ChatMessage } from "ai";
 import { connectChatWS } from "./test-utils";
 
-interface ToolCallPart {
-  type: string;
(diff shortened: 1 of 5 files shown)


threepointone (commented):
actually hold up I think I want to refactor a bit more, I'll land it when happy

iTrooz:
I was planning to submit a second PR refactoring the file after this, but if you prefer to do it yourself, sure

threepointone:
let's land this and then work on the next one

Issue #718, opened 2025-12-06
"jsonSchema not initialized" intermittent error when calling agent.mcp.getAITools()


jldec:
After upgrading to latest agent sdk I noticed this intermittent error.
The agent stops handling messages after the error happens.

```
Error on server: Error: jsonSchema not initialized.
    at MCPClientManager.getAITools (/Users/jldec/cloudflare/redwood/agents-chat/node_modules/.vite/deps_worker/chunk-UBIZWKXI.js:2263:33)
    at ChatAgentAgentDO.onChatMessage (/Users/jldec/cloudflare/redwood/agents-chat/src/app/chat-agent-agent/ChatAgentAgentDO.ts:26:19)
    at /Users/jldec/cloudflare/redwood/agents-chat/node_modules/.vite/deps_worker/agents_ai-chat-agent.js:168:32
Override onError(error) to handle server errors 
```

I have an attempted repro in https://github.com/jldec/agents-chat/pull/111

That PR removes the workaround which avoids the call to `mcp.getAITools()`
My project is a bit old, and may be using some old APIs. This issue is not a blocker for me - feel free to prioritize accordingly.

threepointone:
Ok, I figured this out. Fix here: https://github.com/cloudflare/agents/pull/746

threepointone:
actually, that might not fix it, hrmm

threepointone:
A better fix here https://github.com/cloudflare/agents/pull/762

threepointone:
fixed in main

PR #762, opened 2025-12-30
eagerly load this.mcp.jsonSchema in ai chat agent

threepointone:
fixes https://github.com/cloudflare/agents/issues/718

Because of hibernation, our preloaded this.mcp.jsonSchema can be lost when the DO wakes up again. We should probably move this helper into anotherl ibrary. Until then, a patch is to eagerly load it when an ai chat agent starts up.

We also throw the error only if there's an active mcp connection.

### DIFF
--- a/.changeset/beige-melons-brake.md
+++ b/.changeset/beige-melons-brake.md
@@ -1,0 +1,12 @@
+---
+"@cloudflare/ai-chat": patch
+"agents": patch
+---
+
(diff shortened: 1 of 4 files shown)


threepointone:
there's an annoying race condition I'll have to dive into, probably in partyserver

threepointone:
fixed in main

PR #768, opened 2026-01-06
pipe SQL errors into the existing onError method using a new SqlError class

### LINKED ISSUE #767: Allow customizing logger for SQL errors
## Problem

The `sql` method in `Agent` uses `console.error` directly for logging SQL errors:

```typescript
} catch (e) {
  console.error(`failed to execute sql query: ${query}`, e);
  throw this.onError(e);
}
```

This bypasses any structured logging the application may have configured. During DO resets (e.g., code deployments), this produces unstructured logs like:

```
(error) failed to execute sql query:
    INSERT OR REPLACE INTO cf_agents_state (id, state)
    VALUES (?, ?)
   Error: Durable Object reset because its code was updated.
```

## Proposal

Allow users to customize logging behavior, either via:

1. **Override method** - `onSqlError(query: string, error: unknown)` that defaults to `console.error` but can be overridden
2. **Logger injection** - Accept a logger in Agent options/constructor
3. **Use observability** - Route through the existing `observability` interface

Option 1 is minimal and consistent with existing patterns like `onError`.

whoiskatrin:
Instead of adding a separate onSqlError hook, I implemented a cleaner approach (suggested by @threepointone) that pipes SQL errors into the existing onError method using a new SqlError class

Fixes #767

### DIFF
--- a/.changeset/spotty-donkeys-compare.md
+++ b/.changeset/spotty-donkeys-compare.md
@@ -2,4 +2,4 @@
 "agents": patch
 ---
 
-add onSqlError hook for customizable SQL error logging
+pipe SQL errors into the existing onError method using a new SqlError class


packages/agents/src/index.ts
```diff
@@ -391,7 +391,7 @@ export class Agent<
       // Execute the SQL query with the provided values
       return [...this.ctx.storage.sql.exec(query, ...values)] as T[];
     } catch (e) {
-      console.error(`failed to execute sql query: ${query}`, e);
```
threepointone, line 394: instead why don't we use our logger directly here?

threepointone: or pipe into onError

whoiskatrin: agree with piping, done that

PR #775, opened 2026-01-12
Fix React act environment setup in vitest browser tests

threepointone:
sunil: Since we were using plain react act() calls, I had to add a setup file for vitest that enabled the IS_REACT_ACT_ENVIRONMENT flag. Then I discovered that vitest-browser-react@1 has a bug where it would set the global IS_REACT_ACT_ENVIRONMENT to false after a run of it's own render (which wraps act()), so if you ran react's own act() after it, you'd get the "The current testing environment is not configured to support act(...)" error. After fixing that, react correctly identified some tests had mutations that weren't wrapped with act, so I fixed them with a small sleep helper.

copilot: Adds a setup file to properly set the IS_REACT_ACT_ENVIRONMENT global for React tests, updates vitest config to use this setup, and patches vitest-browser-react to restore the environment variable after act calls. Also refactors tests to use async act and a sleep helper for more reliable async behavior.

### DIFF
--- a/packages/ai-chat/src/react-tests/use-agent-chat.test.tsx
+++ b/packages/ai-chat/src/react-tests/use-agent-chat.test.tsx
@@ -10,6 +10,10 @@ import {
 } from "../react";
 import type { useAgent } from "agents/react";
 
+function sleep(ms: number) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
 function createAgent({ name, url }: { name: string; url: string }) {
   const target = new EventTarget();
   const baseAgent = {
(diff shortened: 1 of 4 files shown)


packages/ai-chat/src/react-tests/use-agent-chat.test.tsx
```diff
@@ -10,6 +10,10 @@ import {
 } from "../react";
 import type { useAgent } from "agents/react";
 
+function sleep(ms: number) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
```
anonrig, line 14: ps: node:timers/promises will actieve the same thing

```
await setTimeout(number)
```

threepointone: nerrrrd

PR #779, opened 2026-01-14
fix: handle invalid cron expressions in alarm handler

HueCodes:
`parseCronExpression` throws for malformed cron strings. In the alarm
  handler this prevents remaining schedules from executing and leaves
  the invalid schedule in the database, blocking future alarms.

  Wrap in try/catch and remove invalid schedules to unblock the loop.


HueCodes:
Updated to validate cron expressions at input in schedule()

PR #783, opened 2026-01-15
fix: WorkerTransport doesn't persist clientCapabilities, breaking elicitation in serverless

### LINKED ISSUE #777: WorkerTransport doesn't persist clientCapabilities, breaking elicitation in serverless
## Description

When using `WorkerTransport` with storage for MCP in a serverless environment (like Cloudflare Workers), the client capabilities are not persisted across requests. This causes elicitation to fail with:

```
Error: Client does not support form elicitation.
    at Server.elicitInput
```

## Root Cause

The `TransportState` interface only stores:
```typescript
export interface TransportState {
  sessionId?: string;
  initialized: boolean;
}
```

However, `_clientCapabilities` is stored on the `Server` instance and set during the `initialize` request. When a new serverless function instance handles a subsequent request (like a tool call that triggers elicitation), the capabilities are lost because:

1. The transport state is restored from storage ✅
2. But `_clientCapabilities` on the Server instance is not restored ❌

## Steps to Reproduce

1. Create an MCP server using `WorkerTransport` with storage
2. Register a tool that uses `server.elicitInput()`
3. Connect a client that advertises elicitation capability
4. Call the tool - elicitation fails because capabilities weren't persisted

## Expected Behavior

Client capabilities should be persisted as part of the transport state so they survive across serverless function invocations.

## Suggested Fix

Include `clientCapabilities` in `TransportState`:

```typescript
export interface TransportState {
  sessionId?: string;
  initialized: boolean;
  clientCapabilities?: ClientCapabilities; // Add this
}
```

And restore them when handling requests after initialization.

## Environment

- `@modelcontextprotocol/sdk`: 1.25.2
- `agents`: latest
- Runtime: Cloudflare Workers (wrangler dev)

Muhammad-Bin-Ali:
## Summary

Persist and restore client capabilities (`initializeParams`) across serverless restarts when using `WorkerTransport` with storage.

## Problem

When using `WorkerTransport` with storage in a serverless environment (like Cloudflare Workers), client capabilities (like elicitation support) were **not persisted** across requests. This caused elicitation and other capability-dependent features to fail after a serverless cold start with:

```
Error: Client does not support form elicitation.
```

**Root Cause:**
- `TransportState` only stored `{ sessionId, initialized }`
- `_clientCapabilities` is stored in-memory on the MCP SDK's `Server` instance
- After serverless restart, `_clientCapabilities` was `undefined` because it wasn't restored

## Solution

Store the full `initializeParams` (capabilities, clientInfo, protocolVersion) and replay the initialize request during state restoration to restore `_clientCapabilities` on the Server instance.

### Changes

**`packages/agents/src/mcp/worker-transport.ts`:**
- Added `RESTORE_REQUEST_ID` constant for the synthetic restore request ID
- Extended `TransportState` interface to include `initializeParams`
- Capture `initializeParams` during initialization before calling `saveState()`
- Replay initialize request via `this.onmessage()` during `restoreState()` to restore `_clientCapabilities`
- Filter restore response in `send()` to prevent error from missing stream

**`packages/agents/src/tests/mcp/transports/worker-transport.test.ts`:**
- Added new test suite `"Client Capabilities Persistence (Serverless Restart)"` with 5 tests covering persistence, restoration, and backward compatibility

## Breaking Changes

None. The `TransportState` interface change is backward compatible - old stored state without `initializeParams` will continue to work (capabilities just won't be restored).

## Future Improvement

The ideal solution would be to extend the [MCP TypeScript SDK](https://github.com/modelcontextprotocol/typescript-sdk) with a function to directly update `clientCapabilities` on the Server instance, rather than replaying the initialize request. This would be a cleaner approach that doesn't rely on internal implementation details.

Closes #777

🤖 Generated with [OpenCode](https://opencode.ai) using Claude

### DIFF
--- a/.changeset/three-kings-lay.md
+++ b/.changeset/three-kings-lay.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
+fix saving initialize params for stateless MCP server (effects eliciations and other optional features)


packages/agents/src/mcp/worker-transport.ts
```diff
@@ -27,6 +28,7 @@ import type {
 } from "@modelcontextprotocol/sdk/server/streamableHttp.js";
 
 const MCP_PROTOCOL_VERSION_HEADER = "MCP-Protocol-Version";
+const RESTORE_REQUEST_ID = "__restore__";
```
mattzcarey, line 31: the id here would normally be `1` I think but please check :)

packages/agents/src/mcp/worker-transport.ts
```diff
@@ -538,6 +557,23 @@ export class WorkerTransport implements Transport {
 
       this.sessionId = this.sessionIdGenerator?.();
       this.initialized = true;
+
+      const initMessage = messages.find(isInitializeRequest);
+      if (initMessage && "params" in initMessage) {
+        const params = initMessage.params as {
+          capabilities?: ClientCapabilities;
```
mattzcarey: prefer to use type here from the official SDK

packages/agents/src/mcp/worker-transport.ts
```diff
@@ -802,6 +838,10 @@ export class WorkerTransport implements Transport {
       requestId = message.id;
     }
 
+    if (requestId === RESTORE_REQUEST_ID) {
+      return;
+    }
```
mattzcarey, line 832: this should never happen if im reading this correctly. It probs means something bad is happening, should we `console.warn` here?

packages/agents/src/mcp/worker-transport.ts
mattzcarey: unless this needs to be different to what the mcp server would actually send?

packages/agents/src/mcp/worker-transport.ts
mattzcarey: my bad this is correct

PR #784, opened 2026-01-16
feat: add Props type parameter to ExportedHandler for typed ctx.props

### LINKED ISSUE #501: Make ExportedHandler support setting ExecutionContext<Props>
Here's what I like to do:

```ts
export default {
	fetch: async (request, env, ctx) => {
		const url = new URL(request.url)

		if (url.pathname === '/mcp') {
			ctx.props.baseUrl = url.origin

			const mcp = EpicMeMCP.serve('/mcp', {
				binding: 'EPIC_ME_MCP_OBJECT',
			})
			return mcp.fetch(request, env, ctx)
		}

		return new Response('Not found', { status: 404 })
	},
} satisfies ExportedHandler<Env>
```

The `satisfies ExportedHandler<Env>` makes it so I don't have to manually type my `fetch`, but I'm getting an error when setting `ctx.props.baseUrl` because `ctx.props` is unknown. Here's my workaround:

```ts
export default {
	fetch: async (request: Request, env: Env, ctx: ExecutionContext<Props>) => {
		const url = new URL(request.url)

		if (url.pathname === '/mcp') {
			ctx.props.baseUrl = url.origin

			const mcp = EpicMeMCP.serve('/mcp', {
				binding: 'EPIC_ME_MCP_OBJECT',
			})
			return mcp.fetch(request, env, ctx)
		}

		return new Response('Not found', { status: 404 })
	},
}
```

I don't like the workaround as much.

HueCodes:
Add a Props type parameter to ExportedHandler so that ctx.props can be properly typed when using `satisfies ExportedHandler<Env, Props>`.

- Add enhanced ExportedHandler<Env, Props, QueueHandlerMessage, CfHostMetadata>
- Add Props-aware handler types (ExportedHandlerFetchHandler, etc.)
- Export types from main entry point and agents/types
- Add generic params to createMcpHandler and McpAgent.serve()

Fixes #501


packages/agents/src/types.ts
mattzcarey, line 1: I'm not sure we should be typing these here. These are already defined in the output of `wrangler types` if I am not mistaken.

packages/agents/src/mcp/handler.ts
```diff
@@ -25,20 +25,23 @@ export interface CreateMcpHandlerOptions extends WorkerTransportOptions {
   transport?: WorkerTransport;
 }
 
-export function createMcpHandler(
+export function createMcpHandler<
```
mattzcarey: this seems like a good improvement independently of the types below.

packages/agents/src/mcp/handler.ts
```diff
@@ -25,20 +25,23 @@ export interface CreateMcpHandlerOptions extends WorkerTransportOptions {
...
+  Env = unknown,
+  Props = unknown
```
deathbyknowledge:
```suggestion
  Props extends Record<string, unknown> = Record<string, unknown>
```

packages/agents/src/mcp/handler.ts
```diff
@@ -25,20 +25,23 @@ export interface CreateMcpHandlerOptions extends WorkerTransportOptions {
   transport?: WorkerTransport;
 }
 
-export function createMcpHandler(
+export function createMcpHandler<
+  Env = unknown,
```
deathbyknowledge:
```suggestion
  Env extends Cloudflare.Env = Cloudflare.Env
```

packages/agents/src/mcp/handler.ts
```diff
@@ -109,13 +112,16 @@ let didWarnAboutExperimentalCreateMcpHandler = false;
 /**
  * @deprecated This has been renamed to createMcpHandler, and experimental_createMcpHandler will be removed in the next major version
  */
-export function experimental_createMcpHandler(
+export function experimental_createMcpHandler<
+  Env = unknown,
```
deathbyknowledge: Same here please

packages/agents/src/mcp/index.ts
```diff
@@ -361,7 +361,7 @@ export abstract class McpAgent<
   /** Return a handler for the given path for this MCP.
    * Defaults to Streamable HTTP transport.
    */
-  static serve(
+  static serve<Env = unknown, Props = unknown>(
```
deathbyknowledge, line 364: Same here please

packages/agents/src/types.ts
deathbyknowledge: I believe what @mattzcarey meant was that the changes to add the Props parameter was better fit for `workerd` repo which is where these types come from, instead of duplicating them here.

I've created a PR already there with the changes https://github.com/cloudflare/workerd/pull/5989 so the new types are not needed

packages/agents/src/mcp/handler.ts
```diff
@@ -25,10 +25,7 @@ export interface CreateMcpHandlerOptions extends WorkerTransportOptions {
   transport?: WorkerTransport;
 }
 
-export function createMcpHandler<
-  Env extends Cloudflare.Env = Cloudflare.Env,
-  Props extends Record<string, unknown> = Record<string, unknown>
->(
+export function createMcpHandler<Env = unknown, Props = unknown>(
```
threepointone, line 28: hmm I preferred this to default to Cloudflare.Env, why was it necessary to change it back to unknown? trouble with typescript?

PR #786, opened 2026-01-16
fix: allow callable methods to return this.state

### LINKED ISSUE #598: agent.stub types break when callable returns this.state
**Describe the bug**

In a Agent define this method

```
  @callable()
  async getInternal() {
    return this.state;
  }
```

then in a client using `useAgent` try and call this:

`const state = await agent.stub.getInternal();`

getInternal() throws a typescript error. 

if you change getInternal() to 

```
  @callable()
  async getInternal() {
    return "hello world";
  }
```

it works fine. 

**Version:**
latest

deathbyknowledge:
Fixes #598 

Callable methods returning `this.state` broke stub types because `unknown`, the default type, didn't satisfy the `SerializableReturnValue` check in RPCMethod.

Added an `IsUnknown` helper to allow `unknown` return types while still rejecting explicitly non-serializable types.

### DIFF
--- a/.changeset/five-coins-play.md
+++ b/.changeset/five-coins-play.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
+fix: allow callable methods to return this.state


mattzcarey:
dude you hero

threepointone (approved):

PR #787, opened 2026-01-16
fix mcp: improve error code detection when listing tools fails

kooba:
When discovering MCP server capabilities, listing tools (and other resources) would often fail because the error code was being read incorrectly.

MCP servers return a `-32601` (Method not found) error for unsupported capabilities, but the transport layer (e.g., Streamable HTTP) wraps these errors. The previous code only checked for `e.code === -32601`, missing cases where the code was embedded in the error message string.

This fix properly detects the error in both forms, allowing capability discovery to continue gracefully instead of failing entirely.

Changes
- Reuse existing getErrorCode helper from errors.ts for consistent error code extraction
- Check both direct error code and message string for -32601

### DIFF
--- a/.changeset/tall-mirrors-provide.md
+++ b/.changeset/tall-mirrors-provide.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
+fix mcp: improve error code detection when listing tools fails


mattzcarey:
this is very silly of the transport. Will have a look, thanks :)

packages/agents/src/mcp/client-connection.ts
```diff
@@ -663,17 +664,23 @@ export class MCPClientConnection {
...
-    return (e: { code: number }) => {
-      // server is badly behaved and returning invalid capabilities. This commonly occurs for resource templates
-      if (e.code === -32601) {
+    return (e: unknown) => {
+      // Server is badly behaved and returning invalid capabilities. This commonly occurs for resource templates.
+      // Check both the error code property and the error message for -32601 (Method not found)
+      const errorCode = getErrorCode(e);
+      const errorMessage = toErrorMessage(e);
+      const isMethodNotFound =
+        errorCode === -32601 || errorMessage.includes('"code":-32601');
```
mattzcarey, line 673: On second thoughts this feels like it will be too broad. Do you have an example of a server which breaks this? I'd prefer to fix this on the MCP SDK side.

mattzcarey:
closing this as it is being fixed in MCP v2. If you need backports please open PRs against the v1.x branch in https://github.com/modelcontextprotocol/typescript-sdk

PR #791, opened 2026-01-20
Feature request: Add structured memory primitives for better context engineering

mchockal:
### Summary

This PR introduces structured memory primitives for building context-aware agents, based on the tiered memory architecture principles from [Google's ADK whitepaper](https://developers.googleblog.com/en/architecting-efficient-context-aware-multi-agent-framework-for-production/), but adapted to agents sdk. 

**Update 2/16:** Added **experimental session primitives** to agents/experimental/memory — SQL-backed sessions, append-only event storage, ephemeral WorkingContext. This is a complete redesign of the original v0 memory primitives based on PR feedback, focusing just on the Session and Working Context first. A separate class `SessionAgent` is used to keep these changes purely experimental and independent of the `Agent` class.

### Why?

Currently, agent developers must manually manage conversation history, handle context window limits, and implement their own compaction strategies. This leads to:

- **Duplicated effort** across agent implementations
- **Inconsistent approaches** to memory management
- **Tight coupling** between conversation state and model-specific formats
- **Complex compaction logic** scattered throughout application code

These primitives provide a **model-agnostic foundation** that separates the ground truth (Session) from the computed view (WorkingContext), enabling clean abstractions for context management and (later)compaction.

---

> ⚠️ **Experimental** — this module is unstable and may change without notice. All exports carry @experimental JSDoc tags.

---

### SessionAgent — Agent subclass with session/event SQL tables

Extends Agent and creates cf_agents_sessions and cf_agents_events tables in the constructor (matching the Agent pattern for cf_agents_state, cf_agents_queues, etc.).

**RPC-safe methods** (plain serializable objects, callable from Workflows/Workers):

| Method | Description |
|--------|-------------|
| createSession(metadata?) | Creates a session row, returns ID |
| getSession(sessionId) | Returns StoredSession \| null |
| listSessions() | All sessions for this agent |
| deleteSession(sessionId) | Deletes session + all events |
| loadEvents(sessionId, opts?) | Hydrated SessionEvent[], ordered by seq. Default limit: 50 |
| appendEvents(sessionId, events) | Validates session, assigns monotonic seq, inserts rows |
| deleteEvents(sessionId, eventIds) | Deletes specific events by ID |

**Local-only methods** (class instances don't survive RPC serialization):

| Method | Description |
|--------|-------------|
| _buildWorkingContext(sessionId, opts?) | protected — loads events + builds WorkingContext |
| persistWorkingContext(sessionId, ctx) | Extracts new messages, converts to events, appends |

### WorkingContext — Ephemeral in-memory context

Built per-request from session events. Accumulates messages during an agentic loop, tracks which are new via _initialCount. Thrown away after the request.

```ts
const ctx = this._buildWorkingContext(sessionId, {
  systemInstructions: ["You are a helpful assistant."],
  limit: 100,
});
ctx.addMessage({ role: "user", content: userMessage });
// ... LLM call, tool calls ...
ctx.addMessage({ role: "assistant", content: response });
this.persistWorkingContext(sessionId, ctx); // batch-persist new messages only
```

### buildWorkingContext() — Pure function

For use outside the Agent (Workflows, Workers, tests). Takes SessionEvent[] from loadEvents() RPC and builds a WorkingContext locally.

### workersAIAdapter — Model format adapter

Stateless adapter converting WorkingContext messages to Workers AI chat completions format. Handles system instructions, structured toolCalls, and tool results. Support for other frameworks will be added as new adapters.

### Pure utility functions

- hydrateEvent(row) — SQL row → typed SessionEvent 
- dehydrateEvent(event) — SessionEvent → SQL row
- eventToMessage(event) — SessionEvent → ContextMessage 
- messageToEvent(sessionId, msg) — ContextMessage → SessionEvent 

---

## SQL Schema

```sql
CREATE TABLE cf_agents_sessions (
  id TEXT PRIMARY KEY NOT NULL,
  agent_id TEXT NOT NULL,
  created_at INTEGER NOT NULL,  -- ms since epoch
  updated_at INTEGER NOT NULL,
  metadata TEXT                 -- JSON
);

CREATE TABLE cf_agents_events (
  id TEXT PRIMARY KEY NOT NULL,
  session_id TEXT NOT NULL,
  seq INTEGER NOT NULL,         -- monotonic per session
  action TEXT NOT NULL,
  content TEXT,
  metadata TEXT,                -- JSON
  created_at INTEGER NOT NULL
);

CREATE INDEX idx_events_session_seq ON cf_agents_events(session_id, seq);
```

---

```
POST /chat
  → buildContext(sessionId)           # load completed events → WorkingContext
  → ctx.addMessage(userMessage)       # in-memory only
  → LLM call + tool loop             # accumulate in-memory
  → persistWorkingContext(sessionId)  # batch-persist user + assistant atomically
```

---

## Example: examples/session-agent/ 

A complete working example with:
- HTTP endpoints: POST /chat, GET /session, DELETE /session 
- Agentic loop with tool calling (echo + dummy-weather tools)

---

## Tests 

```
✓ adapters.test.ts (8 tests)  — Workers AI format, tool calls, full agentic loop conversation
✓ context.test.ts  (11 tests) — WorkingContext construction, new message tracking, custom mappers
✓ utils.test.ts    (26 tests) — hydrate/dehydrate roundtrips, event↔message mapping, edge cases
```

Tests run via npm run test:memory (added to the CI test pipeline).

---

## Known Limitations

- **Workers AI only** — OpenAI/Anthropic adapters planned
- **No compaction/summarization** — event schema and the current design supports it, orchestration (processors) not yet built
- **No token estimation** — bring your own estimator. Users can either   a.) work with an approximation - chars//3.  Faster, no additional cost. Optimistic approximation. We only need this to know when to trigger compaction, so 70% of model context with this estimate is still a good starting point for v0. 
b). spend one inference call with output_tokens limited to 1 to obtain the input-only token_usage_count
- **Concurrent request divergence** — concurrent LLM responses may be contextually divergent (each request sees only completed turns)
- **No Vercel AI SDK integration** — only raw env.AI.run() via adapter
- **_buildWorkingContext is protected** — subclasses need a public wrapper for helper functions

---

## Files Changed / Added

```
packages/agents/src/experimental/memory/
├── index.ts              # Barrel exports + @experimental JSDoc
├── types.ts              # EventAction, SessionEvent, ContextMessage, interfaces
├── utils.ts              # hydrateEvent, dehydrateEvent, eventToMessage, messageToEvent
├── context.ts            # WorkingContext class + buildWorkingContext pure function
├── session-agent.ts      # SessionAgent extends Agent
├── adapters/
│   ├── index.ts          # Re-exports
│   └── workers-ai.ts     # Workers AI adapter
└── __tests__/
    ├── vitest.config.ts
    ├── utils.test.ts
    ├── context.test.ts
    └── adapters.test.ts

packages/agents/package.json         # exports + test:memory script
packages/agents/scripts/build.ts     # entry point for experimental/memory
docs/memory-primitives.md            # Full documentation
examples/session-agent/              # Working example with agentic loop
```


packages/agents/src/memory/processors.ts
```diff
@@ -0,0 +1,654 @@
...
+    (context.metadata as Record<string, unknown>).cacheablePrefix =
+      context.systemInstructions.length;
+  }
+  return context;
+};
+
+/**
+ * Token limit processor - ensures context fits within token limits
+ */
+export const tokenLimitRequestProcessor: RequestProcessor<
```
deathbyknowledge: This relies on `estimateTokenCount` which just guesses tokens. Since there's quite some variation in tokenizers depending on the model at use, I'd prefer if we remove this one and let users provide their own. 

Maybe we can add "character" truncation instead?

packages/agents/src/memory/session.ts
```diff
@@ -0,0 +1,292 @@
...
+/**
+ * Session - The definitive state of a conversation
+ *
+ * A Session is the durable, structured log of an agent's interaction with a user.
+ * It maintains the ground truth of the conversation and provides methods for
+ * querying and manipulating the event stream.
+ */
+export class Session {
+  readonly metadata: SessionMetadata;
+  readonly events: Event[];
```
deathbyknowledge: Do we want to use the DO sqlite to handle storage here? Or perhaps storing events directly, kind of like an event store?

deathbyknowledge (commented):
This is not a final review yet but I'm struggling to see the bigger picture with this, it seems quite opinionated and not sure if the benefits outweigh the negatives.

Could you update one or two of the existing examples (or make a new one) to use this approach and see how if feels?

threepointone:
I'll review further on wednesday

packages/agents/src/memory/session.ts
mchockal: Good call. sqlite storage would be better, and fits better with the existing pattern. A `cf_agents_session` table might be a better approach to store user's session. Primary key will be a `session_id`, and events can be stored in order for easy retrieval of last 'N' user-agent turns in a given session, which would serve as the working-context for the current request. What are your thoughts? 

(Apologies about the late response. I missed the email notification thread for this set of comments, and only noticed earlier today )

packages/agents/src/memory/processors.ts
mchockal: True. Ideally the choice of this processor implementation would be something left to the user. That would be step 0. But it could potentially be something to include at a later stage when context compaction to keep long-conversation going becomes more common, and if there's benefits to including it in the sdk at that point. The entire `processors.ts` can be part of a later PR.

threepointone:
Some thoughts: (that opus cleaned up for me)

Thanks for putting this together - the architecture works and the problem space (context management, compaction, model-agnostic event logging) is real. I think this is worth shipping as an **experimental** feature with some adjustments.

### Overall Take

The Session/WorkingContext/ProcessorPipeline split is reasonable. For users building long-running conversations, multi-agent systems, or agents that need audit trails, this provides useful primitives that don't exist in the SDK today.

That said, most agents-sdk users are building simpler things and won't need this level of abstraction **yet**. So I'd recommend shipping this as explicitly experimental - opt-in for power users who want to explore these patterns. Maybe under agents/experimental/sessions?

### Before Merging

**1. Add clear experimental labeling**

The exports and docs should make it obvious this is experimental and the API may change:
- Consider exporting from `agents/experimental/sessions` and adding `Experimental` prefix to class names
- Add `@experimental` JSDoc tags
- Docs should lead with a warning about API instability

**2. Add at least one complete example**

The PR description has good code snippets, but there's no end-to-end example showing how to wire this into an actual agent. Something in `examples/` that demonstrates:
- Creating a Session, adding events
- Running the processor pipeline
- Calling an LLM with the resulting WorkingContext
- (Bonus) Triggering compaction after N turns

Without this, developers won't know how to connect the pieces.

**3. Document known limitations**

Be explicit about what this doesn't do (yet):
- Only supports Workers AI format (OpenAI/Anthropic adapters planned)
- Compaction requires you to provide your own summarizer
- No integration with Vercel AI SDK (yet)
- Token estimation is approximate (~chars/4) we should probably fix this before landing tbh
- Will probably break as we iterate on it (schemas, apis, what not)

### Code-Level Items

A few things to clean up:

**Token counting** — The `chars / 4` estimation can be off significantly. Either document this limitation clearly or consider making the estimator pluggable.

**`truncateToFit` algorithm** — The current implementation breaks on the first content that doesn't fit, which may not be optimal. Consider whether this matters for the use cases you're targeting.

**`readonly` modifiers** — `events`, `statistics`, and `compactionConfig` are marked `readonly` but then mutated. This is confusing - either use actual immutability or drop the `readonly` keyword.

**Event ID generation** — `Date.now()` + random suffix could theoretically collide under high throughput. Consider `crypto.randomUUID()` for guaranteed uniqueness.

**Tests** — Would be good to have basic coverage for:
- Session serialization/deserialization roundtrip
- Processor pipeline ordering
- Compaction filtering logic

### Summary

Ship it as experimental with:
- [ ] Clear experimental labeling in exports and docs
- [ ] At least one working example in `examples/`
- [ ] Known limitations documented
- [ ] Basic test coverage
- [ ] Fix the `readonly` confusion

We'll learn more from real usage than from more design iteration. Just make sure the "experimental" signal is loud enough that people don't accidentally depend on API stability.

packages/agents/src/memory/session.ts
threepointone: yeah do sqlite is fine. might need to work around 2mb row limitation, just consider that.

mchockal:
Thanks for detailed feedback! I'll update the PR with the changes over the next 2-3 days and have it ready for review by Friday latest.

mchockal:
@deathbyknowledge , @threepointone , @whoiskatrin : Could I get another review on these changes? Apologies for the huge delay. A few points:

**Scope:** This PR currently focuses on Session and WorkingContext, but it establishes the necessary foundation to expand into processors.

**Token Estimation:** This was discussed already, but the default tokenEstimation function (which can be overriden by the user to bring their own) can be designed to be "optimistically conservative" (chars//3) to trigger auto-compaction. Since auto-compaction usually kicks in at 70–80% capacity, and this function doesn't rely on any additional calls (to tokenization endpoint), this still remains a solid starting point imo.

For a more accurate, "Workers AI native" approach, the ideal path would be requesting a `countToken()` endpoint from the Workers AI team (similar to `toMarkdown`, preferably exposed via bindings too). While users _can_ host their own tokenization endpoints via Hugging Face libraries, built-in support would be much more efficient.

token count endpoint available from different commercial llm vendors for reference, 

1. Gemini :  https://ai.google.dev/gemini-api/docs/tokens
2. OpenAI:  https://developers.openai.com/api/docs/guides/token-counting
3. Anthropic: https://platform.claude.com/docs/en/build-with-claude/token-counting

I’d like to open this up for discussion before moving forward with any changes on that front, preferably as a different PR.
 
**Notes on tokenizers library (relevant for open source models, ones hosted on WorkersAI) :**
image
- This cloudflare [blog](https://blog.cloudflare.com/cloudflares-most-efficient-ai-inference-engine/) (not sure how relevant this is today) talks about using the rust crate for tokenizers from Huggingface for models hosted on workers AI 
- Library on github: https://github.com/huggingface/tokenizers

whoiskatrin:
Closing this as we've added memory primitives last month

PR #796, opened 2026-01-25
chore: make playground example fallback to WorkersAI when OpenAI key is not set

iTrooz:
This PR allows you to use the `playground` example without an OpenAI key, by adding a fallback to WorkersAI

I'm already using it locally to test other PRs I'm submitting to this repository since I don't want to be billed

I figured the `playground` example was the most appropriate for this additions, since it's the most general


threepointone:
very nice, thanks!

PR #798, opened 2026-01-26
feat(react): add enabled prop to useAgent hook for conditional connections

### LINKED ISSUE #533: Feature request: Conditional agent connection in react
When using the `useAgent` hook in react it would be useful to be able to provide a `enabled?: boolean` parameter as part of the agent options (much like React Query does) where the websocket connection is only started once the enabled parameter returns true.

Suhaib3100:
## Summary

This PR adds an `enabled` prop to the `useAgent` hook, allowing for conditional agent connections. This is a feature that was requested in issue #533.

## Problem

Currently, when using `useAgent`, the connection is always established immediately when the hook is called. This makes it difficult to implement patterns like:
- Only connecting when a user is authenticated
- Deferring connection until a component is visible (lazy loading)
- Feature flag based conditional connections

## Solution

Added an `enabled` optional boolean prop to `UseAgentOptions` that controls whether the agent connection should be established.

### Implementation Details

1. **Type Definition**: Added `enabled?: boolean` to `UseAgentOptions` with JSDoc documentation explaining the behavior
2. **Default Value**: `enabled` defaults to `true` to maintain backward compatibility
3. **Initial State**: Uses `startClosed: !enabled` to control the initial connection state via PartySocket
4. **State Transitions**: Added a `useEffect` with a `prevEnabledRef` to handle dynamic changes:
   - When `enabled` transitions from `false` to `true`: calls `agent.reconnect()`
   - When `enabled` transitions from `true` to `false`: calls `agent.close()`

### Usage Example

```tsx
const [isAuthenticated, setIsAuthenticated] = useState(false);

// Connection is only established when isAuthenticated is true
const agent = useAgent({
  agent: MyAgent,
  enabled: isAuthenticated,
});
```

## Changes

- `packages/agents/src/react.tsx`: Added `enabled` prop to `UseAgentOptions` type and implementation in `useAgent` hook
- `packages/agents/src/react-tests/enabled-prop.test.ts`: Added comprehensive test suite (14 tests) covering:
  - Type definitions
  - Connection lifecycle (startClosed logic)
  - State transition logic
  - Integration with other options
  - Common use cases (auth-based, feature flag, lazy loading)

## Testing

All 14 new tests pass:
```
✓ useAgent enabled prop (issue #533) > Type definitions > should accept enabled as an optional boolean prop
✓ useAgent enabled prop (issue #533) > Type definitions > should default enabled to true when not specified
✓ useAgent enabled prop (issue #533) > Connection lifecycle > should start closed when enabled is false
... (11 more tests)
```

## Checklist

- [x] Tests added
- [x] Types updated
- [x] Documentation (JSDoc) added
- [x] Changeset added
- [x] Backward compatible (defaults to `true`)

Closes #533

### DIFF
--- a/.changeset/add-enabled-prop.md
+++ b/.changeset/add-enabled-prop.md
@@ -1,0 +1,24 @@
+---
+"agents": minor
+---
+
+feat: add `enabled` prop to `useAgent` hook for conditional connections
+
+This adds an `enabled` optional prop to `useAgent` that allows conditionally connecting to an Agent. When `enabled` is `false`, the connection will not be established. When it transitions from `false` to `true`, the connection is established. When it transitions from `true` to `false`, the connection is closed.
+
+This is useful for:
+
+- Auth-based conditional connections (only connect when authenticated)
+- Feature flag based connections
+- Lazy loading patterns
+
+Usage:
+
+```tsx
+const agent = useAgent({
+  agent: "my-agent",
+  enabled: isAuthenticated // only connect when authenticated
+});
+```
+
+Closes #533
--- a/packages/agents/src/react-tests/enabled-prop.test.ts
+++ b/packages/agents/src/react-tests/enabled-prop.test.ts
@@ -1,0 +1,210 @@
+import { describe, expect, it, beforeEach } from "vitest";
+
+/**
+ * Tests for the `enabled` prop in useAgent hook.
+ *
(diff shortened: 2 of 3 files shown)


threepointone:
fixed in https://github.com/cloudflare/partykit/pull/315

PR #828, opened 2026-02-02
Move email utilities to agents/email

threepointone:
Move email-specific resolvers and types into a dedicated agents/email subpath and update imports across docs, examples, and tests. Add a ./email export in package.json and update package changelog to a minor release. Keep createHeaderBasedEmailResolver exported at the package root for upgrade discoverability but make it throw a migration error with guidance and a docs link; update examples/docs/tests to import other resolvers from agents/email.


threepointone:
This doesn't need a new changeset

PR #839, opened 2026-02-04
Invalidate query cache on disconnect to fix stale auth tokens

### LINKED ISSUE #836: Feature request: always re‑run async query on connect/reconnect (awaited)
Thanks for the recent TTL fix in #726 ! We’re still seeing frequent auth failures on reconnect because useAgent only re‑runs async query when the cache key changes or the TTL expires. Reconnects alone don’t trigger a refresh, so a short‑lived token can be reused on reconnect and rejected by the server.

**Current behavior (post‑ #726 ):**

- Async query is re‑run only when cacheKey changes or TTL expires.
- Reconnect does not re‑run query.
- Changing the query function identity is ignored if the cache entry is still valid.

**Request:**

Could we add an option to always invoke async query on every connect/reconnect, and await it before connecting? That would let apps safely do:

```typescript
query: async () => ({ token: await getIdToken(true) })
```

…without needing manual token refresh logic or time‑based cache hacks.

Lowering TTL helps, but it’s still time‑based while reconnects are triggered by events and then spaced by backoff timers. If a reconnect happens between TTL refreshes, the cached async query result is still reused, so we can still reconnect with an expired token even with aggressive TTLs.

**Possible API ideas:**

- `queryCacheMode: "ttl" | "always"`
- `alwaysRefreshQuery: true`
- `cacheTtl: 0` meaning “no caching” (currently it effectively caches forever)

**Why this helps:**

It would eliminate noisy auth failures on reconnect for apps using short‑lived tokens and make reconnect behavior more deterministic.

**Environment**

- agents: Observed in 0.3.5, behavior still present in 0.3.8.
- React 19
- `useAgent` on the client, Cloudflare Worker backend

whoiskatrin:
## Summary
- Invalidates the query cache entry when the WebSocket connection closes
- Ensures async query functions (e.g., for auth tokens) are re-executed on reconnect
- Eliminates auth failures caused by reusing stale cached tokens after disconnect

## Problem
When using `useAgent` with an async `query` function for authentication:
```js
query: async () => ({ token: await getIdToken(true) })
```

The cached query result was reused on reconnect even if the token had expired. Reconnects don't change the cache key or wait for TTL expiration, so stale tokens were sent to the server causing auth failures.

## Solution
By calling `deleteCacheEntry(cacheKey)` in the `onClose` handler, we invalidate the cache on disconnect. When the socket reconnects, the async query will be re-executed to fetch a fresh token.

This is simpler than adding new API options (`queryCacheMode`, `alwaysRefreshQuery`, etc.) and provides the correct behavior by default - reconnects should always use fresh auth data.

Fixes #836

### DIFF
--- a/packages/agents/src/react.tsx
+++ b/packages/agents/src/react.tsx
@@ -311,6 +311,10 @@ export function useAgent<State>(
   // Track cache invalidation to force re-render when TTL expires
   const [cacheInvalidatedAt, setCacheInvalidatedAt] = useState<number>(0);
 
+  // Disable socket while waiting for async query to refresh after disconnect
+  const isAsyncQuery = query && typeof query === "function";
+  const [awaitingQueryRefresh, setAwaitingQueryRefresh] = useState(false);
+
   // Get or create the query promise
   // biome-ignore lint/correctness/useExhaustiveDependencies: cacheInvalidatedAt intentionally forces re-evaluation when TTL expires
   const queryPromise = useMemo(() => {
@@ -392,6 +396,13 @@ export function useAgent<State>(
     }
   }
 
+  // Re-enable socket after async query resolves
+  useEffect(() => {
+    if (awaitingQueryRefresh && resolvedQuery !== undefined) {
+      setAwaitingQueryRefresh(false);
+    }
+  }, [awaitingQueryRefresh, resolvedQuery]);
+
   // Store identity in React state for reactivity
   const [identity, setIdentity] = useState({
     name: options.name || "default",
@@ -439,8 +450,11 @@ export function useAgent<State>(
         ...restOptions
       };
 
+  const socketEnabled = !awaitingQueryRefresh && (restOptions.enabled ?? true);
+
   const agent = usePartySocket({
     ...socketOptions,
+    enabled: socketEnabled,
     onMessage: (message) => {
       if (typeof message.data === "string") {
         let parsedMessage: Record<string, unknown>;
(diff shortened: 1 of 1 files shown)


packages/agents/src/react.tsx
```diff
@@ -534,6 +543,15 @@ export function useAgent<State>(
...
       setIdentity((prev) => ({ ...prev, identified: false }));
 
+      // Invalidate the query cache and trigger re-render so reconnection will
+      // re-run the async query with fresh tokens. We must both:
+      // 1. Delete the cache entry so getCacheEntry() returns undefined
+      // 2. Bump cacheInvalidatedAt to trigger useMemo to recompute queryPromise
+      // Without the state bump, the component keeps the old resolved query value
+      // and PartySocket would reconnect with stale query params.
+      deleteCacheEntry(cacheKeyRef.current);
+      setCacheInvalidatedAt(Date.now());
```
ghost: Main remaining concern is a race - the socket’s reconnect loop can fire with the old URL/query before the re‑render (triggered by cache invalidation + state bump) completes. The only obvious way I see to eliminate the race is to pause reconnects while the query refresh happens.

Looks like we can use `enabled` (UsePartySocketOptions) to handle this? So a pattern like:

- onClose: setEnabled(false) + invalidate cache + bump state
- after query resolves / re-render: setEnabled(true)

Idea being to guarantee no reconnect attempt happens before the new query is ready.

ghost:
Could we add an integration test where the query returns a different token on each invocation, then assert which token appears on the first reconnect? Probably makes sense to make it intentionally race‑y by:

1. Delaying the async query and
2. Forcing fast reconnects (set minReconnectionDelay/maxReconnectionDelay low).

Then assert which token the first reconnect used:

```
let calls = 0;
query: async () => { await delay(100); return { token: `t-${++calls}` } };

force reconnect();
assert first reconnect used "t-2" (not "t-1")
```

Without that timing pressure, the test could pass even if a race still exists, simply because the reconnect happens after the refresh render completes. This would validate the behavior the PR claims (fresh token on reconnect), beyond just cache deletion.

packages/agents/src/react.tsx
whoiskatrin: yes, that's valid, we might just do that in the end

whoiskatrin:
@aurewill-gavel give the latest version a go, please, if there will be more gaps, we will revisit

ghost:
@whoiskatrin Logic looks sound. That test I mentioned would be ideal but obv your decision whether or not to add it.

threepointone:
let's land this, I suspect suspense might help out here a bit. and because it's not an api change, we can revisit and make it stronger if we need to.

PR #841, opened 2026-02-04
fix: escape html in external oauth error message

mattzcarey:


mattzcarey:
happy with this now :)

examples/mcp-client/src/server.ts
```diff
@@ -1,26 +1,14 @@
...
     // Optionally configure OAuth callback. Here we use popup-closing behavior since we're opening a window on the client
     this.mcp.configureOAuthCallback({
-      customHandler: (result: MCPClientOAuthResult) => {
-        if (result.authSuccess) {
-          return new Response("<script>window.close();</script>", {
-            headers: { "content-type": "text/html" },
-            status: 200
-          });
-        } else {
-          const safeError = JSON.stringify(result.authError || "Unknown error");
```
mattzcarey, line 15: remove this hot garbage

packages/agents/src/mcp/client-connection.ts
```diff
@@ -95,6 +95,7 @@ export type MCPDiscoveryResult = {
 export class MCPClientConnection {
   client: Client;
   connectionState: MCPConnectionState = MCPConnectionState.CONNECTING;
+  connectionError: string | null = null;
```
mattzcarey, line 98: add this string as a first class thing which can be bubbled up

packages/agents/src/mcp/client.ts
```diff
@@ -1,4 +1,5 @@
 import type { Client } from "@modelcontextprotocol/sdk/client/index.js";
+import escapeHtml from "escape-html";
```
mattzcarey, line 2: this library is already in the bundle as a transient dep so thats handy

packages/agents/src/mcp/client.ts
```diff
@@ -44,6 +45,13 @@ export type MCPServerOptions = {
   };
 };
 
+/**
+ * Result of an OAuth callback request
+ */
+export type MCPOAuthCallbackResult =
```
mattzcarey, line 51: this type didnt exist but it was nice to define it

packages/agents/src/mcp/client.ts
```diff
@@ -187,6 +195,19 @@ export class MCPClientManager {
     );
   }
 
+  private failConnection(
```
mattzcarey, line 198: this standardises what happens when we have a connection so we know which one to mark as failed. extracting this logic means I dont have to repeat it for the 3 occasions this happens in

packages/agents/src/mcp/client.ts
```diff
@@ -695,89 +715,61 @@ export class MCPClientManager {
       throw new Error(`Could not find serverId: ${serverId}`);
     }
```
mattzcarey, line 716: in these cases we cant get a connection id to mark the connection as failed so not much choice other to error out. In future this might change.

packages/agents/src/mcp/client.ts
```diff
@@ -695,89 +715,61 @@ export class MCPClientManager {
...
     const conn = this.mcpConnections[serverId];
-    if (!conn.options.transport.authProvider) {
-      throw new Error(
-        "Trying to finalize authentication for a server connection without an authProvider"
-      );
-    }
 
-    const authProvider = conn.options.transport.authProvider;
-    authProvider.serverId = serverId;
+    try {
```
mattzcarey, line 721: from here we have a connection id so any errors are wrapped in this try catch and we can fail the connection gracefully.

packages/agents/src/mcp/client.ts
```diff
@@ -695,89 +715,61 @@ export class MCPClientManager {
...
-    if (error) {
-      return {
-        serverId,
-        authSuccess: false,
-        authError: errorDescription || error
-      };
-    }
+      if (error) {
+        // Escape external OAuth error params to prevent XSS
+        throw new Error(escapeHtml(errorDescription || error));
```
mattzcarey, line 740: the big baddie xss

packages/agents/src/mcp/client.ts
```diff
@@ -695,89 +715,61 @@ export class MCPClientManager {
...
-      };
-    }
+      // Already authenticated - just return success
+      if (
+        conn.connectionState === MCPConnectionState.READY ||
+        conn.connectionState === MCPConnectionState.CONNECTED
+      ) {
+        this.clearServerAuthUrl(serverId);
+        return { serverId, authSuccess: true };
+      }
```
mattzcarey, line 754: happy path making sure to clear the auth url. Maybe we also need to clear the auth error also but it shouldnt be set.

packages/agents/src/mcp/client.ts
```diff
@@ -695,89 +715,61 @@ export class MCPClientManager {
...
+        );
+      }
 
-    try {
       await authProvider.consumeState(state);
       await conn.completeAuthorization(code);
       await authProvider.deleteCodeVerifier();
       this.clearServerAuthUrl(serverId);
+      conn.connectionError = null;
       this._onServerStateChanged.fire();
```
mattzcarey, line 767: other happy path.

packages/agents/src/mcp/client.ts
```diff
@@ -695,89 +715,61 @@ export class MCPClientManager {
...
-
-      return {
-        serverId,
-        authSuccess: false,
-        authError: errorMessage
-      };
+      return { serverId, authSuccess: true };
+    } catch (err) {
+      const message = err instanceof Error ? err.message : String(err);
+      return this.failConnection(serverId, message);
```
mattzcarey, line 772: extracted logic

packages/agents/src/tests/mcp/client-manager.test.ts
```diff
@@ -399,7 +399,7 @@ describe("MCPClientManager OAuth Integration", () => {
       expect(result.authError).toBe("User denied access");
     });
 
-    it("should throw error for callback without code or error", async () => {
+    it("should fail connection for callback without code or error", async () => {
```
mattzcarey, line 402: this makes sense since we can tie it to a connection id. This would be the result of a server misbehaving so we wouldnt want to get stuck in authentication

packages/agents/src/tests/mcp/client-manager.test.ts
```diff
@@ -496,7 +496,7 @@ describe("MCPClientManager OAuth Integration", () => {
       expect(result.serverId).toBe(serverId);
     });
 
-    it("should error when callback received for connection in failed state", async () => {
+    it("should fail connection when callback received for connection in failed state", async () => {
```
mattzcarey, line 499: same here, developer should manually flip the connection state if they want to initiate a retry

PR #849, opened 2026-02-05
feat(codemode): make model configurable instead of hardcoded gpt-4.1

### LINKED ISSUE #794: CodeMode LLM pinned to gpt-4.1
https://github.com/cloudflare/agents/blob/main/packages/codemode/src/ai.ts#L70

It appears CodeMode is pinned to `gpt-4.1`. I was surprised to find this wasn't configurable. It would be beneficial to allow the `codemode` function to take in a model, same as `streamText` for example.

```ts
import { codemode } from "agents/codemode/ai";

const {system, tools} = codemode({
  model: openai("gpt-5"),  //<---- new functionality
  system: "You are a helpful assistant",
  tools: {
    // tool definitions 
  },
  // ...config
})

const stream = streamText({
  model: openai("gpt-5"),
  system,
  tools,
  messages: [
    { role: "user", content: "Write a function that adds two numbers" }
  ]
})
```

Muhammad-Bin-Ali:
Resolves #794 

### Problem
`experimental_codemode` hardcodes `openai("gpt-4.1")` in the `generateObject` call ([`packages/codemode/src/ai.ts:70`](https://github.com/cloudflare/agents/blob/main/packages/codemode/src/ai.ts#L70)), preventing users from choosing their own model.
### Solution
Add a required `model: LanguageModel` parameter to `experimental_codemode`, allowing callers to pass any AI SDK-compatible model.
### Changes
- **`packages/codemode/src/ai.ts`**
  - Added `model: LanguageModel` to the options type
  - Replaced `openai("gpt-4.1")` with `options.model` in the `generateObject` call
  - Removed unused `@ai-sdk/openai` import
- **`examples/codemode/src/server.ts`**
  - Passes the existing `model` variable to the `codemode()` call
### Usage
```ts
import { experimental_codemode } from "@cloudflare/codemode/ai";
import { openai } from "@ai-sdk/openai";
const { prompt, tools } = await experimental_codemode({
  model: openai("gpt-5"), // <-- now configurable
  prompt: "You are a helpful assistant",
  tools: {
    /* ... */
  }
  // ...other config
});
```
### Notes
- This is a **breaking change** since `model` is required. Callers must now explicitly provide a model. Chosen to make it a breaking change so users are explicitly aware of the model they're using. 
- The `dist/` artifacts (`ai.d.ts`, `ai.js`) need to be regenerated via `npm run build` in `packages/codemode/`.

_This PR description was generated by [OpenCode](https://opencode.ai)_

### DIFF
--- a/packages/codemode/src/ai.ts
+++ b/packages/codemode/src/ai.ts
@@ -8,6 +8,7 @@ import {
 } from "zod-to-ts";
 import { getAgentByName } from "agents";
 import { env, WorkerEntrypoint } from "cloudflare:workers";
+import { openai } from "@ai-sdk/openai";
 
 function toCamelCase(str: string) {
   return str
@@ -35,12 +36,12 @@ export class CodeModeProxy extends WorkerEntrypoint<
 }
 
 export async function experimental_codemode(options: {
-  model: LanguageModel;
   tools: ToolSet;
   prompt: string;
   globalOutbound: Fetcher;
   loader: WorkerLoader;
   proxy: Fetcher<CodeModeProxy>;
+  model?: LanguageModel;
 }): Promise<{
   prompt: string;
   tools: ToolSet;
@@ -67,7 +68,7 @@ export async function experimental_codemode(options: {
     execute: async ({ functionDescription }) => {
       try {
         const response = await generateObject({
-          model: options.model,
+          model: options.model ? options.model : openai("gpt-4.1"),
           schema: z.object({
             code: z.string()
           }),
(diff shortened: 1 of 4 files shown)


threepointone:
This should be a patch changeset

packages/codemode/src/ai.ts
```diff
@@ -67,7 +67,7 @@ export async function experimental_codemode(options: {
     execute: async ({ functionDescription }) => {
       try {
         const response = await generateObject({
-          model: openai("gpt-4.1"),
+          model: options.model,
```
threepointone: This should default to the previous default model

packages/codemode/src/ai.ts
```diff
@@ -36,6 +35,7 @@ export class CodeModeProxy extends WorkerEntrypoint<
 }
 
 export async function experimental_codemode(options: {
+  model: LanguageModel;
```
threepointone: This should be an optional

PR #850, opened 2026-02-06
fix: surface MCP OAuth errors to browser instead of silently swallowing

Muhammad-Bin-Ali:
## Surface MCP OAuth errors instead of swallowing them
Previously, when an OAuth callback failed (e.g. user denied access, invalid state, missing code), `handleCallbackRequest` would throw an unhandled exception that bubbled up as a raw 500 in the browser with no useful information.
### Problem
`handleCallbackRequest` mixed two patterns: returning `MCPOAuthCallbackResult` objects for some paths and throwing exceptions for others. The thrown errors were never caught by the Agent's callback handler, so they became opaque 500s. OAuth errors from the authorization server (like `access_denied`) also weren't routed through `failConnection()`, leaving the connection stuck in `"authenticating"` instead of properly transitioning to `"failed"`.
### What changed
- Extracted a `validateCallbackRequest()` method in `MCPClientManager` that returns structured results instead of throwing. Validation errors with an active connection now route through `failConnection()` so state properly transitions to `"failed"`.
- The default OAuth callback handler redirects back to the origin instead of rendering a hardcoded HTML error page. Auth errors are surfaced to clients via WebSocket broadcast (`onMcpUpdate` / `server.error`).
- `MCPOAuthCallbackResult` and `MCPClientOAuthResult` types updated so `serverId` is optional on error results.
- Updated the mcp-client example to handle auth failures in `customHandler` and display errors inline in the UI.
---
*Generated by [OpenCode](https://opencode.ai)*


whoiskatrin:
cc @mattzcarey

whoiskatrin:
/bonk can you review this PR please

examples/mcp-client/src/server.ts
```diff
@@ -4,11 +4,31 @@ export class MyAgent extends Agent {
...
+  <p style="color:#666;">${error}</p>
+  <p style="color:#999;font-size:0.85em;">This window will close automatically...</p>
+  <script>setTimeout(() => window.close(), 3000);</script>
+</body>
+</html>`,
+          {
+            headers: { "content-type": "text/html" },
+            status: 400
+          }
+        );
```
mattzcarey: I dont want to show the error in a script tag, possibility of XSS attack

packages/agents/src/index.ts
```diff
@@ -3226,9 +3227,21 @@ export class Agent<
...
+      // This fires onServerStateChanged event which triggers broadcast
+      result = await this.mcp.handleCallbackRequest(request);
+    } catch (err) {
+      // handleCallbackRequest throws when it can't find the in-memory connection
+      // (e.g., after hibernation). Catch here so we return a proper HTTP error
+      // response instead of letting it bubble as a raw 500.
+      const errorMessage = err instanceof Error ? err.message : String(err);
+      const url = new URL(request.url);
+      const state = url.searchParams.get("state");
+      const serverId = state?.split(".")[1] ?? "unknown";
```
mattzcarey: I think there is a utility function to do this. otherwise it would be good to make one.

packages/agents/src/index.ts
```diff
@@ -3294,7 +3307,28 @@ export class Agent<
...
+</body>
+</html>`,
+        {
+          status: 400,
+          headers: { "content-type": "text/html; charset=utf-8" }
+        }
+      );
+    }
+
+    // Default success: redirect to base URL
```
mattzcarey: I would prefer developers use the first class way of accessing errors on the client through the react hook, rather than show a hardcoded error page. 

https://github.com/cloudflare/agents/blob/main/site/ai-playground/src/app.tsx#L95

examples/mcp-client/src/client.tsx
```diff
@@ -162,6 +162,17 @@ function App() {
                 />
                 {server.state} (id: {id})
               </div>
+              {server.state === "failed" && server.error && (
```
mattzcarey, line 167: this is good :)

packages/agents/src/index.ts
mattzcarey: we could also put this try catch inside `handleCallbackRequest` since then we dont need to mutate the result at this level.

mattzcarey:
/bonk review this. Understand the previous review comments and explain the fixes. Explain the design requirements of this fix

packages/agents/src/index.ts
```diff
@@ -4053,7 +4053,8 @@ export class Agent<
       }
     }
 
-    // Default: redirect to base URL
+    // Default: redirect back to the application.
+    // Auth errors are surfaced to clients via the WebSocket broadcast (onMcpUpdate / server.error).
```
mattzcarey: probs remove this :)

packages/agents/src/mcp/client.ts
```diff
@@ -106,11 +106,9 @@ export type MCPClientOAuthCallbackConfig = {
   customHandler?: (result: MCPClientOAuthResult) => Response;
 };
 
-export type MCPClientOAuthResult = {
-  serverId: string;
-  authSuccess: boolean;
-  authError?: string;
-};
+export type MCPClientOAuthResult =
```
mattzcarey, line 108: love this

threepointone:
cleaned up a few things, looks good to me now. landing.

PR #868, opened 2026-02-08
Fix MCP OAuth callback URL leaking instance name

### LINKED ISSUE #854: MCP OAuth callback URL leaks instance name when `sendIdentityOnConnect` is `false`
## Description
When an Agent connects to a remote MCP server with OAuth (`addMcpServer`), the OAuth callback URL unconditionally exposes the agent class name and instance name, even when `sendIdentityOnConnect: false` is set to prevent identity leakage. Additionally, the callback URL is hardcoded to the standard `/agents/...` routing pattern with no way to override it, which is problematic when using custom URL routing.

## Problem
 Instance name leaks in OAuth callback URL despite `sendIdentityOnConnect: false`
The [docs recommend](https://developers.cloudflare.com/agents/api-reference/routing/#disabling-identity-for-security) setting `sendIdentityOnConnect: false` when instance names contain sensitive data (session IDs, internal user IDs):
```js 
class SecureAgent extends Agent {
  static options = { sendIdentityOnConnect: false };
}
```

However, the MCP OAuth callback URL construction [at index.ts:3082](https://github.com/cloudflare/agents/blob/main/packages/agents/src/index.ts#L3082) has no such check

During the OAuth flow, this URL is:
1. Sent to the remote MCP server's OAuth provider as the redirect URI
2. Used as a browser redirect — visible in the address bar and browser history

This undermines the security intent of sendIdentityOnConnect: false.

## Expected Behavior
1. When `sendIdentityOnConnect: false` is set, the callback URL should not expose the raw instance name. Options could include using an opaque/hashed identifier, or at minimum logging a warning.
2. addMcpServer should accept an optional callbackUrl (or callbackPath) override so developers using custom routing can specify a callback path that works with their routing setup. For example:
```js
await this.addMcpServer("my-server", "https://remote-mcp.example.com", {
  callbackPath: "/mcp-callback",  // instead of /agents/{agent}/{name}/callback
});
```

This would let developers route the callback to the correct agent instance server-side, just as they do with basePath + getAgentByName:
```js
if (url.pathname === "/mcp-callback") {
  const session = await getSession(request);
  const agent = await getAgentByName(env.SecureAgent, session.userId);
  return agent.fetch(request);
}
```

threepointone:
Closes #854

## Problem

When an agent uses `sendIdentityOnConnect: false` to protect sensitive instance names (session IDs, user IDs), the MCP OAuth callback URL still exposes the raw instance name:

```
https://example.com/agents/secure-agent/secret-user-id-123/callback
```

This URL is sent to the remote OAuth provider as `redirect_uri`, displayed in the browser address bar, and saved in browser history — completely undermining the security intent. Additionally, the hardcoded `/agents/{class}/{name}/callback` pattern doesn't work for developers using `basePath` routing with `getAgentByName`.

## Design decisions

**`callbackPath` is required when `sendIdentityOnConnect: false`.**
Rather than silently hashing or obfuscating the instance name (which would require complex coordination between the routing layer and DO storage), we make the developer explicitly provide a safe callback path. This is consistent with how `basePath` already works — if you've opted into custom routing to protect instance names, you handle the routing yourself via `getAgentByName`. The error message is actionable and tells the developer exactly what to do.

**`callbackPath` is optional when `sendIdentityOnConnect: true`.**
The default auto-constructed URL (`/agents/{class}/{name}/callback`) still works for developers who haven't opted out of identity exposure. This keeps the simple case simple. Developers can still provide `callbackPath` if they prefer cleaner URLs or a different routing pattern.

**Callback detection uses `state` parameter with origin + pathname verification.**
Previously, `isCallbackRequest()` checked `url.includes("/callback")` as a loose heuristic — this would break with custom callback paths (e.g., `/mcp-oauth-return`). The `state` query parameter (format: `{nonce}.{serverId}`) is now the primary mechanism for identifying which server a callback belongs to. As defense-in-depth, we also verify that the request's origin and pathname match the registered `callback_url` for that server, preventing unrelated GET requests with a `state` param from being intercepted.

## Changes

- **`AddMcpServerOptions` type** — new `callbackPath?: string` field with JSDoc documenting it should be a plain path (no query strings or fragments)
- **`addMcpServer()`** — resolves `callbackPath`, enforces it when `sendIdentityOnConnect: false`, uses it for URL construction when provided. Also normalizes trailing slashes on `callbackHost` to prevent double-slash URLs.
- **`isCallbackRequest()`** — replaced loose `url.includes("/callback")` substring check with state-based server ID matching plus origin + pathname verification against the stored `callback_url`
- **Tests** — new unit tests for custom callback path detection, pathname-based matching, and origin defense-in-depth in `client-manager.test.ts`. New integration tests in `add-mcp-server.test.ts` verifying the `callbackPath` enforcement (throws when `sendIdentityOnConnect: false` without `callbackPath`, does not throw when `callbackPath` is provided).

## Notes for reviewers

- The `isCallbackRequest` change is the subtlest part. The old `/callback` substring check is replaced by a two-layer approach: (1) extract `serverId` from the `state` param and confirm it exists in storage, (2) verify the request's origin + pathname match the stored `callback_url` for that server. This is both more precise and more permissive than before — it works with any callback path while preventing false-positive interception.

- The runtime enforcement (throw when `sendIdentityOnConnect: false` without `callbackPath`) can't be done at the type level since `sendIdentityOnConnect` is on `static options` and `callbackPath` is on the `addMcpServer` call — TypeScript can't connect those two. The runtime error is clear and fails fast.

### DIFF
--- a/packages/agents/src/index.ts
+++ b/packages/agents/src/index.ts
@@ -269,6 +269,14 @@ export type MCPServer = {
 export type AddMcpServerOptions = {
   /** OAuth callback host (auto-derived from request if omitted) */
   callbackHost?: string;
+  /**
+   * Custom callback URL path — bypasses the default `/agents/{class}/{name}/callback` construction.
+   * Required when `sendIdentityOnConnect` is `false` to prevent leaking the instance name.
+   * When set, the callback URL becomes `{callbackHost}/{callbackPath}`.
+   * The developer must route this path to the agent instance via `getAgentByName`.
+   * Should be a plain path (e.g., `/mcp-callback`) — do not include query strings or fragments.
+   */
+  callbackPath?: string;
   /** Agents routing prefix (default: "agents") */
   agentsPrefix?: string;
   /** MCP client options */
@@ -3342,12 +3350,15 @@ export class Agent<
         }
       | undefined;
 
+    let resolvedCallbackPath: string | undefined;
+
     if (
       typeof callbackHostOrOptions === "object" &&
       callbackHostOrOptions !== null
     ) {
       // New API: options object as third parameter
       resolvedCallbackHost = callbackHostOrOptions.callbackHost;
+      resolvedCallbackPath = callbackHostOrOptions.callbackPath;
       resolvedAgentsPrefix = callbackHostOrOptions.agentsPrefix ?? "agents";
       resolvedOptions = {
         client: callbackHostOrOptions.client,
@@ -3358,6 +3369,15 @@ export class Agent<
       resolvedCallbackHost = callbackHostOrOptions;
       resolvedAgentsPrefix = agentsPrefix ?? "agents";
       resolvedOptions = options;
+    }
+
+    // Enforce callbackPath when sendIdentityOnConnect is false
+    if (!this._resolvedOptions.sendIdentityOnConnect && !resolvedCallbackPath) {
+      throw new Error(
+        "callbackPath is required in addMcpServer options when sendIdentityOnConnect is false — " +
+          "the default callback URL would expose the instance name. " +
+          "Provide a callbackPath and route the callback request to this agent via getAgentByName."
+      );
     }
 
     // If callbackHost is not provided, derive it from the current request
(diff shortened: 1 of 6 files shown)


packages/agents/src/index.ts
threepointone, line 2423: these "errors" are unhandled rejections we need to solve in our vitest library, but aren't breaking tests, it's ok

PR #870, opened 2026-02-09
Speed up CI: split jobs, fix Playwright caching, clean up workflows

threepointone:
(or attempt to, anyway)

## Summary

- **Split PR workflow into parallel jobs**: `check`, `test`, and `publish-preview` run concurrently instead of sequentially. Lint/typecheck feedback arrives without waiting for tests.
- **Fix Playwright installation**: Removed `playwright install --with-deps chromium` from `postinstall` so it doesn't run on every `npm ci`. Test job installs browsers conditionally (only downloads on cache miss, only installs system deps on cache hit). Non-test jobs skip Playwright entirely via `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD`.
- **Add concurrency groups**: PR workflows cancel in-progress runs when new commits are pushed. Release workflow queues instead of overlapping.
- **Add path filtering**: PR workflow skips entirely for docs-only, design, markdown, and changeset-only changes.
- **Merge `pkg-pr-new.yml` into `pullrequest.yml`**: One fewer workflow triggered per PR, one shared concurrency group.
- **Remove `sync-docs.yml`**: No longer needed.
- **Improve Claude workflows**: Code review uses `use_sticky_comment`, inline comments, Sonnet model, and `--max-turns` cap. Interactive assistant gets `contents: write`, npm/git tool access, and a Node.js setup step so it can actually build and test.
- **Add timeouts**: All jobs now have explicit `timeout-minutes` to prevent runaway runners.

### DIFF
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -63,10 +63,6 @@ jobs:
         if: steps.playwright-cache.outputs.cache-hit != 'true'
         run: npx playwright install --with-deps chromium
 
-      - name: Install Playwright system dependencies
-        if: steps.playwright-cache.outputs.cache-hit == 'true'
-        run: npx playwright install-deps chromium
-
       - run: npm run build
       - run: CI=true npm run test
 


threepointone:
@whoiskatrin deleted the sync flow, let's setup and try bonk for a while

PR #882, opened 2026-02-10
fix: pass client tool schemas during tool continuation

alexanderjacobsen:
## Summary

- Fix multi-step client tool calling: when `autoContinueAfterToolResult` triggers a server continuation, the `onChatMessage` callback was called without `clientTools`, causing the LLM to fail with `tool-output-error` when trying to call another client tool
- Store client tool schemas from the initial request (`_lastClientTools`) and pass them during continuation
- Add `toolResolutionTrigger` on the client side to handle race conditions where new tool calls arrive while the resolution effect is busy

## Test plan

- [ ] Configure a chat agent with `autoContinueAfterToolResult: true` and multiple client-side tools
- [ ] Send a prompt that triggers multiple sequential client tool calls (e.g. "change background to green then black")
- [ ] Verify all tool calls execute successfully without `tool-output-error`
- [ ] Verify single tool calls still work correctly
- [ ] Verify server-side tools are unaffected

### DIFF
--- a/packages/ai-chat/src/tests/client-tools-continuation.test.ts
+++ b/packages/ai-chat/src/tests/client-tools-continuation.test.ts
@@ -1,0 +1,314 @@
+import { env } from "cloudflare:test";
+import { describe, it, expect } from "vitest";
+import { MessageType } from "../types";
+import type { UIMessage as ChatMessage } from "ai";
+import { connectChatWS } from "./test-utils";
+import { getAgentByName } from "agents";
+
+describe("Client tools continuation", () => {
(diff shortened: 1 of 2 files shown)


whoiskatrin:
@alexanderjacobsen can you please add the changeset and address valid bonk review comments?

alexanderjacobsen:
@whoiskatrin Should be done now

whoiskatrin:
@alexanderjacobsen can we add some tests for this? otherwise happy to ship in the next release

alexanderjacobsen:
@whoiskatrin Added tests now

whoiskatrin (approved):

PR #886, opened 2026-02-11
feat(ai-chat): forward custom body fields to onChatMessage options

whoiskatrin:
## Summary

- Forward custom body fields from client `sendMessage` / `prepareSendMessagesRequest` to the server-side `onChatMessage` handler via `options.body`
- Previously, any extra fields in the request body (beyond `messages` and `clientTools`) were silently dropped during parsing
- Add `body?: Record<string, unknown>` to `OnChatMessageOptions` type and extract custom fields using rest destructuring

## Motivation

Users sending custom data alongside chat messages (e.g. model selection, temperature, metadata) had no way to access it server-side. The data was parsed but only `messages` and `clientTools` were extracted — everything else was discarded.

Ref: https://discord.com/channels/595317990191398933/1344688586146189475/1470897271649341491

## Usage

```typescript
// Client
await sendMessage(
  { role: "user", parts: [{ type: "text", text }] },
  { body: { model: "gpt-4", temperature: 0.7 } }
);

// Server
async onChatMessage(onFinish, options) {
  const model = options?.body?.model ?? "default-model";
}
```

## Changes

- `packages/ai-chat/src/index.ts` — Add `body` to `OnChatMessageOptions`; extract custom fields via rest destructuring and pass to `onChatMessage`
- `packages/ai-chat/src/tests/worker.ts` — Capture `options.body` in test agent for verification
- `packages/ai-chat/src/tests/custom-body.test.ts` — Tests for body forwarding, exclusion of internal fields, and undefined when empty

### DIFF
--- a/.changeset/forward-body-to-onchatmessage.md
+++ b/.changeset/forward-body-to-onchatmessage.md
@@ -1,5 +1,5 @@
 ---
-"@cloudflare/ai-chat": minor
+"@cloudflare/ai-chat": patch
 ---
 
 Forward custom body fields from client requests to `onChatMessage` options


threepointone:
bonk is right, let's make this a patch, and then ship it

threepointone:
wait why is it showing as major

threepointone:
sigh I'll fix it. but also let's make this a patch?

whoiskatrin:
/bonk can you make a changeset a patch please?

threepointone:
/bonk can you file an issue to followup on the nit

PR #889, opened 2026-02-11
migrate scheduling from zod/3 to zod

### LINKED ISSUE #816: fix scheduling nits
move from zod/3 to zod 
fix type errors when you do that 
try not to need providerOptions

deathbyknowledge:
- Moved `scheduleSchema` import from `zod/v3` to `zod`. The `/v3` subpath only exists in zod 4's compat layer and would break for users on zod 3 (which the peer dep allows)
- Restructured `scheduleSchema` from a flat object with optional fields to a `z.discriminatedUnion` on `when.type`, so each variant only has the fields it needs

I haven't found a way to remove `providerOptions` that satisfies OpenAI's strict schema :/ 

closes #816

### DIFF
--- a/.changeset/many-turkeys-sit.md
+++ b/.changeset/many-turkeys-sit.md
@@ -2,9 +2,9 @@
 "agents": patch
 ---
 
-Fix scheduling schema compatibility with zod v3 and OpenAI strict mode.
+Fix scheduling schema compatibility with zod v3 and improve schema structure.
 
 - Change `zod/v3` import to `zod` so the package works for users on zod v3 (who don't have the `zod/v3` subpath).
-- Replace `z.discriminatedUnion` with a flat object using nullable fields. OpenAI strict structured outputs rejects `oneOf` (produced by discriminated unions), so this removes the need for `providerOptions: { openai: { strictJsonSchema: false } }`.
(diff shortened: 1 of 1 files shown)


threepointone:
Could you test that this actually works with models from openai/gemini(this one caused problems last time I tried a discriminated union)/and maybe something good from workers ai?

deathbyknowledge:
> Could you test that this actually works with models from openai/gemini(this one caused problems last time I tried a discriminated union)/and maybe something good from workers ai?

Thanks for calling that out, switching from OpenRouter to OpenAI did error. I tried a bunch of ways to get around OpenAI's strict schema but no luck, so I've added `providerOptions` back and left the changes to be just upgrade to `zod`.

threepointone:
Gemini passed?

deathbyknowledge:
> Gemini passed?

Yep, Gemini passes fine. One _possible_ solution would be to make all fields nullable instead of optional (which is what OpenAI doesn't allow), but that makes the type of all fields be `string | null`. I've tested current approach and nullable and both work, but nullable works without `providerOptions`.

threepointone:
/bonk do a review and leave a compliment for @deathbyknowledge

threepointone:
please update the changeset before landing this

PR #897, opened 2026-02-12
Fix client tool schemas lost after DO restart

alexanderjacobsen:
## Summary

Follow-up to #882. Client tool schemas (`_lastClientTools`) are stored in-memory on the Durable Object and lost on restart. This breaks `autoContinueAfterToolResult` when the user refreshes between a tool call and its confirmation — the continuation calls `onChatMessage` without tool schemas and the AI responds saying it doesn't have the tools.

**Fix:** Include `clientTools` in the `CF_AGENT_TOOL_RESULT` WebSocket message. The client is the source of truth for its tools and always has the current schemas, so re-sending is both simpler and more correct than server-side persistence.

## Changes

- **`types.ts`** — Add optional `clientTools` field to `CF_AGENT_TOOL_RESULT` incoming message
- **`react.tsx`** — Include `extractClientToolSchemas()` in all `CF_AGENT_TOOL_RESULT` sends (`sendToolOutputToServer`, automatic tool resolution, `addToolResultAndSendMessage`)
- **`index.ts`** — Extract `clientTools` from tool result data and pass to `onChatMessage` during continuation
- **`tests/worker.ts`** — Capture `clientTools` from `onChatMessage` options for test assertions
- **`tests/client-tools-reconnect.test.ts`** — Test that tools from `CF_AGENT_TOOL_RESULT` are used for continuation, plus backwards compatibility test

## Test plan

- [x] All 39 existing tests pass
- [x] New test: client tools from `CF_AGENT_TOOL_RESULT` are passed to `onChatMessage` during continuation (simulates reconnect after DO restart)
- [x] New test: backwards compatibility — continuation works without `clientTools` in `CF_AGENT_TOOL_RESULT`
- [x] Manual test: tool with `requiresConfirmation` works after page refresh

### DIFF
--- a/.changeset/fix-client-tools-reconnect.md
+++ b/.changeset/fix-client-tools-reconnect.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/ai-chat": patch
+---
+
+Fix client tool schemas lost after DO restart by re-sending them with CF_AGENT_TOOL_RESULT
--- a/packages/ai-chat/src/index.ts
+++ b/packages/ai-chat/src/index.ts
@@ -482,7 +482,8 @@ export class AIChatAgent<
 
         // Handle client-side tool result
         if (data.type === MessageType.CF_AGENT_TOOL_RESULT) {
-          const { toolCallId, toolName, output, autoContinue } = data;
+          const { toolCallId, toolName, output, autoContinue, clientTools } =
+            data;
 
           // Apply the tool result
           this._applyToolResult(toolCallId, toolName, output).then(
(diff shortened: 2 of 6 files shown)


whoiskatrin:
@alexanderjacobsen this PR should have already the fix for it? https://github.com/cloudflare/agents/pull/899

give it a go

alexanderjacobsen:
@whoiskatrin Tested #899 locally and the bug is still there

whoiskatrin:
My bad, this is not related to streaming, just tools. LGTM

Issue #858, opened 2026-02-07
D1 write action is temporarily disabled


kentcdodds:
Heyo friends 👋 

**Describe the bug**

I'm working on a project starter [epicflare](https://github.com/epicweb-dev/epicflare) that includes authenticated MCP with agents out of the box. It's really neat!

I'm trying to put together a demo for it: [Home Energy](https://github.com/kentcdodds/home-energy). Just a simple app that adds up the amount of energy your home appliances are rated for (to help you estimate how much electricity they could use).

Unfortunately, I keep running into D1 "write action is temporarily disabled" errors whenever ChatGPT attempts to call a tool that writes to the database :disappointed: Sometimes it'll go through and when it does, it's awesome (I literally take a picture of an appliance label, tell ChatGPT to add it to my Home Energy and it does, it's so cool). But most of the time it fails. I often will also get a `waitUntil()` error around the same time (probably related?).

I tried adding [some graceful retrying to handle this](https://github.com/kentcdodds/home-energy/commit/f0aedab34c4a85358725d0d2b4921a76a32952b5), but that did not work.

**To Reproduce**

1. Deploy https://github.com/kentcdodds/home-energy to your own account
2. Create an account on your version
3. [Add the app to ChatGPT](https://www.epicai.pro/mcp-now-available-on-chatgpt-com-lbx6m)
4. Watch the logs
5. Tell it to add a refrigerator that takes 150 watts

If it works the first time, tell it to add a toaster, eventually you'll have the problem.

**Expected behavior**

I expect ChatGPT to add it without issue

**Screenshots**

<img width="866" height="971" alt="Image" src="https://github.com/user-attachments/assets/0105b525-deff-4d8d-8196-e71e51c7e582" />

<img width="857" height="752" alt="Image" src="https://github.com/user-attachments/assets/93e753bc-03b4-4467-bc9a-3aae9c9abc10" />

**Version:**

^0.3.10 (latest as of right now)

**Additional context**

You can find the code for this tool call here: https://github.com/kentcdodds/home-energy/blob/main/mcp/tools.ts

<details>
<summary>the waitUntil error</summary>

```
{
  "level": "warn",
  "message": "waitUntil() tasks did not complete within the allowed time after invocation end and have been cancelled. See: https://developers.cloudflare.com/workers/runtime-apis/context/#waituntil",
  "$workers": {
    "truncated": false,
    "event": {
      "request": {
[...]

PR #898, opened 2026-02-12
fix: add keepalive ping to POST SSE response streams

jvg123:
## Summary

The GET SSE handler in `WorkerTransport` sets up a 30-second keepalive interval that writes `event: ping` to prevent proxies and infrastructure from closing idle connections. The POST SSE handler was missing this keepalive, causing POST response streams to be silently dropped during long-running operations like MCP `tools/call`.

This is a one-line logical fix (adding the same `setInterval` pattern already used in the GET handler) plus proper cleanup in the stream mapping.

## Problem

When a client sends a POST request that returns an SSE stream (e.g., `tools/call`), the response stream has no keepalive pings. During long-running tool executions (even just a few seconds), proxies, load balancers, and infrastructure can close the idle SSE connection, causing the client to never receive the response.

**GET handler** (line 350) — has keepalive:
```typescript
const keepAlive = setInterval(() => {
  try {
    writer.write(encoder.encode("event: ping\ndata: \n\n"));
  } catch {
    clearInterval(keepAlive);
  }
}, 30000);
```

**POST handler** (line 636) — was missing keepalive entirely.

### Impact

This affects any MCP server using `WorkerTransport` with Streamable HTTP transport. Clients report errors like:
- Claude: "No result received from client-side tool execution"
- ChatGPT: "TypeError: fetch failed"
- MCP SDK: "MCP error -32001: Request Timeout"

Cloudflare Workers analytics show these as `responseStreamDisconnected` with short wall times (1-10s), confirming the SSE stream dies during tool execution rather than timing out.

## Fix

Added the same keepalive `setInterval` to the POST handler, with `clearInterval` in the cleanup function. The diff is minimal — it mirrors the existing GET handler pattern exactly.

## Test plan

- [x] Added `post-keepalive.test.ts` with 3 tests:
  - GET SSE stream sets up keepalive interval (existing behavior, passes before and after)
  - POST SSE stream sets up keepalive interval (**failed before fix, passes after**)
  - POST keepalive cleanup clears the interval on stream close
- [x] All 467 existing tests still pass (`npx vitest run --project workers`)
- [x] `npm run format` passes (oxfmt)

### DIFF
--- a/packages/agents/src/tests/mcp/transports/post-keepalive.test.ts
+++ b/packages/agents/src/tests/mcp/transports/post-keepalive.test.ts
@@ -15,7 +15,7 @@ import { z } from "zod";
  * from closing idle connections during long-running operations.
  */
 describe("WorkerTransport POST stream keepalive", () => {
-  let setIntervalSpy: ReturnType<typeof vi.spyOn>;
+  let setIntervalSpy = vi.spyOn(globalThis, "setInterval");
 
   const createSlowToolServer = () => {
     const server = new McpServer(


mattzcarey (approved):
LGTM not sure how I missed that

mattzcarey:
This could be the cause of https://github.com/cloudflare/agents/issues/858

mattzcarey:
@bonk can you make sure check script passes. Make the smallest change possible and commit it.

whoiskatrin:
/bonk fix the npm run check issue

PR #900, opened 2026-02-13
handle data parts in AIChat

### LINKED ISSUE #575: Custom data parts (data-*) not persisted to message history
## Description

Custom data message parts (e.g., `data-CustomDataV1`) created via `writer.write()` in `onChatMessage` are broadcast to clients via WebSocket in real-time, but are **not added to `message.parts`** array, causing them to disappear from message history after streaming completes.

## Steps to Reproduce

1. Create a custom AIChatAgent that extends `AIChatAgent`
2. In `onChatMessage`, use `createUIMessageStream` with `writer.write()` to send a custom data part:

```typescript
const stream = createUIMessageStream({
  execute: async ({ writer }) => {
    // Send custom data part
    writer.write({
      type: 'data-CustomDataV1',
      data: JSON.stringify({ title: 'Test', content: 'Hello' })
    });
  }
});

return createUIMessageStreamResponse({ stream });
```

3. Observe client receives the data via WebSocket during streaming
4. After streaming completes, check persisted messages
5. Custom data part is missing from `message.parts` array

## Expected Behavior

Custom data parts should be added to `message.parts` array and persisted in message history, similar to how `text`, `file`, `source-url`, and other standard parts are handled.

## Actual Behavior

- Custom data parts ARE broadcast via `_broadcastChatMessage()` (line after switch statement)
- Custom data parts are NOT added to `message.parts` (no case in switch statement handles them)
- When `persistMessages()` is called, custom data parts are missing
- UI loses access to the data after re-render or reconnection

## Root Cause

In `packages/agents/src/ai-chat-agent.ts`, the `_reply()` method has a `switch (data.type)` statement that handles known SSE chunk types. PR #499 added comprehensive handling for standard types but left a comment:

```typescript
// Do we want to handle data parts?
```

Custom data types fall through the switch without being added to `message.parts`.

## Proposed Solution

Add a default case or explicit handling for data parts in the switch statement:

```typescript
// In the switch statement, after existing cases:
default: {
  // Handle custom data parts (data-*)
  if (data.type.startsWith('data-')) {
    message.parts.push({
      type: data.type,
      data: data.data,
      ...(data.id ? { id: data.id } : {})
    });
  }
  break;
}
```

This would maintain parity with the AI SDK's data part streaming behavior and ensure custom data persists in message history.

## Environment

- `agents` version: 0.2.8
- Issue appears to affect all versions after PR #499

## Related

- PR #499 "handle all message types in the streaming handler"
- Issue #497 "AIChatAgent response handler is incomplete"
- AI SDK documentation: https://ai-sdk.dev/docs/ai-sdk-ui/streaming-data

## Impact

This prevents using custom data parts for structured data streaming (e.g., domain-specific objects, progressive results, typed data) in agent applications, forcing workarounds like:
- Using metadata (loses streaming capability)
- Using transient parts (loses persistence)
- Manual message reconstruction (defeats purpose of framework)

Thank you for considering this enhancement!

deathbyknowledge:
Add support for `data-*` stream parts in the shared message builder. Data parts are developer-defined typed JSON blobs that the AI SDK attaches to messages alongside text — they were previously silently dropped.

### What changed

**`packages/ai-chat/src/message-builder.ts`** — Handle `data-*` chunk types in `applyChunkToParts`. This is the shared parser used by both the server (`_streamSSEReply`) and client (stream resume / cross-tab broadcast), so data parts are handled consistently everywhere. Transient parts (`transient: true`) are broadcast but not persisted. Non-transient parts support reconciliation by type+id — a second chunk with the same composite key updates `.data` in-place.

**`packages/ai-chat/src/react.tsx`** — Forward the `onData` callback for `data-*` chunks on the stream resume and cross-tab broadcast codepaths. The transport path (new messages from this tab) already goes through the AI SDK's internal pipeline which invokes `onData`. This is the correct way to consume transient data parts on the client since they are not in `message.parts`.

**`packages/ai-chat/src/tests/message-builder.test.ts`** — 9 new tests covering append, id preservation, type+id reconciliation, non-matching ids/types, id-less parts, transient skip, and coexistence with other part types.

**`examples/resumable-stream-chat/`** — Updated to demonstrate three data part patterns: `data-sources` (reconciliation), `data-thinking` (transient via `onData`), and `data-usage` (persisted metadata). The client uses `UIMessage<unknown, { sources; thinking; usage }>` for fully typed data parts — no manual casts needed.

### What about the items from [this comment](https://github.com/cloudflare/agents/pull/900#issuecomment-3902010012)?

1. **`onData` callback** — Addressed. Extracted from options, stored in a ref, forwarded to `useChat`, and explicitly called for the resume/broadcast paths.
2. **File part size validation** — Removed per "don't skip persistence" feedback. PR #899 handles oversized rows with compression at the storage layer.
3. **Typed data parts on `useAgentChat`** — Already works. The `ChatMessage` generic flows through `useAgentChat` → `useChat` end-to-end. `dataPartSchemas` is available via the inherited `ChatInit` options. The example now demonstrates this — `part.type === "data-sources"` narrows `part.data` to the correct shape automatically.

Closes #575

### DIFF
--- a/docs/chat-agents.md
+++ b/docs/chat-agents.md
@@ -573,11 +573,11 @@ export class ChatAgent extends AIChatAgent {
 
 ### Three Patterns
 
-| Pattern | How | Persisted? | Use case |
-| --- | --- | --- | --- |
-| **Reconciliation** | Same `type` + `id` → updates in-place | Yes | Progressive state (searching → found) |
-| **Append** | No `id`, or different `id` → appends | Yes | Log entries, multiple citations |
-| **Transient** | `transient: true` → not added to `message.parts` | No | Ephemeral status (thinking indicator) |
+| Pattern            | How                                              | Persisted? | Use case                              |
+| ------------------ | ------------------------------------------------ | ---------- | ------------------------------------- |
+| **Reconciliation** | Same `type` + `id` → updates in-place            | Yes        | Progressive state (searching → found) |
+| **Append**         | No `id`, or different `id` → appends             | Yes        | Log entries, multiple citations       |
+| **Transient**      | `transient: true` → not added to `message.parts` | No         | Ephemeral status (thinking indicator) |
 
 Transient parts are broadcast to connected clients in real time but excluded from SQLite persistence and `message.parts`. Use the `onData` callback to consume them.
 


threepointone:
I think I handled this in my pr, can you check?

threepointone:
Oh this is different

deathbyknowledge:
> Oh this is different

Yes, I think your PR is missing data parts but you have added a compression for messages too large and I added a simpler size check to skip persistence. Should I remove that?

threepointone:
Yeah don't skip persistence

threepointone:
cursor told me this: 

What's still NOT covered even after this PR:
Looking at the AI SDK's full data parts feature set:
onData callback on the client side — The AI SDK's useChat has an onData callback that fires for all data parts as they stream (including transient ones). Our useAgentChat hook doesn't expose this. The example works around it by checking isStreaming to conditionally render the transient thinking part, but that's a UI hack — once streaming ends, the part vanishes from the render, not because the framework removed it, but because the render check hides it. A proper onData callback would give users a clean way to handle transient parts (show toasts, update external state, etc.).
File part size validation — The PR description mentions it, but I don't see it in the diff for index.ts. The diff only shows the data-* handling in the default case.
Typed data parts on useAgentChat — The AI SDK supports useChat<MyUIMessage> with typed data part schemas, so message.parts gives you type-safe access to your custom data-weather, data-sources, etc. The example uses manual type assertions (as DataPart<SourcesData>). Our useAgentChat would need to pass through the generic type parameter to fully support this

threepointone:
I'll review this later this evening, let's get it in shape by then?

threepointone:
what parts of that feedback I shared were addressed? could you update the pr description? could you add tests (unit/e2e as relevant) thanks! I'll review tomorrow morning before landing.

deathbyknowledge:
> what parts of that feedback I shared were addressed? could you update the pr description? could you add tests (unit/e2e as relevant) thanks! I'll review tomorrow morning before landing.
- Now `onData` does trigger client side
- File part size validation. It was removed entirely because it was added here #899
- Typed data parts already work, I've added the proper types in the example to showcase it

I've added tests too 🫡

threepointone:
love this, banger. will land once CI passes.

PR #934, opened 2026-02-18
Allow simultaneous inputs on AIChat

### LINKED ISSUE #24: ai chat agent should handle multiple inputs coming in at the same time

deathbyknowledge:
Fixes #24: AIChatAgent now handles multiple messages arriving simultaneously by queuing them and processing one at a time.

## Problem

When multiple chat messages arrived before the first response completed, the `onChatMessage → _reply` cycles would race, causing:
- Interleaved streaming responses
- Potential message corruption
- Unpredictable behavior

## Solution

Added a request queue that serializes chat message processing:

- **`_chatRequestQueue`** — Array of pending work functions
- **`_chatRequestInFlight`** — Guard to prevent concurrent processing  
- **`_enqueueChatRequest()`** — Adds work to queue, kicks off drain if idle
- **`_drainChatRequestQueue()`** — Processes queue items sequentially

## Changes

- Queue mechanism for serializing `onChatMessage` calls
- Tool result continuations now use the queue instead of `waitForStream()` + 500ms timeout hack
- Tool approval continuations similarly refactored
- Removed dead `_streamCompletionPromise` / `_streamCompletionResolve` code (no longer needed)
- Queue cleared on `destroy()` and `CF_AGENT_CHAT_CLEAR`

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/934" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/packages/ai-chat/src/tests/concurrent-messages.test.ts
+++ b/packages/ai-chat/src/tests/concurrent-messages.test.ts
@@ -1,0 +1,163 @@
+import { describe, it, expect } from "vitest";
+import type { UIMessage as ChatMessage } from "ai";
+import { MessageType } from "../types";
+import { connectChatWS, isUseChatResponseMessage } from "./test-utils";
+
+describe("Concurrent Message Handling", () => {
+  it("processes multiple simultaneous messages in order without interleaving", async () => {
+    const room = crypto.randomUUID();
+    const { ws } = await connectChatWS(`/agents/test-chat-agent/${room}`);
+
+    const receivedResponses: Array<{
+      id: string;
+      body: string;
+      done: boolean;
+      timestamp: number;
+    }> = [];
+
+    // Track when all 3 responses are complete
+    let completedCount = 0;
+    let resolveAll: () => void;
+    const allDone = new Promise<void>((res) => {
+      resolveAll = res;
+    });
+
+    const timeout = setTimeout(() => {
+      throw new Error(
+        `Timeout: only received ${completedCount}/3 complete responses`
+      );
+    }, 10000);
+
+    ws.addEventListener("message", (e: MessageEvent) => {
+      const data = JSON.parse(e.data as string);
+      if (isUseChatResponseMessage(data)) {
+        receivedResponses.push({
+          id: data.id,
+          body: data.body,
+          done: data.done,
+          timestamp: Date.now()
+        });
+        if (data.done) {
+          completedCount++;
+          if (completedCount === 3) {
+            clearTimeout(timeout);
+            resolveAll();
+          }
+        }
+      }
+    });
+
+    // Send 3 messages as fast as possible (simulating rapid user input)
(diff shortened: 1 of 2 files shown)


threepointone:
May I please have some tests

threepointone:
should we update the ai-chat example for this? what's the ux like? also is this behaviour alsways desirable / should it be a flag?

deathbyknowledge:
> May I please have some tests

Donezo. Added one to check response order of new messages and one to make sure the queue empties on clear

mattzcarey:
Matt's AI review:

**Critical concerns:**

1. **No abort on clear/close** — The queue is cleared but the currently in-flight request keeps running. After a `CF_AGENT_CHAT_CLEAR`, the active request can still write messages post-clear. Need to abort it or gate on a generation token.

2. **Errors are swallowed** — Previously `_tryCatchChat` was awaited directly. Now errors inside queued handlers are only logged in `_drainChatRequestQueue` and never propagate to the call site. This changes behavior for anything expecting failures to surface (metrics, retries, WS handler lifecycle).

3. **Stream completion timing** — The old explicit wait for stream completion was removed. If `_reply` resolves before the stream is fully persisted, the next queued message could start too early, re-introducing the interleaving bug the queue was intended to prevent. Worth verifying `_reply` resolves only after stream end/persistence.

4. **Silent message loss on clear** — `this._chatRequestQueue.length = 0` drops queued requests with no response or error sent back to the client. Could lead to "lost" user inputs with no feedback.

**Test gaps:**

- Ordering test only checks `done` messages, not that partial *chunks* aren't interleaved — could false-positive if chunks interleave but finish in order
- Clear test doesn't assert that `req-1` was actually dropped or that no response arrives for it after clear
- Random delay in `TestChatAgent` could cause flakes — deterministic delays tied to message content would be more robust

mattzcarey:
looks good otherwise. preemptively stamping but some of the above could be good to finalise.

mattzcarey (approved):

threepointone:
> should we update the ai-chat example for this? what's the ux like? also is this behaviour alsways desirable / should it be a flag? 

Still looking for answers on this

deathbyknowledge:
> should we update the ai-chat example for this? what's the ux like? also is this behaviour alsways desirable / should it be a flag?

My bad I never saw this comment. There are no changes needed to examples because the UX hasn't been changed at all, the changes are only to the AIChat `private` internals.

Right now if you send 2 messages you'll get interleaved stream chunks in the same stream and the data race produces garbage outputs so I don't think adding a flag for it is worth it.

That being said, we might want to eventually add a flag to decide *when* queued messages are processed (after the first tool response or after the final LLM response). Right now only final LLM is available so I think that can be done in a different PR?

threepointone:
hokay. I'll review this later today.

PR #947, opened 2026-02-19
ci: publish all non-private packages to pkg.pr.new in PR workflow

### LINKED ISSUE #923: setup pkg pr new to publish all our packages, not just `agents`

whoiskatrin:
## Summary

Fixes #923 by updating the PR preview publish step to publish all non-private packages in `packages/`, instead of only `packages/agents`.

## Changes

- Updated `.github/workflows/pullrequest.yml`
- Replaced single-package publish command with a loop over `packages/*/package.json`
- Skip packages where `private: true`
- Publish each public package via `npx pkg-pr-new publish <package-dir>`

## Why

The current workflow only publishes one package preview, which misses the rest of the SDK packages in this monorepo. This ensures PR previews cover all publishable packages.

### DIFF
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -83,17 +83,4 @@ jobs:
       - run: npm run build
 
       - name: Publish packages to pkg.pr.new
-        run: |
-          for pkg_json in packages/*/package.json; do
-            if node -e "const pkg=require('./' + process.argv[1]); process.exit(pkg.private ? 0 : 1)" "$pkg_json"; then
-              echo "Skipping private package: $pkg_json"
-              continue
-            fi
-
-            pkg_dir=$(dirname "$pkg_json")
-            echo "Publishing $pkg_dir"
-            (
-              cd "$pkg_dir"
-              npx pkg-pr-new publish --comment create
-            )
-          done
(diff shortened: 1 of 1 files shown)


threepointone:
Tell me, will this pick up the right dependency? Like, will the preview build of ai-chat pick up the preview build of agents? Do we need to do anything special there?

.github/workflows/pullrequest.yml
```diff
@@ -82,5 +82,18 @@ jobs:
...
+            if node -e "const pkg=require('./' + process.argv[1]); process.exit(pkg.private ? 0 : 1)" "$pkg_json"; then
+              echo "Skipping private package: $pkg_json"
+              continue
+            fi
+
+            pkg_dir=$(dirname "$pkg_json")
+            echo "Publishing $pkg_dir"
+            (
+              cd "$pkg_dir"
+              npx pkg-pr-new publish --comment create
```
threepointone: Will this add n comments? Isn't there a first class way to do al at once? Feels like pkg or might already support something here instead of us running this script, am I mistaken?

whoiskatrin: Check DMs lol

whoiskatrin: Should be solved now

Issue #957, opened 2026-02-20
AIChatAgent: assistant messages duplicated in SQLite on every saveMessages() round-trip


alexander-zuev:
## Issue

Every assistant message gets stored twice in the DO's SQLite — once with the server-generated `assistant_*` ID, once with the client's nanoid ID.

After 3 user messages, SQLite has 8 rows (should be 6):

```
rowid  id                                 role       text
-----  ---------------------------------  ---------  --------------------------------------------------
1      MOhQOEEhu3YnXgsx                   user
2      assistant_1771621652018_po6druuea   assistant  Hello! How can I help you today?
3      RJHGHXxn1iP0pOi7                   assistant  Hello! How can I help you today?          ← DUP
4      kwSt140SvyNPigp5                   user
5      assistant_1771621789109_s9rq2j0va   assistant  Sure! Everything seems to be working fine...
6      q01cA3vA2iT0g8yT                   assistant  Sure! Everything seems to be working fine... ← DUP
7      mFvYne67dIJ4YJjx                   user
8      assistant_1771621793779_ka5a8yav8   assistant  Yes, I'm doing great...
```

Rows 2+3 and 5+6 have identical content, different IDs. Row 8 not yet duplicated (no subsequent \`saveMessages()\` round-trip).

## Root cause

The server generates \`assistant_*\` IDs for assistant messages. The client (\`useAgentChat\`) independently generates nanoid IDs for the same messages during SSE streaming. When the client calls \`saveMessages()\`, both IDs arrive as separate rows.

The existing dedup mechanisms (\`_mergeIncomingWithServerState\` and \`_resolveMessageForToolMerge\`) only reconcile by \`toolCallId\` — plain text assistant messages have no \`toolCallId\`, so they pass through unmatched and get \`INSERT\`'d as duplicates.

## Proposed fix

**Option A:** Client uses server-generated assistant_* ID from the SSE stream instead of generating its own nanoid.

**Option B:** Client stops sending assistant messages back via saveMessages() entirely. The server already has them — it generated them. Only new user messages need to round-trip.

Either fix eliminates the need for content-based reconciliation.

## Reproduction

Minimal repro (code based on [Build a Chat Agent](https://developers.cloudflare.com/agents/getting-started/build-a-chat-agent/)):

https://github.com/alexander-zuev/ai-chat-message-duplication-repro

```bash
git clone https://github.com/alexander-zuev/ai-chat-message-duplication-repro
cd ai-chat-message-duplication-repro
[...]

PR #958, opened 2026-02-20
fix(ai-chat) prevent duplicate assistant persistence on history round-trips

whoiskatrin:
## Summary

Fix for #957.

This PR addresses duplicate assistant rows being persisted when the client resends full chat history with local assistant IDs that differ from server-persisted IDs.

## Root cause

`persistMessages()` previously upserted by `message.id`.

For plain-text assistant messages (no `toolCallId`), when a client round-tripped history using a different local assistant ID, the server treated it as a new message and inserted a duplicate row.

Tool messages were already protected by `_resolveMessageForToolMerge`, but plain assistant text messages were not.

### 1) Repro test added
- **File:** `packages/ai-chat/src/tests/merge-server-state.test.ts`
- Added test:
  - `reuses server assistant IDs for plain text messages to avoid duplicates`
- The test reproduces the issue:
  1. Persist user + assistant(server ID)
  2. Persist full history again with same assistant content but different client ID
  3. Assert no duplicate assistant row and server ID is reused

### 2) ID reconciliation for non-tool assistant messages
- **File:** `packages/ai-chat/src/index.ts`
- `_mergeIncomingWithServerState()` now:
  1. keeps existing tool-output merge behavior
  2. runs assistant ID reconciliation against current server history

- Added helpers:
  - `_reconcileAssistantIdsWithServerState(...)`
  - `_hasToolCallPart(...)`
  - `_assistantMessageContentKey(...)`

### Reconciliation behavior
For assistant messages **without tool parts**:
- if incoming ID already exists on server: unchanged
- else, if content (`sanitized.parts`) matches a server assistant message in sequence: replace incoming ID with server ID
- else: unchanged

This preserves current tool-flow logic and prevents duplicate inserts for plain assistant text round-trips.

## Why this approach

- Keeps server as source of truth for persisted assistant IDs.
- Fixes duplication without requiring client transport protocol changes.
- Avoids touching tool-call merge paths that already have dedicated logic.

### Suggested review order
1. **Test first**
   - `packages/ai-chat/src/tests/merge-server-state.test.ts`
   - Verify the new test reproduces #957 semantics.
2. **Fix implementation**
   - `packages/ai-chat/src/index.ts`
   - Start at `_mergeIncomingWithServerState()` and then inspect new helper methods.
3. **Regression considerations**
   - Confirm tool-call paths remain handled by existing `_resolveMessageForToolMerge` behavior.

### What to validate
- No duplicate assistant rows when same assistant content is round-tripped with different IDs.
- Existing tool merge tests still pass.
- Existing non-tool persistence behavior remains stable when IDs already match.

### Commands used locally
```bash
cd packages/ai-chat
npm run test:workers -- merge-server-state.test.ts
npm run test:workers -- merge-server-state.test.ts client-tool-duplicate-message.test.ts
```

### Tests

- [x]  check e2e scenarios in all existing examples


packages/ai-chat/src/index.ts
```diff
@@ -998,38 +998,125 @@ export class AIChatAgent<
...
-    if (serverToolOutputs.size === 0) {
+    // Merge server's tool outputs into incoming messages.
+    const withMergedToolOutputs =
+      serverToolOutputs.size === 0
+        ? incomingMessages
+        : incomingMessages.map((msg) => {
+            if (msg.role !== "assistant") return msg;
+
+            let hasChanges = false;
+            const updatedParts = msg.parts.map((part) => {
```
threepointone, line 1009: this is n ⨉ m, but it's fine tbh. we'll revisit in the future if needed.

threepointone:
Reviewed and tested locally — all 855 tests pass (654 agents + 201 ai-chat), including the 2 new tests added in this PR.

The fix is correct, well-scoped, and backward compatible. For existing apps where client and server IDs already match, the fast path handles everything with no behavioral change. For apps hitting #957, the reconciliation prevents future duplicates without disrupting existing data.

One thing to note: this fix prevents *new* duplicates but doesn't clean up *existing* duplicate rows that may have accumulated from the bug. During the reconciliation scan we already identify the exact moment a client ID maps to a server ID — we could piggyback on that to delete the orphaned row from the DB. There's an edge case where the client resends the duplicate's own ID (exact-match fast path fires, content matching never runs), which would need a separate dedup pass.

Not worth prioritizing now — existing duplicates are inert and will rotate out via `maxPersistedMessages` naturally. If we hear reports of bloated DBs we can add the cleanup in a follow-up PR.

threepointone (approved):
strong, let's ship it

PR #962, opened 2026-02-21
fix(codemode): validate tool arguments against Zod schema before execution

tumberger:
## Summary

- `createCodeTool()` extracts `.execute` functions from tools but **discards their Zod schemas**, passing them unwrapped to `ToolDispatcher`
- `ToolDispatcher.call()` does `JSON.parse(argsJson) → fn(args)` with **no schema validation**, so sandboxed code can invoke tool functions with arbitrary JSON payloads that bypass the constraints defined by tool authors
- This patch wraps each extracted `execute` function with its corresponding Zod schema (`inputSchema` or `parameters`) so arguments are validated before reaching the tool function — invalid payloads throw a `ZodError` caught by the existing error handler

## Root cause

In `tool.ts:101-111` (before this patch):

```typescript
for (const [name, t] of Object.entries(tools)) {
  const execute = "execute" in t
    ? (t.execute as (args: unknown) => Promise<unknown>)
    : undefined;
  if (execute) {
    fns[sanitizeToolName(name)] = execute; // ← schema discarded
  }
}
```

The Zod schemas (`t.inputSchema` / `t.parameters`) are available but never extracted. When the sandbox calls a tool via `ToolDispatcher.call()`, the raw `JSON.parse`'d payload reaches the tool function directly.

## Fix

```typescript
const schema: ZodType | undefined =
  "inputSchema" in t ? (t.inputSchema as ZodType)
  : "parameters" in t ? ((t as Record<string, unknown>).parameters as ZodType)
  : undefined;

fns[sanitizeToolName(name)] = schema
  ? async (args: unknown) => execute(schema.parse(args))
  : execute;
```

### DIFF
--- a/packages/codemode/src/tool.ts
+++ b/packages/codemode/src/tool.ts
@@ -1,5 +1,5 @@
-import { tool, type Tool } from "ai";
-import { z, type ZodType } from "zod";
+import { tool, type Tool, asSchema } from "ai";
+import { z } from "zod";
 import type { ToolSet } from "ai";
 import * as acorn from "acorn";
 import { generateTypes, sanitizeToolName, type ToolDescriptors } from "./types";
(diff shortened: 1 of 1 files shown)


mattzcarey:
thanks for this. will look at getting it merged

PR #991, opened 2026-02-25
feat(experimental): add session memory with compaction

mattzcarey:
## Summary

Adds an experimental **Session Memory** API to the Agents SDK — persistent conversation history stored in Durable Object SQLite with automatic compaction to keep context windows manageable.

### Why

Agents need to remember conversations across requests. Messages grow unbounded, tool outputs bloat context, and eventually you hit model limits. Session Memory solves this with two layers of compaction:

1. **microCompaction** (runs on every `append()`, no LLM) — truncates oversized tool outputs and long text parts in older messages
2. **Full compaction** (user-supplied function, typically LLM summarization) — triggers automatically when estimated tokens exceed a threshold

### Architecture

```
Session (user-facing API — append, getMessages, compact)
  ├─ compaction orchestration
  ├─ utils/compaction.ts — microCompact() pure function
  ├─ utils/tokens.ts — heuristic token estimation
  └─ SessionProvider (interface — pure storage CRUD)
       └─ AgentSessionProvider (DO SQLite)
       └─ [future providers]
```

The key design: **compaction lives in `Session`, not in the provider**. Providers are pure storage. Any future backend gets compaction for free by wrapping with `Session`.

### Usage

```typescript
import { Agent, callable } from "agents";
import { Session, AgentSessionProvider } from "agents/experimental/memory/session";
import type { UIMessage } from "ai";

export class ChatAgent extends Agent<Env> {
  // microCompaction enabled by default — truncates tool outputs in older messages
  session = new Session(new AgentSessionProvider(this), {
    compaction: {
      tokenThreshold: 10000,
      fn: async (messages) => {
        // Your LLM summarization logic here
        const summary = await summarize(messages);
        return [{ id: crypto.randomUUID(), role: "assistant", parts: [{ type: "text", text: summary }] }];
      }
    }
  });

  @callable()
  async chat(message: string): Promise<string> {
    await this.session.append({
      id: `user-${crypto.randomUUID()}`,
      role: "user",
      parts: [{ type: "text", text: message }]
    });

    const response = await generateResponse(this.session.getMessages());

    await this.session.append({
      id: `asst-${crypto.randomUUID()}`,
      role: "assistant",
      parts: [{ type: "text", text: response }]
    });

    return response;
  }

  @callable()
  getMessages(): UIMessage[] {
    return this.session.getMessages();
  }
}
```

### Session API

| Method | Description |
|---|---|
| `append(messages)` | Insert messages, run microCompaction on older messages, trigger full compaction if threshold exceeded |
| `getMessages(options?)` | Query messages with optional filtering (limit, offset, role, before/after) |
| `getMessage(id)` | Get single message by ID |
| `getLastMessages(n)` | Get N most recent messages |
| `updateMessage(message)` | Update an existing message |
| `deleteMessages(ids)` | Delete messages by ID |
| `clearMessages()` | Clear all messages |
| `compact()` | Manually trigger full compaction |

### SessionProvider Interface

| Method | Description |
|---|---|
| `getMessages(options?)` | Read messages with optional filtering |
| `getMessage(id)` | Read single message |
| `getLastMessages(n)` | Read N most recent |
| `appendMessages(messages)` | Insert one or more messages |
| `updateMessage(message)` | Update existing message |
| `deleteMessages(ids)` | Delete by IDs |
| `clearMessages()` | Delete all |
| `getOlderMessages(keepRecent)` | Read all messages except the most recent N (used by microCompaction) |
| `replaceMessages(messages)` | Atomic bulk replace (used by compact — transactional with rollback) |

### Implementing a Custom Provider

To add a new storage backend (KV, D1, external DB), implement `SessionProvider`:

```typescript
import type { SessionProvider } from "agents/experimental/memory/session";

class KVSessionProvider implements SessionProvider {
  // Implement all methods from the interface above.
  // Key points:
  // - getOlderMessages(keepRecent) must return all messages EXCEPT the N most recent
  // - replaceMessages(messages) should be atomic (transaction or equivalent)
  // - Preserve created_at timestamps in replaceMessages when possible
  // - Read methods are currently synchronous — async provider support is planned

  // Then wrap with Session to get compaction for free:
  // const session = new Session(new KVSessionProvider(...), { compaction: { ... } });
}
```

> **Note:** Read methods on `SessionProvider` are currently synchronous. This works for DO SQLite but would need to become async for KV/D1/external DB backends. This is a known limitation that will be addressed before the API stabilizes.

### Compaction Flow

**On `append()`:**
1. Insert messages into storage
2. Check token threshold → if exceeded, run full compaction and return
3. If full compaction fails, fall back to microCompaction
4. Otherwise, run microCompaction on older messages (beyond `keepRecent` window)

**On `compact()` (manual or auto):**
1. Read all messages from storage
2. Run user-supplied compaction function
3. Atomically replace all messages via `replaceMessages()` (transactional)

**microCompaction defaults:**
- Truncate tool outputs > 30,000 chars
- Truncate text parts > 10,000 chars
- Keep 4 most recent messages intact
- Configurable via `{ microCompaction: { truncateToolOutputs: 2000, keepRecent: 10 } }`

### Files

**Core (`packages/agents/src/experimental/memory/`):**
- `session/session.ts` — Session class (compaction orchestrator)
- `session/provider.ts` — SessionProvider interface
- `session/providers/agent.ts` — AgentSessionProvider (DO SQLite)
- `session/types.ts` — all types
- `session/index.ts` — barrel exports
- `utils/compaction.ts` — microCompact internals
- `utils/tokens.ts` — heuristic token estimation
- `utils/index.ts` — utils barrel
- `index.ts` — memory barrel

**Example (`examples/experimental/session-memory/`):**
- Full working chat app with Kumo UI
- Deployed at https://agents-session-memory-example.mattzcarey.workers.dev

**Tests (`packages/agents/src/tests/`):**
- `experimental/memory/session/provider.test.ts` — 24 tests
- `agents/session.ts` — 3 test agent variants (default, no-micro, custom rules)

### Reviewer Notes

- **TODO: enforce 2MB row limit in Agent SQLite.** Individual message rows could theoretically exceed the DO SQLite row size limit. We should validate message size on `appendMessages()` and `updateMessage()` and throw before hitting the storage limit. Not blocking for experimental, but must be addressed before stabilizing.

## Test plan

- [x] All 730 tests pass (44 test files), including 24 new session provider tests
- [x] `oxfmt --check .` passes
- [x] `oxlint` passes — 0 warnings, 0 errors
- [x] Example builds and deploys
- [x] Message persistence verified (survives page reload)
- [x] Compaction works end-to-end
- [x] `replaceMessages()` wrapped in transaction (rollback on failure)
- [x] Config validation rejects invalid microCompaction rules
- [x] Reviewed by GPT-5.3-Codex — approved after 2 rounds

### DIFF
--- a/examples/experimental/session-memory/src/server.ts
+++ b/examples/experimental/session-memory/src/server.ts
@@ -28,7 +28,7 @@ async function compactMessages(messages: UIMessage[]): Promise<UIMessage[]> {
 
   return [
     {
-      id: `summary-${Date.now()}`,
+      id: `summary-${crypto.randomUUID()}`,
       role: "assistant",
       parts: [{ type: "text", text: `[Conversation Summary]\n${text}` }]
     }
@@ -43,7 +43,7 @@ export class ChatAgent extends Agent<Env> {
   @callable()
   async chat(message: string): Promise<string> {
     await this.session.append({
-      id: `user-${Date.now()}`,
+      id: `user-${crypto.randomUUID()}`,
       role: "user",
       parts: [{ type: "text", text: message }]
     });
@@ -56,7 +56,7 @@ export class ChatAgent extends Agent<Env> {
     });
 
     await this.session.append({
-      id: `assistant-${Date.now()}`,
+      id: `assistant-${crypto.randomUUID()}`,
       role: "assistant",
       parts: [{ type: "text", text }]
     });
--- a/packages/agents/src/experimental/memory/session/session.ts
+++ b/packages/agents/src/experimental/memory/session/session.ts
@@ -55,8 +55,9 @@ export class Session {
 
     // 2. Full compaction if token threshold exceeded — runs instead of microCompaction
     if (this.shouldAutoCompact()) {
-      await this.compact();
-      return;
+      const result = await this.compact();
+      if (result.success) return;
+      // Fall through to microCompaction if full compaction failed
     }
 
     // 3. MicroCompaction on older messages (only if no full compaction)
--- a/examples/experimental/session-memory/src/client.tsx
+++ b/examples/experimental/session-memory/src/client.tsx
@@ -70,7 +70,7 @@ function Chat() {
     setIsLoading(true);
 
     const userMsg: UIMessage = {
-      id: `user-${Date.now()}`,
+      id: `user-${crypto.randomUUID()}`,
       role: "user",
       parts: [{ type: "text", text }]
     };
@@ -79,7 +79,7 @@ function Chat() {
     try {
       const response = await agent.call<string>("chat", [text]);
       const assistantMsg: UIMessage = {
-        id: `assistant-${Date.now()}`,
+        id: `assistant-${crypto.randomUUID()}`,
         role: "assistant",
         parts: [{ type: "text", text: response }]
       };
--- a/packages/agents/src/experimental/memory/session/providers/agent.ts
+++ b/packages/agents/src/experimental/memory/session/providers/agent.ts
@@ -224,18 +224,26 @@ export class AgentSessionProvider implements SessionProvider {
       timestampMap.set(row.id, row.created_at);
     }
 
-    // Clear and re-insert
-    this.agent.sql`DELETE FROM cf_agents_session_messages`;
+    // Atomic replace: transaction ensures no data loss on partial failure
+    this.agent.sql`BEGIN TRANSACTION`;
+    try {
+      this.agent.sql`DELETE FROM cf_agents_session_messages`;
 
-    const now = new Date().toISOString();
-    for (const message of messages) {
-      const json = JSON.stringify(message);
-      const created_at = timestampMap.get(message.id) ?? now;
-      this.agent.sql`
-        INSERT INTO cf_agents_session_messages (id, message, created_at)
-        VALUES (${message.id}, ${json}, ${created_at})
-        ON CONFLICT(id) DO UPDATE SET message = excluded.message
-      `;
+      const now = new Date().toISOString();
+      for (const message of messages) {
+        const json = JSON.stringify(message);
+        const created_at = timestampMap.get(message.id) ?? now;
+        this.agent.sql`
+          INSERT INTO cf_agents_session_messages (id, message, created_at)
+          VALUES (${message.id}, ${json}, ${created_at})
+          ON CONFLICT(id) DO UPDATE SET message = excluded.message
+        `;
+      }
+
+      this.agent.sql`COMMIT`;
+    } catch (err) {
+      this.agent.sql`ROLLBACK`;
+      throw err;
     }
   }
 
--- a/packages/agents/src/experimental/memory/session/types.ts
+++ b/packages/agents/src/experimental/memory/session/types.ts
@@ -66,7 +66,7 @@ export interface CompactionConfig {
 
   /**
    * Function to compact messages.
-   * Receives current messages (after microCompaction if enabled), returns new messages.
+   * Receives current messages as stored, returns new messages.
    */
   fn: CompactFunction;
 }
@@ -90,7 +90,7 @@ export interface SessionProviderOptions {
    * Truncates tool outputs and long text in older messages.
    *
    * Runs automatically on every `append()` — older messages (beyond `keepRecent`)
-   * are truncated in storage. Also runs during `compact()` before the custom fn.
+   * are truncated in storage.
    * This is a destructive operation: original content is permanently replaced.
    * `getMessages()` returns stored content as-is (already compacted).
    *
@@ -104,7 +104,6 @@ export interface SessionProviderOptions {
 
   /**
    * Full compaction with custom function (typically LLM summarization).
-   * Runs after microCompact if both are configured.
    */
   compaction?: CompactionConfig;
 }
--- a/packages/agents/src/experimental/memory/utils/compaction.ts
+++ b/packages/agents/src/experimental/memory/utils/compaction.ts
@@ -39,23 +39,34 @@ export function parseMicroCompactionRules(
     };
   }
 
-  // Custom rules object
-  return {
-    truncateToolOutputs:
-      config.truncateToolOutputs === false
-        ? false
-        : config.truncateToolOutputs === true ||
-            config.truncateToolOutputs === undefined
-          ? DEFAULTS.truncateToolOutputs
-          : config.truncateToolOutputs,
-    truncateText:
-      config.truncateText === false
-        ? false
-        : config.truncateText === true || config.truncateText === undefined
-          ? DEFAULTS.truncateText
-          : config.truncateText,
-    keepRecent: config.keepRecent ?? DEFAULTS.keepRecent
-  };
+  // Custom rules object — validate numeric values
+  const keepRecent = config.keepRecent ?? DEFAULTS.keepRecent;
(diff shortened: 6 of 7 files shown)


threepointone:
Great work on this, @mattzcarey! Session memory with compaction is a real need for long-running agent conversations, and the two-tier approach (cheap truncation vs. LLM summarization) is pragmatic. Shipping under `experimental/` is the right call. Here's a detailed review:

---

## 🔴 Must-fix before merge

### 1. `compact()` is not atomic — data loss risk

The `clear()` + re-insert loop in `compact()` isn't wrapped in a transaction. If the DO is evicted or the loop fails partway through, messages are lost. Consider wrapping in `BEGIN`/`COMMIT`, or replacing with a delete-and-replace strategy that doesn't clear everything first.

### 2. `getMessages()` query branch explosion

~150 lines of near-duplicate SQL for every permutation of `role`/`before`/`after`/`limit`/`offset`. This is fragile — any schema change requires touching 12+ code paths. Consider building WHERE clauses dynamically instead.

### 3. `microCompaction` timing semantics are confusing

`applyMicroCompaction` only runs inside `compact()`, not on `getMessages()`. Since microCompaction defaults to `true`, users will expect their token usage is being automatically managed on reads. Either:
- Apply it as a read-time transformation in `getMessages()`, or
- Make it very clear this is a destructive write-time-only operation (and reconsider defaulting to `true`)

---

## 🟡 Should-fix

### 4. Token estimation should be documented or renamed

`CHARS_PER_TOKEN = 4` is the sole driver of `tokenThreshold`. The config name implies token-level precision. Either document that it's character-based under the hood, allow users to supply a custom counting function, or rename to `charThreshold`.

### 5. `append()` double-reads all messages

Every `append()` with a `tokenThreshold` does a full table scan + JSON parse to check the threshold. Then `compact()` does another full read. Consider tracking approximate size in-memory or in a metadata row.

### 6. `compact()` loses `created_at` timestamps

Re-inserted messages get new `CURRENT_TIMESTAMP` values, breaking `before`/`after` query semantics. Preserve original timestamps during compaction.

### 7. `delete()` and `append()` use per-row loops

One SQL statement per ID/message. For bulk operations this is unnecessarily slow. Not critical for typical usage but worth batching.

---

## 🟢 Minor / follow-ups

- **No changeset** — this adds a new public export (`agents/experimental/memory/session`), needs one per repo rules
- **Tests don't cover `before`/`after` date filtering** — the 150-line query branch for dates has zero coverage
- **`SessionProvider` interface mixes sync/async** — `getMessages()` is sync, which prevents async backends from implementing the interface. Consider making all methods async if this is meant to be pluggable
- **Example uses `onRequest` + raw `fetch()`** — per example conventions, prefer `@callable` + `useAgent`/`agent.call()`
- **Example missing `ThemeProvider` wrapper and `public/favicon.ico`** — per example AGENTS.md conventions

---

## Design question

How does this relate to the existing message storage in `@cloudflare/ai-chat` / `AIChatAgent`? Worth clarifying whether this is meant to replace, complement, or eventually be adopted by the chat agent.

---

Overall this is a solid foundation — the core concept and API shape are good. The main blockers are the non-atomic compact, the query code duplication, and clarifying microCompaction semantics. Fix those three and this is ready to ship as experimental. Nice work! 🚀

threepointone:
thanks windsurf

mattzcarey:
niceeee. so api good but provider sucks. I can work with this

examples/experimental/session-memory/src/server.ts
```diff
@@ -0,0 +1,130 @@
...
+        role: m.role as "user" | "assistant" | "system",
+        content: m.parts
+          .filter((p): p is { type: "text"; text: string } => p.type === "text")
+          .map((p) => p.text)
+          .join("\n")
+      }));
+
+      const workersai = createWorkersAI({ binding: this.env.AI });
+      const { text } = await generateText({
+        model: workersai("@cf/meta/llama-3.3-70b-instruct-fp8-fast"),
```
mattzcarey: use a better model like glm. look in the rest of the repo for example.

examples/experimental/session-memory/src/server.ts
```diff
@@ -0,0 +1,130 @@
...
+      };
+      await this.session.append(assistantMessage);
+
+      return Response.json({
+        response: text,
+        messageCount: this.session.count()
+      });
+    }
+
+    if (request.method === "POST" && url.pathname.endsWith("/compact")) {
```
mattzcarey: this should be a callable. not a fetch endpoint.

examples/experimental/session-memory/src/server.ts
```diff
@@ -0,0 +1,130 @@
...
+    compaction: {
+      tokenThreshold: 10000,
+      fn: compactMessages
+    }
+  });
+
+  async onRequest(request: Request): Promise<Response> {
+    const url = new URL(request.url);
+
+    if (request.method === "GET" && url.pathname.endsWith("/messages")) {
```
mattzcarey: this should be a callable too

packages/agents/src/experimental/memory/session/session.ts
```diff
@@ -0,0 +1,143 @@
...
+				for (let i = 0; i < older.length; i++) {
+					if (compacted[i] !== older[i]) {
+						this.storage.update(compacted[i]);
+					}
+				}
+			}
+		}
+
+		// 3. Full compaction if token threshold exceeded (fast heuristic check first)
+		if (this.shouldAutoCompactFast()) {
```
mattzcarey: we should do compaction if it is going to happen before micro compaction. and we should return. 

micro compaction happens if we dont need compaction.

packages/agents/src/experimental/memory/session/session.ts
```diff
@@ -0,0 +1,143 @@
...
+
+			// Run microCompaction first (if enabled) — skip recent messages
+			if (this.microCompactionRules) {
+				const rules = this.microCompactionRules;
+				result = result.map((msg, i) => {
+					const isRecent = i >= result.length - rules.keepRecent;
+					if (isRecent) return msg;
+					return microCompact([msg], rules)[0];
+				});
+			}
```
mattzcarey: remove microcompaction here.

packages/agents/src/experimental/memory/session/session.ts
```diff
@@ -0,0 +1,143 @@
...
+				error: err instanceof Error ? err.message : String(err),
+			};
+		}
+	}
+
+	/**
+	 * Fast pre-check for auto-compaction using character-count heuristic.
+	 * Avoids parsing all messages when token threshold is clearly not met.
+	 */
+	private shouldAutoCompactFast(): boolean {
```
mattzcarey: rename to `shouldAutoCompact`

packages/agents/src/experimental/memory/session/provider.ts
```diff
@@ -0,0 +1,70 @@
...
+import type { UIMessage } from "ai";
+import type { MessageQueryOptions } from "./types";
+
+/**
+ * Session storage provider interface.
+ *
+ * Implement this interface to create custom session storage backends.
+ * Providers handle CRUD only — compaction is orchestrated by the Session wrapper.
+ */
+export interface SessionProvider {
```
mattzcarey, line 17: namings of the methods feel weird. some do not contain message/messages where as some do. I would prefer to have it for all. 

also there are some we dont actually need. 

do we use update?
do we need count?

PR #993, opened 2026-02-26
fix(ai-chat): preserve server tool outputs for approval-responded state

ferdousbhai:
## Summary

`_mergeIncomingWithServerState` only preserved server-side tool outputs when the incoming client part was in `input-available` state. This missed `approval-responded` — when a `needsApproval` tool is executed server-side (in `onChatMessage` after client approval), the server sets the part to `output-available`, but the client still holds `approval-responded` since it never received that state update.

On the next `persistMessages` call, the client's stale `approval-responded` overwrites the server's `output-available` + output, making the tool appear unexecuted. This causes duplicate tool execution if the server re-runs approved tools from message history.

### Repro scenario

1. Model calls a tool with `needsApproval: true` (e.g. `create_doc`)
2. Client shows approval UI → user approves → server executes tool, sets `output-available`
3. Model continues → calls another `needsApproval` tool (e.g. `edit_doc`)
4. User approves second tool → client sends full message history via `sendAutomaticallyWhen`
5. Client's message history still has the first tool in `approval-responded` (stale)
6. `persistMessages` → `_mergeIncomingWithServerState` checks `part.state === "input-available"` → **doesn't match** `approval-responded`
7. Server's `output-available` is overwritten → first tool appears unexecuted → **gets re-run**

### Fix

Expand the merge condition to also restore `output-available` when the incoming part is in `approval-responded` state.

### Test plan

- [x] Added test: preserves server-side tool outputs when client sends approval-responded state
- [x] All 230 existing tests pass

### DIFF
--- a/.changeset/fix-hibernated-stream-state.md
+++ b/.changeset/fix-hibernated-stream-state.md
@@ -1,0 +1,9 @@
+---
+"@cloudflare/ai-chat": patch
+---
+
+Fix active streams losing UI state after reconnect and dead streams after DO hibernation.
+
+- Send `replayComplete` signal after replaying stored chunks for live streams, so the client flushes accumulated parts to React state immediately instead of waiting for the next live chunk.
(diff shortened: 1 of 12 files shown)


threepointone (approved):
lgtm, thank you!

PR #1085, opened 2026-03-07
fix(mcp): remove unnecessary storage operations in McpAgent

threepointone:
## Changes

- **Skip redundant props storage read in `onStart`**: When props are passed directly, `updateProps()` already writes to storage and sets `this.props`. The subsequent `storage.get("props")` was unnecessary. Now only reads from storage on the hibernation recovery path (no props passed).

- **Replace elicitation storage polling with in-memory Promise/resolver**: The old pattern used 6+ storage operations per elicitation (put request, poll get response every 100ms, delete request, delete response, plus get/put in the handler). Replaced with zero-storage in-memory signaling via a `_pendingElicitations` Map. The DO can't hibernate while a tool call is actively running, so in-memory state is safe here.

## Tests

- Added `elicitNameCustom` test tool that exercises `McpAgent.elicitInput()` directly
- 4 new integration tests in `elicitation.test.ts`: accept/cancel/error via streamable HTTP, accept via SSE
- All 330 existing MCP tests pass with no regressions

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1085" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/.changeset/schema-version-gating.md
+++ b/.changeset/schema-version-gating.md
@@ -1,0 +1,11 @@
+---
+"agents": patch
+---
+
+Simplify Agent storage: schema version gating and single-row state
+
+- Skip redundant DDL migrations on established DOs by tracking schema version in `cf_agents_state`
+- Eliminate `STATE_WAS_CHANGED` row — state persistence now uses a single row with row-existence check, correctly handling falsy values (null, 0, false, "")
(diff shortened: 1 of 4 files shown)


threepointone:
good catches!

PR #1086, opened 2026-03-07
feat(agents): schema version gating and single-row state optimization

threepointone:
## What

Two backward-compatible storage optimizations for the base `Agent` class:

### 1. Schema version gating

Every time a Durable Object wakes from hibernation, the constructor runs ~12 DDL statements (`CREATE TABLE IF NOT EXISTS`, `ALTER TABLE ADD COLUMN`, `SELECT sql FROM sqlite_master`, etc.). These are all no-ops on established DOs — the tables already exist with the right schema.

Now the constructor tracks a schema version in `cf_agents_state` (a `cf_schema_version` row). On wake, it reads this single row. If the version is current, **all DDL is skipped**. Fresh DOs and DOs from older SDK versions still run the full migration path and get stamped with the current version.

**Benefit**: Eliminates ~12 redundant SQL statements on every DO wake. Only cost on the hot path is one `CREATE TABLE IF NOT EXISTS` (for the state table itself) + one `SELECT` (to read the version row).

### 2. Single-row state optimization

Previously, `setState()` wrote two rows to `cf_agents_state`:
- `cf_state_row_id` — the actual state JSON
- `cf_state_was_changed` — a boolean flag

The getter also ran two queries — one for the flag, one for the state.

Now `setState()` writes only the state row. The getter checks **row existence** (`result.length > 0`) instead of a truthiness flag. This correctly handles falsy state values (`null`, `0`, `false`, `""`) that the old truthiness check would have missed.

Legacy `cf_state_was_changed` rows are cleaned up during schema migration.

**Benefit**: One fewer SQL write per `setState()` call, one fewer SQL read per cold `state` access, and correct handling of falsy state values.

### 3. Schema DDL snapshot test

A test that compares the actual `sqlite_master` DDL for all `cf_agents_*` tables against a hardcoded snapshot. If anyone changes a table definition without bumping `CURRENT_SCHEMA_VERSION`, the test fails with a clear message explaining what to do.

**Benefit**: Makes it impossible to forget to bump the schema version when changing table definitions.

### 4. Bug fix: corrupted state test helper

The existing `insertCorruptedState()` test helper was using wrong row IDs (`'STATE'` instead of `'cf_state_row_id'`), so the corruption recovery test was passing for the wrong reason. Fixed to use correct IDs and proper cache reset.

## Backward compatibility

Both changes are fully backward-compatible:
- Existing DOs with `user_version = 0` run the full migration path on next wake
- Legacy `STATE_WAS_CHANGED` rows are cleaned up during migration
- The state getter works with both old (two-row) and new (single-row) formats

## Tests

26 new tests covering schema gating, single-row state, falsy values, corruption recovery, legacy compatibility, and the DDL snapshot. All 350 existing tests pass with zero regressions.

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1086" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/packages/agents/src/index.ts
+++ b/packages/agents/src/index.ts
@@ -393,10 +393,12 @@ const KEEP_ALIVE_INTERVAL_MS = 30_000;
 /**
  * Schema version for the Agent's internal SQLite tables.
  * Bump this when adding new tables, columns, or migrations.
- * The constructor uses PRAGMA user_version to skip DDL on established DOs.
+ * The constructor stores this as a row in cf_agents_state and checks it
+ * on wake to skip DDL on established DOs.
  */
 const CURRENT_SCHEMA_VERSION = 1;
 
+const SCHEMA_VERSION_ROW_ID = "cf_schema_version";
 const STATE_ROW_ID = "cf_state_row_id";
 // Legacy key — no longer written, but read for backward compatibility with
 // DOs that were created before the single-row state optimization.
@@ -842,15 +844,16 @@ export class Agent<
       throw new SqlError(query, e);
     }
   }
-  constructor(ctx: AgentContext, env: Env) {
-    super(ctx, env);
-
-    if (!wrappedClasses.has(this.constructor)) {
-      // Auto-wrap custom methods with agent context
-      this._autoWrapCustomMethods();
-      wrappedClasses.add(this.constructor);
-    }
-
+  /**
+   * Create all internal tables and run migrations if needed.
+   * Called by the constructor on every wake. Idempotent — skips DDL when
+   * the stored schema version matches CURRENT_SCHEMA_VERSION.
+   *
+   * Protected so that test agents can re-run the real migration path
+   * after manipulating DB state (since ctx.abort() is unavailable in
+   * local dev and the constructor only runs once per DO instance).
+   */
+  protected _ensureSchema(): void {
     // Schema version gating: skip all DDL on established DOs whose schema
     // is already up-to-date. We always create cf_agents_state first (cheap
     // idempotent DDL) and store the version as a row inside it.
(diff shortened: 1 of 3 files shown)


packages/agents/src/index.ts
```diff
@@ -335,7 +335,16 @@ export type AddRpcMcpServerOptions = {
 
 const KEEP_ALIVE_INTERVAL_MS = 30_000;
 
+/**
+ * Schema version for the Agent's internal SQLite tables.
+ * Bump this when adding new tables, columns, or migrations.
+ * The constructor uses PRAGMA user_version to skip DDL on established DOs.
```
deathbyknowledge: The code is not actually using `PRAGMA user_version` but a custom row in the `cf_agents_state`. Also probably worth putting the `cf_schema_version` row id in a constant like we do for the rest.

packages/agents/src/tests/agents/state.ts
```diff
@@ -74,24 +79,117 @@ export class TestStateAgent extends Agent<Record<string, unknown>, TestState> {
...
+  resetSchemaVersion() {
+    this.ctx.storage.sql.exec(
+      "DELETE FROM cf_agents_state WHERE id = 'cf_schema_version'"
+    );
+  }
+
+  // Re-run the migration logic that normally runs in the constructor.
+  // Useful for testing migration behavior since getAgentByName returns
+  // the same DO instance (constructor won't re-run).
+  runSchemaMigration() {
```
deathbyknowledge, line 165: This is not really testing anything from the Agent class, instead it updates the `cf_schema_version` manually. If the issue is not being able to re-run the constructor, maybe we could use https://developers.cloudflare.com/durable-objects/api/state/#abort to evict and cause a re-run?

packages/agents/src/tests/schema-and-state-optimization.test.ts
```diff
@@ -0,0 +1,533 @@
+/**
+ * Tests for two storage optimizations:
+ *
+ * 1. Schema version gating (PRAGMA user_version)
```
deathbyknowledge: Same, there's no `PRAGMA user_version`

PR #1116, opened 2026-03-16
fix(ai-chat): avoid duplicate assistant messages during approval continuations

### LINKED ISSUE #1108: useAgentChat: tool approval continuation creates duplicate assistant message in live client state
## Bug Description

After a tool approval triggers a server continuation (`_reply` with `continuation: true`), the client renders a **duplicate assistant message** containing the same tool parts. The persisted state is correct — a page reload shows a single, properly merged message — but the live client state has two separate assistant messages.

This affects both single-tool and grouped (multi-tool) approval flows.

**Version:** `@cloudflare/ai-chat@0.1.8`

## Steps to Reproduce

1. Configure a server tool with `needsApproval` (returns true)
2. Send a message that triggers the tool
3. Client renders the tool part in `approval-requested` state
4. Call `addToolApprovalResponse({ id, approved: true })`
5. Server processes approval, runs tool, starts continuation via `_reply(continuationId, response, [], { continuation: true })`
6. **Observe:** client shows two assistant messages — the original (with tool parts) and a duplicate from the continuation stream

**Expected:** Single merged assistant message (which is what you see after page reload).

## Root Cause Analysis

The issue is a race condition between `CF_AGENT_MESSAGE_UPDATED`, `CF_AGENT_CHAT_MESSAGES`, and `CF_AGENT_USE_CHAT_RESPONSE` on the client during the tool approval continuation flow.

### Server flow (index.js)

When a tool approval arrives with `autoContinue: true`:

**Step 1** — `_applyToolApproval` updates the tool part and calls `_findAndUpdateToolPart`, which calls `persistMessages`. This broadcasts `CF_AGENT_CHAT_MESSAGES` (with `excludeBroadcastIds = []`) and `CF_AGENT_MESSAGE_UPDATED` to all clients.

**Step 2** — After `waitForStream()`, a new continuation starts with a **new** `continuationId = nanoid()`:

```js
// Line ~601
waitForStream().then(() => {
  const continuationId = nanoid();
  // ...
  await this._reply(continuationId, response, [], {
    continuation: true,
    chatMessageId: continuationId
  });
});
```

**Step 3** — Inside `_reply`, a new message object is created with a fresh ID:

```js
// Line ~1508
const message = {
  id: `assistant_${Date.now()}_${Math.random().toString(36).slice(2, 11)}`,
  role: "assistant",
  parts: []
};
```

**Step 4** — `_streamSSEReply` broadcasts chunks as `CF_AGENT_USE_CHAT_RESPONSE` with `continuation: true`.

**Step 5** — After stream completes, the continuation branch in `_reply` correctly merges parts into the last assistant message for persistence:

```js
// Line ~1555
if (continuation) {
  // Finds last assistant message, merges parts, calls persistMessages
  const mergedMessage = {
    ...lastAssistant,
    parts: [...lastAssistant.parts, ...message.parts]
  };
  await this.persistMessages(updatedMessages, excludeBroadcastIds);
}
```

**Step 6** — `persistMessages` broadcasts `CF_AGENT_CHAT_MESSAGES` with `mergedMessages` (input parameter), NOT `this.messages` (reloaded from DB).

### Client flow (react.tsx)

The `CF_AGENT_USE_CHAT_RESPONSE` handler processes the continuation:

```ts
case MessageType.CF_AGENT_USE_CHAT_RESPONSE: {
  if (localRequestIdsRef.current.has(data.id)) return;
  const isContinuation = data.continuation === true;

  if (!activeStreamRef.current || activeStreamRef.current.id !== data.id) {
    let messageId = nanoid();
    let existingParts: ChatMessage["parts"] = [];

    if (isContinuation) {
      const currentMessages = messagesRef.current;
      for (let i = currentMessages.length - 1; i >= 0; i--) {
        if (currentMessages[i].role === "assistant") {
          messageId = currentMessages[i].id;
          existingParts = [...currentMessages[i].parts];
          break;
        }
      }
    }

    activeStreamRef.current = { id: data.id, messageId, parts: existingParts };
  }
  // ...
  flushActiveStreamToMessages(activeMsg);
}
```

The continuation handler correctly tries to find the last assistant message and merge. **However**, the problem occurs when:

1. `CF_AGENT_MESSAGE_UPDATED` arrives first (from Step 1) and either **appends** a new message if it can't find a matching ID, or updates the existing message with server-generated parts. The fallback append path:

   ```ts
   // Line ~570
   if (idx >= 0) {
     // Update in place
   }
   return [...prevMessages, updatedMessage]; // Append — creates extra message
   ```

2. `CF_AGENT_CHAT_MESSAGES` from Step 1's `persistMessages` may also arrive, replacing the messages array with server IDs that differ from the client's AI SDK-generated IDs.

3. By the time `CF_AGENT_USE_CHAT_RESPONSE` chunks arrive for the continuation, `messagesRef.current` is in a state where the "last assistant message" lookup captures an ID that is about to be replaced by another broadcast.

4. `flushActiveStreamToMessages` does:

   ```ts
   const existingIdx = prevMessages.findIndex((m) => m.id === activeMsg.messageId);
   if (existingIdx >= 0) {
     // Update in place — correct path
   }
   return [...prevMessages, partialMessage]; // Append — creates duplicate!
   ```

   If `activeMsg.messageId` was captured from a message that was subsequently replaced by a `CF_AGENT_CHAT_MESSAGES` broadcast (which uses different server-side IDs), the `findIndex` fails and the message is **appended** as a new entry.

### The timing problem

Three async message types race against each other on a single WebSocket:

| Event | Source | Effect on client state |
|-------|--------|----------------------|
| `CF_AGENT_MESSAGE_UPDATED` | `_findAndUpdateToolPart` in tool approval | May append if ID not found |
| `CF_AGENT_CHAT_MESSAGES` | `persistMessages` after tool state update | Replaces entire messages array |
| `CF_AGENT_USE_CHAT_RESPONSE` | `_streamSSEReply` continuation | Creates/updates active stream message |

When `CF_AGENT_CHAT_MESSAGES` replaces the messages array between when the continuation captures `messageId` from `messagesRef.current` and when `flushActiveStreamToMessages` runs, the ID mismatch causes a duplicate append.

### Grouped approval amplification

For grouped approvals (multiple tools needing approval in one response), calling `addToolApprovalResponse` for each tool sends N WebSocket messages each with `autoContinue: true`. This triggers N serial continuations (each waits for the previous via `_streamCompletionPromise`), multiplying the problem. Each continuation generates a new `continuationId` and a new message object, potentially creating N duplicate assistant messages.

## Suggested Fixes

1. **Reconcile by role in `flushActiveStreamToMessages`**: When the continuation's `messageId` isn't found by `findIndex`, fall back to finding the last assistant message by role (same logic as the initial lookup) instead of appending.

2. **Stabilize IDs across broadcasts**: When `CF_AGENT_CHAT_MESSAGES` replaces messages, if there's an active continuation stream, update `activeStreamRef.current.messageId` to match the new assistant message ID.

3. **Derive continuation ID from original message**: Instead of `nanoid()` for `continuationId`, use the existing assistant message's ID or derive from the original request ID, ensuring the client's `localRequestIdsRef` and `activeStreamRef` stay consistent.

4. **Deduplicate `CF_AGENT_TOOL_APPROVAL` on the server**: When N approvals arrive for tools in the same message, batch them and trigger a single continuation instead of N serial ones.

## Current Workaround

Render-level dedup that skips assistant messages whose tool parts were all already rendered by a prior message:

```tsx
messages.reduce((acc, message) => {
  if (message.role === 'assistant') {
    const toolIds = message.parts
      .filter(p => 'toolCallId' in p)
      .map(p => p.toolCallId);
    const hasNonToolContent = message.parts.length > toolIds.length;
    if (toolIds.length > 0 && !hasNonToolContent && toolIds.every(id => acc.seenToolIds.has(id))) {
      return acc; // Skip duplicate
    }
    for (const id of toolIds) acc.seenToolIds.add(id);
  }
  acc.elements.push(<Message ... />);
  return acc;
}, { seenToolIds: new Set(), elements: [] }).elements
```

For grouped approvals, we send manual WebSocket messages with controlled `autoContinue` (only the last tool gets `true`) to avoid N serial continuations, then call `addToolApprovalResponse` for each tool to update client state. The server ignores the library's duplicate WS messages because tool states have already transitioned out of `approval-requested`.

whoiskatrin:
## Summary
- keep continuation streams pinned to the existing assistant message instead of replacing that target with a new `start` chunk `messageId`
- fall back to the last assistant message when a continuation target id disappears after a `CF_AGENT_CHAT_MESSAGES` broadcast, avoiding duplicate live messages/tool parts
- add regression coverage for both continuation-id overwrite and mid-stream assistant-id replacement

Fixes #1108.

## Testing
- `npm run test:react -- use-agent-chat.test.tsx` in `packages/ai-chat`
- `npm run build` in `packages/ai-chat`
- `npm run check` in repo root *(currently fails on pre-existing missing export files in unrelated packages: `agents`, `@cloudflare/think`, `@cloudflare/voice`, `@cloudflare/worker-bundler`, `@cloudflare/shell`)*

## Reviewer Notes
- the key behavior change is in `packages/ai-chat/src/react.tsx`: continuation streams now preserve the assistant id selected at stream setup
- the new `continuation` flag on `activeStreamRef` only scopes the merge fallback for continuation streams; non-continuation streams still adopt the streamed `start` chunk id as before
- the second regression test intentionally simulates the race from the issue by replacing assistant ids via `CF_AGENT_CHAT_MESSAGES` between continuation chunks

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1116" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/.changeset/issue-1108-continuation-merge.md
+++ b/.changeset/issue-1108-continuation-merge.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/ai-chat": patch
+---
+
+Preserve the existing assistant message during tool approval continuations so live client state does not render duplicate assistant messages or tool parts
--- a/packages/ai-chat/src/react-tests/use-agent-chat.test.tsx
+++ b/packages/ai-chat/src/react-tests/use-agent-chat.test.tsx
@@ -1710,3 +1710,190 @@ describe("useAgentChat stream resumption (issue #896)", () => {
       .toHaveTextContent("replayed-and live!");
   });
 });
+
+describe("useAgentChat tool approval continuations (issue #1108)", () => {
+  function createAgentWithTarget({ name, url }: { name: string; url: string }) {
+    const target = new EventTarget();
+    const agent = createAgent({ name, url });
+    (agent as unknown as Record<string, unknown>).addEventListener =
+      target.addEventListener.bind(target);
+    (agent as unknown as Record<string, unknown>).removeEventListener =
+      target.removeEventListener.bind(target);
+    return { agent, target };
+  }
+
+  function dispatch(target: EventTarget, data: Record<string, unknown>) {
+    target.dispatchEvent(
+      new MessageEvent("message", { data: JSON.stringify(data) })
+    );
+  }
+
+  const initialMessages: UIMessage[] = [
+    {
+      id: "assistant-local",
+      role: "assistant",
+      parts: [
+        {
+          type: "tool-dangerousAction",
+          toolCallId: "tc-approval-1",
+          state: "approval-responded",
+          input: { action: "delete" },
+          approval: { id: "approval-req-1", approved: true }
+        }
+      ]
+    }
+  ];
+
+  function TestComponent({ agent }: { agent: ReturnType<typeof useAgent> }) {
+    const chat = useAgentChat({
+      agent,
+      getInitialMessages: () => Promise.resolve(initialMessages)
+    });
+    const assistantMessages = chat.messages.filter(
+      (message) => message.role === "assistant"
+    );
+    const textPart = assistantMessages
+      .flatMap((message) => message.parts)
+      .find((part) => part.type === "text") as { text?: string } | undefined;
+    const toolPartsCount = assistantMessages.reduce((count, message) => {
+      return (
+        count +
+        message.parts.filter(
+          (part) => "toolCallId" in part && part.toolCallId === "tc-approval-1"
+        ).length
+      );
+    }, 0);
+
+    return (
+      <div>
+        <div data-testid="assistant-count">{assistantMessages.length}</div>
+        <div data-testid="assistant-ids">
+          {assistantMessages.map((message) => message.id).join(",")}
+        </div>
+        <div data-testid="tool-parts-count">{toolPartsCount}</div>
+        <div data-testid="text">{textPart?.text ?? ""}</div>
+      </div>
+    );
+  }
+
+  it("keeps the existing assistant id for continuation start chunks", async () => {
+    const { agent, target } = createAgentWithTarget({
+      name: "continuation-start-id",
+      url: "ws://localhost:3000/agents/chat/continuation-start-id?_pk=abc"
+    });
+
+    const screen = await act(async () => {
+      const screen = render(<TestComponent agent={agent} />, {
+        wrapper: ({ children }) => (
+          <StrictMode>
+            <Suspense fallback="Loading...">{children}</Suspense>
+          </StrictMode>
+        )
+      });
+      await sleep(10);
+      return screen;
+    });
+
+    await expect
+      .element(screen.getByTestId("assistant-count"))
+      .toHaveTextContent("1");
+
+    await act(async () => {
+      dispatch(target, {
+        type: "cf_agent_use_chat_response",
+        id: "req-continuation-start",
+        continuation: true,
+        body: '{"type":"start","messageId":"assistant-stream"}',
+        done: false
+      });
+      await sleep(10);
+    });
+
+    await expect
+      .element(screen.getByTestId("assistant-count"))
+      .toHaveTextContent("1");
+    await expect
+      .element(screen.getByTestId("assistant-ids"))
+      .toHaveTextContent("assistant-local");
+    await expect
+      .element(screen.getByTestId("tool-parts-count"))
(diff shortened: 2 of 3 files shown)


threepointone:
perfect pr

PR #1120, opened 2026-03-17
fix(agents): re-arm idempotent scheduleEvery alarms after restart

### LINKED ISSUE #1119: `scheduleEvery` idempotent path doesn't re-set the alarm after DO restart
## Summary

When a Durable Object restarts (eviction, local dev restart, etc.), calling `scheduleEvery()` with an existing schedule finds the row in `cf_agents_schedules` via the idempotent check and returns early — but doesn't call `_scheduleNextAlarm()`. The alarm that was backing the schedule is gone (lost with the previous DO instance), so the schedule never fires again.

This means any recurring schedule set with `scheduleEvery()` silently stops working after a DO restart.

## Steps to Reproduce

1. Create an Agent that uses `scheduleEvery(30, "someCallback")` in `onStart()`
2. Let it run — the callback fires every 30 seconds as expected
3. Restart the process (kill wrangler dev, or wait for DO eviction in production)
4. Access the DO again — `onStart()` fires, calls `scheduleEvery(30, "someCallback")`
5. The callback never fires again

You can verify the schedule row exists in `cf_agents_schedules` (with a `time` value in the past), but no alarm is set to trigger it.

## Root Cause

In `scheduleEvery()`, the idempotent path returns without calling `_scheduleNextAlarm()`:

After a restart, `ctx.storage.setAlarm()` from the previous instance is gone. The schedule row in SQLite persists, but nothing re-sets the alarm to back it.

## Suggested Fix

Add `_scheduleNextAlarm()` before the idempotent return

This is safe because `_scheduleNextAlarm()` just reads the soonest schedule time and calls `ctx.storage.setAlarm()` — it's idempotent itself.

whoiskatrin:
## Summary
- re-arm the Durable Object alarm when `scheduleEvery()` hits the idempotent existing-row path instead of silently returning with no backing alarm
- teach `_scheduleNextAlarm()` to recover overdue persisted schedules after restart while avoiding tight re-alarm loops for interval jobs that are still running
- add regression coverage for both lost-alarm and overdue-row interval schedules, plus a patch changeset for `agents`

## Testing
- `npm run test:workers -w agents -- src/tests/schedule.test.ts`
- `npm run check` 

## Reviewer Notes
- Root cause: the idempotent `scheduleEvery()` path could return an existing persisted interval row after restart without restoring the backing DO alarm, so the schedule silently stopped firing.
- The fix has two parts in `packages/agents/src/index.ts`: re-arm on the idempotent return path, and make `_scheduleNextAlarm()` recover overdue persisted rows after restart.
- The subtle guard is for interval rows with `running = 1`: we do not schedule them for immediate re-execution unless they are already past `hungScheduleTimeoutSeconds`, which avoids a tight alarm loop while a callback is still running.
- The new methods in `packages/agents/src/tests/agents/schedule.ts` are test-only helpers to inspect and clear alarm state inside Workers tests

Fixes #1119

### DIFF
--- a/packages/ai-chat/CHANGELOG.md
+++ b/packages/ai-chat/CHANGELOG.md
@@ -33,7 +33,6 @@
 - [#1013](https://github.com/cloudflare/agents/pull/1013) [`11aaaff`](https://github.com/cloudflare/agents/commit/11aaaffb89c375eba9bedf97074ced556dcdd0e7) Thanks [@threepointone](https://github.com/threepointone)! - Fix Gemini "missing thought_signature" error when using client-side tools with `addToolOutput`.
 
   The server-side message builder (`applyChunkToParts`) was dropping `providerMetadata` from tool-input stream chunks instead of storing it as `callProviderMetadata` on tool UIMessage parts. When `convertToModelMessages` later read the persisted messages for the continuation call, `callProviderMetadata` was undefined, so Gemini never received its `thought_signature` back and rejected the request.
-
   - Preserve `callProviderMetadata` (mapped from stream `providerMetadata`) on tool parts in `tool-input-start`, `tool-input-available`, and `tool-input-error` handlers — both create and update paths
   - Preserve `providerExecuted` on tool parts (used by `convertToModelMessages` for provider-executed tools like Gemini code execution)
   - Preserve `title` on tool parts (tool display name)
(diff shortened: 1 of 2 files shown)


threepointone:
this looks good to me, thanks @whoiskatrin

PR #1135, opened 2026-03-19
feat: think-cli + think-server — remote coding agent

mattzcarey:
## Summary
- **think-server**: Cloudflare Worker (Durable Object) extending Think with model provider config, workspace tools, code execution in dynamic V8 isolates, git operations via isomorphic-git, and context memory
- **think-cli**: Terminal client using pi-tui that connects via WebSocket — pure rendering client, all execution happens server-side
- Modular tool rendering shared between live streaming and session history restore
- Session naming (`/name`) with fuzzy matching on `/resume`
- Gated fetch for sandbox outbound requests with per-host secret bindings (GitHub token auto-injected, never exposed to LLM)

## Test plan
- [ ] `think` CLI connects to local server, sends messages, streams responses
- [ ] Tool calls render correctly during streaming and on session resume
- [ ] `/name`, `/resume`, `/new`, `/clear`, `/session` commands work
- [ ] Git clone/commit/push operations work in production Workers environment
- [ ] Context memory blocks persist and show staleness warnings

### DIFF
--- a/apps/think-server/src/agent.ts
+++ b/apps/think-server/src/agent.ts
@@ -1,0 +1,204 @@
+/**
+ * ThinkServer — Remote coding agent.
+ *
+ * Extends Think with model provider config (from CLI), workspace tools,
+ * and code execution. All interaction is WebSocket native.
+ *
+ * The CLI sends config via callable `configure()` on connect.
+ * The server saves it in Think's SQLite config and uses it for model calls.
+ */
+
+import { Think } from "@cloudflare/think";
+import { callable } from "agents";
+import type { Connection } from "agents";
+import { createWorkspaceTools } from "@cloudflare/think/tools/workspace";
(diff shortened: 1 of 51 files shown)


apps/think-server/src/agent.ts
```diff
@@ -0,0 +1,204 @@
...
+
+  getSystemPrompt(): string {
+    return this.context.toString();
+  }
+
+  getTools(): ToolSet {
+    const config = this.getConfig();
+    const workspaceTools = createWorkspaceTools(this.workspace);
+    const stateProvider = stateToolsFromBackend(createWorkspaceStateBackend(this.workspace));
+    const gitProvider = gitTools(this.workspace, { token: config?.githubToken });
```
mattzcarey, line 152: Here

PR #1142, opened 2026-03-20
fix(ai-chat): serialize chat turns and saveMessages

### LINKED ISSUE #1110: No public API to detect or wait for active streams — needed for schedule() + saveMessages() safety
## Context

We use `schedule()` to run delayed tasks ([docs](https://developers.cloudflare.com/agents/api-reference/schedule-tasks/)). When a task completes, we surface the result as a message in the chat — similar to how Claude Code injects background task results into the conversation. To do this, we call `saveMessages()` from inside the scheduled callback to persist a synthetic user message and trigger an LLM turn.

## Problem

`saveMessages()` has no concurrency guard. If a user-initiated stream is active when the scheduled callback fires, `_reply()` runs twice concurrently — both loops write to `_streamingMessage` and broadcast interleaved chunks. The reverse is equally unsafe (user sends a message while a `saveMessages()`-initiated stream is running).

There's no public API to detect, wait for, or abort an active stream. Our workaround accesses private internals:

```typescript
// In onChatMessage — serialize all stream starts
if ((this as any)._streamCompletionPromise) {
  await (this as any)._streamCompletionPromise
}

// In onConnect — abort on workflow switch
;(this as any)._destroyAbortControllers()
```

## Ask
Can you consider making public:
1. A way to **wait** for an active stream to finish (like `_streamCompletionPromise`, but public)
2. A way to **detect** if a stream is active (like `_streamingMessage !== null`, but public)
3. A way to **abort** an active stream (like `_destroyAbortControllers()`, but public)

whoiskatrin:
Fixes #1110

## Summary
- serialize `onChatMessage()` + `_reply()` behind a private turn queue so websocket requests, tool auto-continuations, and `saveMessages()` cannot stream concurrently
- add protected `isChatTurnActive()`, `waitForIdle()`, `abortActiveTurn()`, `hasPendingInteraction()`, and `waitForPendingInteractionResolution()` helpers for subclass code that needs to inspect active turns and defer injected work until pending tool interactions resolve
- skip queued continuations and `saveMessages()` calls that were enqueued before a chat clear (epoch guard)
- capture `saveMessages()` context (`_lastClientTools`, `_lastBody`) at enqueue time so a later request cannot overwrite it before execution
- fix `_queueAutoContinuation` to use the caller-provided `clientTools` directly instead of re-reading `this._lastClientTools` at execution time
- enqueue tool-result and tool-approval continuations synchronously (with the apply promise as a prerequisite) instead of from a `.then()` microtask, closing a race where `waitForIdle()` could resolve before the continuation was queued
- prevent unhandled promise rejections from `_applyToolResult` / `_applyToolApproval`
- add regression coverage for queued websocket turns, `saveMessages()`, abort behavior, continuation body isolation, pending-interaction coordination, clear-during-active-turn, and clear-during-queued-saveMessages

## Design Notes
- `AIChatAgent` already assumes a single active resumable stream, so this PR keeps that model and serializes the higher-level chat turn (`onChatMessage()` + `_reply()`) instead of introducing concurrent stream support.
- I intentionally did not expose `_streamCompletionPromise`: it resolved too early for this use case because it did not guarantee final assistant-message persistence.
- Auto-continuations now capture request body context at enqueue time so a later request cannot overwrite continuation state before it runs.
- A `_chatEpoch` counter is incremented on chat clear. Queued turns check the epoch before executing and skip if it changed, preventing stale continuations from writing into a cleared conversation. The counter is in-memory only — it doesn't need to survive hibernation because the queue itself (`_chatTurnQueue`) is in-memory and resets on wake.
- Pending-interaction helpers are additive and opt-in. `saveMessages()` itself still only solves turn serialization; callers that need Claude Code-style queued/scheduled behavior can combine `waitForIdle()` with `waitForPendingInteractionResolution()` before injecting a synthetic message.

## Behavior Changes
- `saveMessages()` now waits for any active turn, runs its own turn exclusively, and resolves only after the reply finishes.
- WebSocket requests, tool-result continuations, and tool-approval continuations now execute in order instead of overlapping.
- Chat clear now skips any queued continuations and `saveMessages()` calls that were enqueued before the clear.
- New protected helpers (`isChatTurnActive()`, `waitForIdle()`, `abortActiveTurn()`) support turn coordination.
- New protected helpers (`hasPendingInteraction()`, `waitForPendingInteractionResolution()`) let subclasses detect unresolved client-tool input/approval state before starting a follow-up turn.

## Non-Goals / Limitations
- This PR does not add support for multiple concurrent active streams.
- `abortActiveTurn()` aborts the active request or stream, but it does not interrupt unrelated pre-stream setup such as `waitForMcpConnections()`.
- Pending-interaction waiting is explicit and opt-in; `saveMessages()` does not block on user interactions by default.

## Review Guide
- `packages/ai-chat/src/index.ts`: queueing model, turn helpers, epoch guard, and pending-interaction helpers
- `packages/ai-chat/src/tests/chat-turn-serialization.test.ts`: websocket/saveMessages/abort ordering, clear-during-turn, tool-continuation coverage
- `packages/ai-chat/src/tests/custom-body-continuation.test.ts`: continuation context isolation
- `packages/ai-chat/src/tests/pending-interaction.test.ts`: pending interaction detection and waiting

## Testing
- `npm install`
- `npm run build`
- `npm run test:workers --workspace @cloudflare/ai-chat -- src/tests/chat-turn-serialization.test.ts src/tests/custom-body-continuation.test.ts src/tests/pending-interaction.test.ts`
- `npm run check`

### DIFF
--- a/.changeset/fix-ai-chat-turn-serialization.md
+++ b/.changeset/fix-ai-chat-turn-serialization.md
@@ -6,4 +6,6 @@ fix(ai-chat): serialize chat turns and expose turn control helpers
 
 - queue `onChatMessage()` + `_reply()` work so user requests, tool continuations, and `saveMessages()` never stream concurrently
 - make `saveMessages()` wait for the queued turn to finish before resolving, and reuse the request id for reply cleanup
+- skip queued continuations and `saveMessages()` calls that were enqueued before a chat clear
+- capture `saveMessages()` context (`_lastClientTools`, `_lastBody`) at enqueue time so a later request cannot overwrite it before execution
 - add protected `isChatTurnActive()`, `waitForIdle()`, `abortActiveTurn()`, `hasPendingInteraction()`, and `waitForPendingInteractionResolution()` helpers for subclass code that needs to coordinate active turns and pending tool interactions
--- a/packages/ai-chat/src/index.ts
+++ b/packages/ai-chat/src/index.ts
@@ -229,6 +229,13 @@ export class AIChatAgent<
   private _activeChatTurnRequestId: string | null = null;
 
   /**
+   * Monotonically increasing counter incremented on chat clear.
+   * Queued turns that were enqueued under an older epoch are skipped,
+   * preventing stale continuations from running after the user clears the chat.
+   */
+  private _chatEpoch = 0;
+
+  /**
    * Set of connection IDs that are pending stream resume.
    * These connections have received CF_AGENT_STREAM_RESUMING but haven't sent ACK yet.
    * They should be excluded from live stream broadcasts until they ACK.
@@ -517,6 +524,7 @@ export class AIChatAgent<
 
         // Handle clear chat
         if (data.type === MessageType.CF_AGENT_CHAT_CLEAR) {
+          this._chatEpoch++;
           this._destroyAbortControllers();
           this.sql`delete from cf_ai_chat_agent_messages`;
           this._resumableStream.clearAll();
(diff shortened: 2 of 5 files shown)


threepointone (approved):
looks good. added some checks for an edge cases and added some tests. landing after ci passes.

PR #1151, opened 2026-03-22
fix(ai-chat): simplify turn coordination API

whoiskatrin:
## Summary

Simplifies the turn coordination API surface from 6 protected helpers to 3, and rewrites `waitForPendingInteractionResolution()` with event-driven semantics (no more polling).

Builds on the turn serialization machinery from #1142. That PR added 6 protected helpers; this PR consolidates them before any release ships.

---

### 1. API surface: 6 protected helpers → 3

**Kept (protected):**

| Helper | Purpose |
|--------|---------|
| `waitUntilStable()` | Waits until fully stable — no active stream, no pending interactions, no queued continuations |
| `resetTurnState()` | Aborts the active turn and invalidates queued continuations; use in clear interceptors |
| `hasPendingInteraction()` | Synchronous check for pending tool input or approval |

**Demoted to private:**

| Helper | Why |
|--------|-----|
| `isChatTurnActive()` | Sync check with no standalone subclass use case — callers always then wait or abort |
| `waitForIdle()` | Strict subset of `waitUntilStable()`, which drains the queue as its first step |
| `abortActiveTurn()` | `resetTurnState()` is correct in every real scenario (clear, workflow switch) |

### 2. `waitForPendingInteractionResolution()` → `waitUntilStable()`

Renamed to match the actual semantics ("waits until the conversation is fully stable").

**Before:** Polled `hasPendingInteraction()` on an interval. Could resolve at the exact moment the SDK queued an auto-continuation, requiring the fragile `waitForIdle() → waitForPendingInteraction() → waitForIdle()` dance.

**After:** Event-driven via a new `_pendingInteractionPromise` field. Drains `_chatTurnQueue`, awaits the in-flight tool apply, then loops to catch any `autoContinue` continuation. One call does everything:

```typescript
const ready = await this.waitUntilStable({ timeout: 30_000 });
if (!ready) return;
await this.saveMessages([...this.messages, syntheticMessage]);
```

### 3. `resetTurnState()` added

Encapsulates the full reset (epoch increment + abort controllers + pending promise cleanup) that the built-in `CF_AGENT_CHAT_CLEAR` handler previously did inline. Subclasses intercepting clear can now call one method instead of knowing three internals.

### 4. Unhandled rejection suppression

The `_pendingInteractionPromise` cleanup chains now attach `.catch(() => {})` so rejected tool-result / approval applies do not leak as unhandled rejections.

### 5. Expanded documentation

The `docs/chat-agents.md` turn coordination section now includes:
- When coordination is needed (and when it is not)
- How the turn lifecycle works (turns vs pending interactions)
- Usage guidance for each helper with code examples
- Hibernation behavior (all coordination state is in-memory; messages persist in SQLite)

---

## Files changed

| File | Change |
|------|--------|
| `packages/ai-chat/src/index.ts` | Demote 3 helpers to private; rename `waitForPendingInteractionResolution` → `waitUntilStable`; add `_pendingInteractionPromise`; rewrite with event-driven semantics; add `resetTurnState()`; suppress unhandled rejections |
| `packages/ai-chat/src/tests/pending-interaction.test.ts` | Rename to `waitUntilStableForTest`; add WebSocket tool-result, tool-approval, auto-continuation, active-turn timeout, and `resetTurnState` coverage |
| `packages/ai-chat/src/tests/worker.ts` | Add `waitUntilStableForTest`, `resetTurnStateForTest`; rewrite demoted wrappers to use `as unknown as` casts for private field access |
| `docs/chat-agents.md` | Rewrite turn coordination section with lifecycle explanation, usage guidance, and examples |
| `packages/ai-chat/README.md` | Trim API table from 6 to 3 helpers |
| `.changeset/...` | Patch changeset for `@cloudflare/ai-chat` |

## Verification

- `npm run check` passes (sherif, export checks, oxfmt, oxlint, typecheck across 65 projects)
- `packages/ai-chat` vitest suite passes (315 tests)

## Reviewer notes

- Recommended review order: `index.ts` (API changes) → tests → docs
- All coordination state is in-memory and intentionally does not survive hibernation — the queue resets on wake, and `hasPendingInteraction()` reads from persisted messages in SQLite
- Public surface is unchanged: all touched methods are `protected` or `private`
- The `as unknown as` pattern in test wrappers for private field access matches the existing `getAbortControllerCount()` pattern in the same file

### DIFF
--- a/packages/ai-chat/src/tests/worker.ts
+++ b/packages/ai-chat/src/tests/worker.ts
@@ -143,13 +143,12 @@ export class TestChatAgent extends AIChatAgent<Env> {
 
   isChatTurnActiveForTest(): boolean {
     return (
-      (this as unknown as { _activeChatTurnRequestId: string | null })
-        ._activeChatTurnRequestId !== null
-    );
+      this as unknown as { isChatTurnActive(): boolean }
+    ).isChatTurnActive();
   }
 
   async waitForIdleForTest(): Promise<void> {
-    await (this as unknown as { _chatTurnQueue: Promise<void> })._chatTurnQueue;
+    await (this as unknown as { waitForIdle(): Promise<void> }).waitForIdle();
   }
 
   getPersistedMessages(): ChatMessage[] {
@@ -476,13 +475,12 @@ export class SlowStreamAgent extends AIChatAgent<Env> {
 
   isChatTurnActiveForTest(): boolean {
     return (
-      (this as unknown as { _activeChatTurnRequestId: string | null })
-        ._activeChatTurnRequestId !== null
-    );
+      this as unknown as { isChatTurnActive(): boolean }
+    ).isChatTurnActive();
   }
 
   async waitForIdleForTest(): Promise<boolean> {
-    await (this as unknown as { _chatTurnQueue: Promise<void> })._chatTurnQueue;
+    await (this as unknown as { waitForIdle(): Promise<void> }).waitForIdle();
     return true;
   }
 
@@ -491,13 +489,9 @@ export class SlowStreamAgent extends AIChatAgent<Env> {
   }
 
   abortActiveTurnForTest(): boolean {
-    const self = this as unknown as {
-      _activeChatTurnRequestId: string | null;
-      _cancelChatRequest: (id: string) => void;
-    };
-    if (!self._activeChatTurnRequestId) return false;
-    self._cancelChatRequest(self._activeChatTurnRequestId);
-    return true;
+    return (
+      this as unknown as { abortActiveTurn(): boolean }
+    ).abortActiveTurn();
   }
 
   resetTurnStateForTest(): void {


threepointone:
hmm we might have to revisit 1142, I'll have a look tomorrow

packages/ai-chat/src/tests/worker.ts
```diff
@@ -474,22 +475,29 @@ export class SlowStreamAgent extends AIChatAgent<Env> {
...
-    timeout?: number;
-  }): Promise<boolean> {
-    return this.waitForPendingInteractionResolution(options);
+  waitUntilStableForTest(options?: { timeout?: number }): Promise<boolean> {
+    return this.waitUntilStable(options);
   }
 
   abortActiveTurnForTest(): boolean {
-    return this.abortActiveTurn();
+    const self = this as unknown as {
```
whoiskatrin: this duplicates the abort logic from our index.ts (_cancelChatRequest). If the abort implementation changes, this won't track it? should we keep abortActiveTurn as private rather than removing it, or exposing it as @internal so test helpers can call it directly without the cast?

threepointone: dope catch, fixing to use the internal implementation

PR #1169, opened 2026-03-23
feat(think): integrate agents Session API

mattzcarey:
## Summary

Replace Think's internal SessionManager with the shared one from agents/experimental/memory/session.

### Changes
- `session/index.ts` — now re-exports from agents + truncation utilities (removed 366-line `storage.ts`)
- `getSystemPrompt()` — now async to support `freezeSystemPrompt()`
- `_sessionId` — `protected` so `onCompact()` overrides can access it
- `onStart()` — uses `SessionManager.create(this).withContext(...)` builder
- Context blocks use shared (non-namespaced) providers so memory persists across session splits
- `maxPersistedMessages` removed — use `onCompact()` for storage bounds
- Test agent updated for new `fork()`, `getBranches()` signatures

### Not breaking for existing Think DOs
Column is `content` (preserved from Think's existing schema). No migration needed.

### Notes for reviewers
- Think's `getContextBlocks()` returns `Omit<ContextConfig, "provider">[]` — builder creates `AgentContextProvider` for writable blocks
- `onCompact()` is a no-op by default (opt-in) — `assembleContext()` calls it when `needsCompaction()` returns true
- Patch changeset included for `@cloudflare/think`

## Stack
1. #1166 Session API core ← `main`
2. #1167 SessionManager + multi-session example ← 1
3. **this PR** ← 2

## Test plan
- Think tests pass (pre-existing `@cloudflare/shell` failures on main unrelated)
- Both `agents` and `think` packages build successfully

### DIFF
--- a/.changeset/session-api-think.md
+++ b/.changeset/session-api-think.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/think": patch
+---
+
+Replace internal SessionManager with shared agents Session API. getSystemPrompt() is now async. Adds chainable builder API for context block configuration.
--- a/examples/assistant/src/server.ts
+++ b/examples/assistant/src/server.ts
@@ -153,7 +153,7 @@ export class ChatSession extends Think<Env, AgentConfig> {
     });
   }
 
-  override getSystemPrompt(): string {
+  override async getSystemPrompt(): Promise<string> {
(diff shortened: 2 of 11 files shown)


mattzcarey:
closing this for the moment

PR #1179, opened 2026-03-24
fix(ai-chat): coalesce rapid auto-continuation turns

### LINKED ISSUE #1171: ai-chat: rapid-fire tool results with autoContinue can queue duplicate continuation turns
## Summary

When two client-side tool calls are resolved nearly simultaneously (e.g., a batch of tools in a single `onToolCall` effect cycle), both send `CF_AGENT_TOOL_RESULT` with `autoContinue: true`. Each message independently sets `_pendingAutoContinuation = true` and queues a separate continuation turn via `_runExclusiveChatTurn`.

## Problem

The client side has a guard (`resumingToolContinuationRef`) that prevents duplicate `resumeStream()` calls, so only one client-side stream is created. But the server queues **two** continuation turns. The first runs normally. The second runs against a message state that already includes the first continuation's output, potentially producing redundant LLM content.

Additionally, the second continuation's broadcast goes through the `onAgentMessage` broadcast path (not the transport stream path), since the client only has one active stream. This could cause message duplication or missed updates on the client.

## Suggestion

Consider deduplicating queued continuations on the server when one is already pending — e.g., skip queuing a second `_queueAutoContinuation` if `_pendingAutoContinuation` is already `true`, or coalesce multiple tool results into a single continuation turn.

## Context

Identified during review of #1162. Not a regression from that PR — the double-queue pattern existed before — but the new client-side `startToolContinuation` guard makes the asymmetry between client (one stream) and server (two turns) more visible.

whoiskatrin:
## Summary
- coalesce rapid client-side tool results and approvals into a single queued auto-continuation turn instead of scheduling duplicate continuation runs
- preserve the queued continuation context until the matching continuation stream actually starts so request body and client tools do not get replaced by later requests
- add regression coverage for duplicate auto-continuations both while a chat turn is active and while the chat is idle, plus a patch changeset for `@cloudflare/ai-chat`

## Testing
- `npm run test`
- `npm run test:react`
- `npm run check`

## Reviewer Notes
- the main behavior change is in `packages/ai-chat/src/index.ts`, where `_enqueueAutoContinuation()` now coalesces adjacent auto-continue events and `_awaitPendingAutoContinuationPrerequisite()` waits for every pending tool-result/approval apply before starting the single continuation turn
- coalescing uses a small `AUTO_CONTINUATION_COALESCE_MS` debounce so back-to-back client tool completions in the same effect cycle collapse into one server continuation without changing the serialized turn model
- `_startStream()` now only clears pending auto-continuation state when the stream belongs to that queued continuation request id; this avoids losing the original continuation body/client tools if another normal request starts first
- the new regressions in `packages/ai-chat/src/tests/chat-turn-serialization.test.ts` are the best place to sanity-check the bug: before this patch they record two continuation request ids, after the patch they record one

## Context
- fixes #1171

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1179" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1331,6 +1331,7 @@
       "resolved": "https://registry.npmjs.org/@ai-sdk/openai-compatible/-/openai-compatible-2.0.35.tgz",
       "integrity": "sha512-g3wA57IAQFb+3j4YuFndgkUdXyRETZVvbfAWM+UX7bZSxA3xjes0v3XKgIdKdekPtDGsh4ZX2byHD0gJIMPfiA==",
       "license": "Apache-2.0",
+      "peer": true,
       "dependencies": {
         "@ai-sdk/provider": "3.0.8",
         "@ai-sdk/provider-utils": "4.0.19"
@@ -1381,6 +1382,7 @@
       "resolved": "https://registry.npmjs.org/@ai-sdk/provider/-/provider-3.0.8.tgz",
       "integrity": "sha512-oGMAgGoQdBXbZqNG0Ze56CHjDZ1IDYOwGYxYjO5KLSlz5HiNQ9udIXsPZ61VWaHGZ5XW/jyjmr6t2xz2jGVwbQ==",
       "license": "Apache-2.0",
+      "peer": true,
       "dependencies": {
         "json-schema": "^0.4.0"
       },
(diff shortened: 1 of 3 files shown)


threepointone:
my keepalive addition in your previous pr introduced a bug in this pr. fixed and pushed. it's pretty solid otherwise. let's see what ci/devin says.

PR #1235, opened 2026-03-30
fix(ai-chat): strip messageId from continuation start chunks (#1229)

### LINKED ISSUE #1229: ai-chat: tool continuation sends new messageId in start chunk, causing duplicate assistant message
## Problem

When a tool continuation stream starts after a client tool result, the server sends a `start` chunk with a **new `messageId`**. The AI SDK's `processUIMessageStream` sets `state.message.id = chunk.messageId`, which replaces the inherited assistant message ID. This causes the AI SDK's `write()` function to push a new message instead of replacing the existing one, resulting in a **temporary duplicate assistant message** visible to the user.

The duplicate gets cleaned up when `CF_AGENT_CHAT_MESSAGES` arrives with the correct server state, but it's visible during the continuation stream.

## Root Cause

In `_createToolContinuationStream` (react.js), raw chunks from the server are forwarded to the AI SDK without filtering:

```javascript
const chunk = JSON.parse(data.body);
controller.enqueue(chunk);  // start chunk has messageId that causes duplicate
```

The AI SDK then processes the `start` chunk:
```javascript
// ai/dist/index.js
case "start": {
    if (chunk.messageId != null) {
        state.message.id = chunk.messageId;  // overrides inherited ID
    }
    // ...
    write();  // pushMessage instead of replaceMessage
}
```

The `write()` function checks `activeResponse.state.message.id === this.lastMessage?.id`. Since the ID was just replaced, it doesn't match → `pushMessage` → duplicate.

## Steps to Reproduce

1. Create an agent with a client-side tool (e.g., `askUserClarification`)
2. Send a message that triggers the tool
3. Answer the tool's question
4. Observe: 2 assistant messages appear briefly during the continuation stream
5. When the stream finishes, `CF_AGENT_CHAT_MESSAGES` corrects it to 1 assistant message

## Workaround

Strip `messageId` from `start` chunks in `_createToolContinuationStream`:

```javascript
if (chunk.type === "start" && chunk.messageId != null) {
    const { messageId: _, ...rest } = chunk;
    controller.enqueue(rest);
} else {
    controller.enqueue(chunk);
}
```

This lets the AI SDK keep the inherited assistant message ID from `createStreamingUIMessageState`, which correctly reuses the existing assistant message.

## Environment

- `@cloudflare/ai-chat`: 0.3.0
- `ai`: 6.0.141
- `agents`: 0.8.7

whoiskatrin:
## Summary

When a tool continuation stream starts after a client tool result, the AI SDK emits a `start` chunk with a fresh `messageId`. On the client, `processUIMessageStream` uses this to override the assistant message ID, which causes the AI SDK's `write()` to push a new message instead of replacing the existing one — resulting in a brief duplicate assistant message.

The server already had a `!continuation` guard on its own message-building path (skipping `messageId` for its internal `message.id`), but was still forwarding the raw `messageId` to clients in the broadcast. This PR strips `messageId` from `start` chunks **server-side** in `_streamSSEReply` before both storage and broadcast, so all clients — not just `WebSocketChatTransport` — are protected.

### Changes

- **`packages/ai-chat/src/index.ts`** — In the chunk-rewriting section of `_streamSSEReply`, strip `messageId` from `start` chunks when `continuation` is true. This sits alongside the existing `finishReason → messageMetadata` transform. Updated the comment to document both transforms.
- **`packages/ai-chat/src/tests/worker.ts`** — Added `sseWithMessageId` body flag to `TestChatAgent` so it can return an SSE response with a `start` chunk containing `messageId` (simulating real `streamText` output).
- **`packages/ai-chat/src/tests/client-tools-continuation.test.ts`** — End-to-end test: triggers a tool-result auto-continuation with an SSE response, uses a second connection to passively observe broadcast chunks, and asserts that continuation `start` chunks have `messageId` stripped.
- **`.changeset/cool-oranges-drum.md`** — Patch changeset for `@cloudflare/ai-chat`.

Closes #1229

## Testing

```bash
npm run test:workers -- src/tests/client-tools-continuation.test.ts
```

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1235" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


threepointone:
Rewrote the approach in this PR — here's what changed and why:

**Moved the fix from client-side to server-side.** The original PR stripped `messageId` in `WebSocketChatTransport._createToolContinuationStream()` on the client. The new version strips it in `_streamSSEReply()` on the server, before both storage and broadcast. This is the right layer because:

- The server already knew `continuation=true` and was guarding its own `message.id` with `!continuation` at line 3260 — but was still forwarding the raw `messageId` to clients. That inconsistency was the actual bug.
- The server-side fix protects **all** clients (other transports, third-party consumers, different SDK versions), not just `WebSocketChatTransport`.
- Stored stream chunks (used for replay on reconnect) are also cleaned, avoiding the same duplicate on reconnection.

**Removed the client-side transport fix and its test** (`ws-chat-transport.ts`, `ws-transport-resume.test.ts`). No longer needed — the problematic `messageId` never reaches the client.

**Removed the React integration test** (`use-agent-chat.test.tsx`). It was testing a scenario that can't happen with the server-side fix.

**Rewrote the server-side test** to actually exercise the SSE code path. The original test used `TestChatAgent`'s default `text/plain` response, which goes through `_sendPlaintextReply` (no `start` chunks). The fix lives in `_streamSSEReply`. So the test was passing vacuously — the `for` loop over start chunks ran zero iterations. The new test:
- Adds an `sseWithMessageId` flag to `TestChatAgent` that returns an SSE response with `{ type: "start", messageId: "..." }`.
- Uses two connections (A triggers the tool result, B passively observes broadcasts) because the originating connection is excluded from live broadcasts via `_pendingResumeConnections` until it ACKs.
- Asserts `startChunks.length > 0` to prove the SSE path is actually exercised, then asserts `messageId` is absent.

PR #1196, opened 2026-03-25
feat(experimental): Postgres session providers with Hyperdrive support

mattzcarey:
## Summary

Adds Postgres-backed session providers for storing conversation history, context blocks, and searchable knowledge in an external database via Hyperdrive. This enables cross-DO queries, analytics, and shared state without relying on DO SQLite.

### What's new

**Providers** (`packages/agents/src/experimental/memory/session/providers/`)
- `PostgresSessionProvider` — tree-structured messages, compaction overlays, message FTS via tsvector
- `PostgresContextProvider` — writable context block storage (memory, cached prompt)
- `PostgresSearchProvider` — searchable knowledge base with tsvector + GIN index

**Framework improvements** (`packages/agents/src/experimental/memory/session/`)
- `Session.create()` accepts `SessionProvider` for external storage (in addition to `SqlProvider` for DO SQLite)
- Hardened context tools: removed enum constraints on label params, validate inside execute, always return error strings instead of throwing (prevents orphaned tool calls with smaller LLMs)
- Simplified `set_context` API: removed `key` param, auto-generates keys from `title` or content slug
- Fixed prompt lifecycle: `freezeSystemPrompt()` returns cached, `refreshSystemPrompt()` force-reloads from providers
- `clearMessages()` calls `refreshSystemPrompt()` to invalidate the cached prompt
- `appendToBlock()` adds newline separator between entries
- Empty writable blocks render in system prompt so the LLM knows they exist
- Clean block tags: `[readonly]`, `[searchable]`, `[loadable]`, `[not searchable]`
- Search provider `get()` returns entry count only (no key listing)

**Example** (`experimental/session-planetscale/`)
- Full Vite + React chat app with Hyperdrive + `pg` driver
- System prompt toggle, FULLTEXT search bar, connection indicator, theme toggle
- `wrapPgClient` helper converts `?` placeholders to `$1, $2, ...` for pg compatibility

**Tests** (`packages/agents/src/tests/experimental/memory/session/postgres-providers.test.ts`)
- 37 tests covering: provider CRUD, message round-trip, dynamic-tool part serialization, `convertToModelMessages` compatibility, prompt lifecycle (freeze/refresh/invalidation/concurrent)

**Docs** (`docs/sessions.md`)
- Postgres setup guide: migration SQL, Hyperdrive config, wire-up code
- System prompt lifecycle docs
- Search provider docs (message search + knowledge search)

### Migration SQL

Customers run this once — providers never create tables:

```sql
[...]

PR #1297, opened 2026-04-13
feat(experimental): Postgres session providers with Hyperdrive support

mattzcarey:
## Summary

Adds Postgres-backed session providers for storing conversation history, context blocks, and searchable knowledge in an external database via Hyperdrive. This enables cross-DO queries, analytics, and shared state without relying on DO SQLite.

Supersedes #1196.

### What's new

**Providers** (`packages/agents/src/experimental/memory/session/providers/`)
- `PostgresSessionProvider` — tree-structured messages, compaction overlays, message FTS via tsvector
- `PostgresContextProvider` — writable context block storage (memory, cached prompt)
- `PostgresSearchProvider` — searchable knowledge base with tsvector + GIN index

**Framework improvements** (`packages/agents/src/experimental/memory/session/`)
- `Session.create()` accepts `SessionProvider` for external storage (in addition to `SqlProvider` for DO SQLite)
- Hardened context tools: removed enum constraints on label params, validate inside execute, always return error strings instead of throwing (prevents orphaned tool calls with smaller LLMs)
- Simplified `set_context` API: removed `key` param, auto-generates keys from `title` or content slug
- Fixed prompt lifecycle: `freezeSystemPrompt()` returns cached, `refreshSystemPrompt()` force-reloads from providers
- `clearMessages()` calls `refreshSystemPrompt()` to invalidate the cached prompt
- `appendToBlock()` adds newline separator between entries
- Empty writable blocks render in system prompt so the LLM knows they exist
- Clean block tags: `[readonly]`, `[searchable]`, `[loadable]`, `[not searchable]`
- Search provider `get()` returns entry count only (no key listing)

**Example** (`experimental/session-planetscale/`)
- Full Vite + React chat app with Hyperdrive + `pg` driver
- System prompt toggle, FULLTEXT search bar, connection indicator, theme toggle
- `wrapPgClient` helper converts `?` placeholders to `$1, $2, ...` for pg compatibility

**Tests** (`packages/agents/src/tests/experimental/memory/session/postgres-providers.test.ts`)
- 37 tests covering: provider CRUD, message round-trip, dynamic-tool part serialization, `convertToModelMessages` compatibility, prompt lifecycle (freeze/refresh/invalidation/concurrent)

**Docs** (`docs/sessions.md`)
- Postgres setup guide: migration SQL, Hyperdrive config, wire-up code
- System prompt lifecycle docs
- Search provider docs (message search + knowledge search)

### Migration SQL

Customers run this once — providers never create tables:

```sql
CREATE TABLE assistant_messages (
  id TEXT PRIMARY KEY, session_id TEXT NOT NULL DEFAULT '', parent_id TEXT,
  role TEXT NOT NULL, content TEXT NOT NULL, created_at TIMESTAMPTZ DEFAULT NOW(),
  content_tsv TSVECTOR GENERATED ALWAYS AS (to_tsvector('english', content)) STORED
);
CREATE TABLE assistant_compactions (
  id TEXT PRIMARY KEY, session_id TEXT NOT NULL DEFAULT '',
  summary TEXT NOT NULL, from_message_id TEXT NOT NULL, to_message_id TEXT NOT NULL,
  created_at TIMESTAMPTZ DEFAULT NOW()
);
CREATE TABLE cf_agents_context_blocks (
  label TEXT PRIMARY KEY, content TEXT NOT NULL, updated_at TIMESTAMPTZ DEFAULT NOW()
);
CREATE TABLE cf_agents_search_entries (
  label TEXT NOT NULL, key TEXT NOT NULL, content TEXT NOT NULL,
  content_tsv TSVECTOR GENERATED ALWAYS AS (to_tsvector('english', content)) STORED,
  created_at TIMESTAMPTZ DEFAULT NOW(), updated_at TIMESTAMPTZ DEFAULT NOW(),
  PRIMARY KEY (label, key)
);
```

## Test plan

- [x] 37 unit tests passing (providers, round-trip, convertToModelMessages, prompt lifecycle)
- [ ] Deploy example and test chat flow end-to-end
- [ ] Verify search_context returns ranked results from knowledge base
- [ ] Verify clearMessages invalidates cached prompt
- [ ] Verify empty memory block renders in system prompt

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1297" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/docs/sessions.md
+++ b/docs/sessions.md
@@ -662,6 +662,197 @@ const myStorage: SessionProvider = {
 
 ---
 
+## Postgres (External Database)
+
+The built-in providers use Durable Object SQLite. If you need session data in an external Postgres database — for cross-DO queries, analytics, or shared state — use `PostgresSessionProvider` and `PostgresContextProvider`.
+
+These work with any Postgres-compatible database (Neon, Supabase, PlanetScale, etc.) via [Cloudflare Hyperdrive](https://developers.cloudflare.com/hyperdrive/) for connection pooling.
+
+### Setup
+
+#### 1. Create a Postgres database
+
+Use any Postgres provider and copy the connection string.
+
+#### 2. Create a Hyperdrive config
+
+```bash
+npx wrangler hyperdrive create my-session-db \
+  --connection-string="postgresql://user:password@host:port/dbname"
+```
+
+Copy the returned Hyperdrive ID.
+
+#### 3. Create the tables
+
+The Postgres user typically won't have `CREATE TABLE` permissions. Run this once in your database console:
+
+```sql
+CREATE TABLE IF NOT EXISTS assistant_messages (
+  id TEXT PRIMARY KEY,
+  session_id TEXT NOT NULL DEFAULT '',
+  parent_id TEXT,
+  role TEXT NOT NULL,
+  content TEXT NOT NULL,
+  text_content TEXT NOT NULL DEFAULT '',
+  created_at TIMESTAMPTZ DEFAULT NOW(),
+  content_tsv TSVECTOR GENERATED ALWAYS AS (to_tsvector('english', text_content)) STORED
+);
+CREATE INDEX IF NOT EXISTS idx_assistant_msg_parent ON assistant_messages (parent_id);
+CREATE INDEX IF NOT EXISTS idx_assistant_msg_session ON assistant_messages (session_id);
+CREATE INDEX IF NOT EXISTS idx_assistant_msg_fts ON assistant_messages USING GIN (content_tsv);
+
+CREATE TABLE IF NOT EXISTS assistant_compactions (
+  id TEXT PRIMARY KEY,
+  session_id TEXT NOT NULL DEFAULT '',
+  summary TEXT NOT NULL,
+  from_message_id TEXT NOT NULL,
+  to_message_id TEXT NOT NULL,
+  created_at TIMESTAMPTZ DEFAULT NOW()
+);
+
+CREATE TABLE IF NOT EXISTS cf_agents_context_blocks (
+  label TEXT PRIMARY KEY,
+  content TEXT NOT NULL,
+  updated_at TIMESTAMPTZ DEFAULT NOW()
+);
+
+CREATE TABLE IF NOT EXISTS cf_agents_search_entries (
+  label TEXT NOT NULL,
+  key TEXT NOT NULL,
+  content TEXT NOT NULL,
+  content_tsv TSVECTOR GENERATED ALWAYS AS (to_tsvector('english', content)) STORED,
+  created_at TIMESTAMPTZ DEFAULT NOW(),
+  updated_at TIMESTAMPTZ DEFAULT NOW(),
+  PRIMARY KEY (label, key)
+);
+CREATE INDEX IF NOT EXISTS idx_search_entries_fts ON cf_agents_search_entries USING GIN (content_tsv);
+```
+
+#### 4. Configure wrangler
+
+```jsonc
+{
+  "compatibility_flags": ["nodejs_compat"],
+  "hyperdrive": [
+    {
+      "binding": "HYPERDRIVE",
+      "id": "<your-hyperdrive-id>"
+    }
+  ],
+  "placement": {
+    "region": "aws:us-east-1" // match your database region
+  }
+}
+```
+
+#### 5. Wire it up
+
+```typescript
+import { Agent, callable } from "agents";
+import {
+  Session,
+  PostgresSessionProvider,
+  PostgresContextProvider,
+  PostgresSearchProvider,
+  type PostgresConnection
+} from "agents/experimental/memory/session";
+import { Client } from "pg";
+
+// Wrap pg Client to match the PostgresConnection interface
+function wrapPgClient(client: Client): PostgresConnection {
+  return {
+    async execute(query, args) {
+      let idx = 0;
+      const pgQuery = query.replace(/\?/g, () => `$${++idx}`);
+      const result = await client.query(pgQuery, args ?? []);
+      return { rows: result.rows };
+    }
+  };
+}
+
+class MyAgent extends Agent<Env> {
+  private _session?: Session;
+  private _pgClient?: Client;
+
+  private async getConnection(): Promise<PostgresConnection> {
+    if (!this._pgClient) {
+      this._pgClient = new Client({
+        connectionString: this.env.HYPERDRIVE.connectionString
+      });
+      await this._pgClient.connect();
+    }
+    return wrapPgClient(this._pgClient);
+  }
+
+  private async getSession(): Promise<Session> {
+    if (this._session) return this._session;
+
+    const conn = await this.getConnection();
+    const sessionId = this.ctx.id.toString();
+
+    this._session = Session.create(new PostgresSessionProvider(conn, sessionId))
+      .withContext("soul", {
+        provider: {
+          get: async () => "You are a helpful assistant."
+        }
+      })
+      .withContext("memory", {
+        description: "Short facts",
+        maxTokens: 1100,
+        provider: new PostgresContextProvider(conn, `memory_${sessionId}`)
+      })
+      .withContext("knowledge", {
+        description: "Searchable knowledge base",
+        provider: new PostgresSearchProvider(conn)
+      })
+      .withCachedPrompt(
+        new PostgresContextProvider(conn, `_prompt_${sessionId}`)
+      );
+
+    return this._session;
+  }
+}
+```
+
+### How it works
+
+When `Session.create()` receives a `SessionProvider` instead of a `SqlProvider`, it skips all SQLite auto-wiring. This means:
+
+- **Context blocks need explicit providers.** No auto-wiring to SQLite — each `withContext()` call needs a `provider` option, or the block will be read-only with no storage.
+- **`withCachedPrompt()` needs an explicit provider.** Pass a `PostgresContextProvider` to persist the frozen system prompt.
+- **Broadcaster is skipped.** WebSocket status broadcasts (`CF_AGENT_SESSION` events) only work with `SqlProvider`-based sessions.
+- **All Session methods are async.** `getHistory()`, `getMessage()`, etc. return Promises since the underlying storage is async.
+
+### System prompt lifecycle
+
+- **`freezeSystemPrompt()`** — returns the cached prompt from the store. On first call (cache miss), loads blocks from providers, renders, and persists. Subsequent calls return the stored value without re-rendering. This preserves LLM prefix cache hits.
+- **`refreshSystemPrompt()`** — force reloads blocks from providers, re-renders, and updates the store. Call this to invalidate the cached prompt (e.g. after `clearMessages`).
+
+### PostgresConnection interface
+
+```typescript
+interface PostgresConnection {
+  execute(
+    query: string,
+    args?: (string | number | boolean | null)[]
+  ): Promise<{ rows: Record<string, unknown>[] }>;
+}
+```
+
+The providers use `?` placeholders internally. When wrapping `pg`, convert to `$1, $2, $3` (see the `wrapPgClient` helper above). Any Postgres driver with a compatible `execute()` method works.
+
+### Search
+
+Two levels of search are available:
+
+- **Message search** — `PostgresSessionProvider.searchMessages()` searches conversation history via the `content_tsv` column on `assistant_messages`.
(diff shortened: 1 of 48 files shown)


experimental/session-planetscale/src/server.ts
```diff
@@ -0,0 +1,172 @@
...
+import type { UIMessage } from "ai";
+import { createWorkersAI } from "workers-ai-provider";
+import { generateText, convertToModelMessages, stepCountIs } from "ai";
+import { Client } from "pg";
+
+/**
+ * Wrap a pg Client to match the PostgresConnection interface.
+ * Converts `?` placeholders to `$1, $2, ...` for pg.
+ */
+function wrapPgClient(client: Client): PostgresConnection {
```
mattzcarey: this feels like boilerplate. could we put this into the postgres session provider or how can we clean this up so people dont have to do this?

packages/agents/src/experimental/memory/session/providers/postgres.ts
```diff
@@ -0,0 +1,300 @@
...
+  }
+
+  // ── Write ──────────────────────────────────────────────────────
+
+  async appendMessage(
+    message: SessionMessage,
+    parentId?: string | null
+  ): Promise<void> {
+    const parent =
+      parentId ?? ((await this.latestLeafRow())?.id as string) ?? null;
```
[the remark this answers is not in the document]
mattzcarey: is this true?

packages/agents/src/experimental/memory/session/providers/postgres.ts
```diff
@@ -0,0 +1,300 @@
...
+
+  private extractText(json: string): string {
+    const msg = this.parse(json);
+    if (!msg) return json;
+    return msg.parts
+      .filter((p) => p.type === "text" && p.text)
+      .map((p) => p.text)
+      .join("\n");
+  }
+}
```
mattzcarey: why do we need this? are we only saving the text?

packages/agents/src/experimental/memory/session/context.ts
```diff
@@ -596,29 +613,17 @@ export class ContextBlocks {
...
-      }
-
       const properties: Record<string, unknown> = {
         label: {
           type: "string" as const,
-          enum: writable.map((b) => b.label),
-          description: "Block label to write to"
+          description:
+            "Block label to write to. Must be one of: " +
+            writable.map((b) => `"${b.label}"`).join(", ")
```
mattzcarey: I think I prefered this as an enum

packages/agents/src/experimental/memory/session/context.ts
```diff
@@ -631,22 +636,14 @@ export class ContextBlocks {
...
-      if (keyedBlocks.some((b) => b.isSkill)) {
-        properties.description = {
-          type: "string" as const,
-          description: "Short description for the skill entry"
+            "Short title for the entry. Used as a stable identifier — " +
+            "entries with the same title are updated, different titles create new entries. " +
+            "Applies to: " +
+            keyedBlocks.map((b) => `"${b.label}"`).join(", ")
         };
       }
```
mattzcarey, line 690: seems like this is for skills right? 

so you can set_context(skill, title here, description here, content here)

need to make sure it is 

`set_context("label of block", {
  metadata_title: a good title,
  metadata_description: "some metadata"
  content: "this is the main thing here" 
})


nothing needs metadata but its good to have it for longer stuff like skills (loadables). we should have this in the description of this tool.

packages/agents/src/experimental/memory/session/context.ts
```diff
@@ -719,8 +710,9 @@ export class ContextBlocks {
           properties: {
             label: {
               type: "string" as const,
-              enum: skillLabels,
-              description: "Skill block label"
+              description:
+                "Skill block label. Must be one of: " +
+                skillLabels.map((l) => `"${l}"`).join(", ")
             },
```
mattzcarey, line 767: again. i prefer enums

packages/agents/src/experimental/memory/session/context.ts
```diff
@@ -780,16 +779,17 @@ export class ContextBlocks {
       toolSet.search_context = {
         description:
           "Search for information in a searchable context block. " +
-          "Available searchable blocks: " +
+          "ONLY these blocks are searchable: " +
           searchLabels.map((l) => `"${l}"`).join(", ") +
-          ".",
+          ". Other blocks (e.g. memory) cannot be searched.",
```
mattzcarey: do not eg anything here.

packages/agents/src/experimental/memory/session/context.ts
```diff
@@ -780,16 +779,17 @@ export class ContextBlocks {
...
         inputSchema: z.fromJSONSchema({
           type: "object" as const,
           properties: {
             label: {
               type: "string" as const,
-              enum: searchLabels,
-              description: "Searchable block label"
+              description:
+                "Searchable block label. Must be one of: " +
+                searchLabels.map((l) => `"${l}"`).join(", ")
```
mattzcarey: again enum is more concise here

packages/agents/src/experimental/memory/session/session.ts
```diff
@@ -81,13 +90,25 @@ export class Session {
...
+   * const session = Session.create(planetscaleProvider)
+   *   .withContext("memory", {
+   *     maxTokens: 1100,
+   *     provider: new PlanetScaleContextProvider(conn, "memory")
+   *   })
+   *   .withCachedPrompt(new PlanetScaleContextProvider(conn, "_prompt"));
    * ```
    */
-  static create(agent: SqlProvider): Session {
+  static create(storageOrAgent: SqlProvider | SessionProvider): Session {
```
mattzcarey: rename this to provider. not storageOrAgent

packages/agents/src/experimental/memory/session/session.ts
```diff
@@ -181,6 +212,9 @@ export class Session {
    */
   private _restoreLoadedSkills(): void {
     const history = this.storage.getHistory();
+    // If the provider is async, history is a Promise — skip restore for async providers
+    if (history instanceof Promise) return;
```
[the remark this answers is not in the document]
mattzcarey: this is bad we should await this.

packages/agents/src/experimental/memory/session/session.ts
```diff
@@ -227,8 +261,8 @@ export class Session {
    * Replace a load_context tool result in conversation history
    * with a short marker to reclaim context space.
    */
-  private _reclaimLoadedSkill(label: string, key: string): void {
-    const history = this.storage.getHistory();
+  private async _reclaimLoadedSkill(label: string, key: string): Promise<void> {
```
mattzcarey, line 329: what does reclaim mean in this context?

packages/agents/src/experimental/memory/session/session.ts
```diff
@@ -468,8 +503,13 @@ export class Session {
     const opts = options ?? {};
     let provider = opts.provider;
     if (!provider) {
+      if (!this._agent) {
+        throw new Error(
+          `addContext("${label}") requires an explicit provider when Session uses a SessionProvider`
```
mattzcarey, line 612: is the tool `addContext` or `add_context`

packages/think/src/think.ts
```diff
@@ -2400,7 +2421,12 @@ export class Think<
           parts: result.parts as UIMessage["parts"]
         };
         const safe = enforceRowSizeLimit(sanitizeMessage(updatedMsg));
-        this.session.updateMessage(safe);
+        await this.session.updateMessage(safe);
+        // Update cache if this message exists there
+        const cacheIdx = this._cachedMessages.findIndex(
+          (m) => m.id === safe.id
+        );
+        if (cacheIdx !== -1) this._cachedMessages[cacheIdx] = safe as UIMessage;
```
mattzcarey: what do we need this for? I cant rememeber

docs/sessions.md
```diff
@@ -662,6 +662,198 @@ const myStorage: SessionProvider = {
...
+
+Copy the returned Hyperdrive ID.
+
+#### 3. Create the tables
+
+The Postgres user typically won't have `CREATE TABLE` permissions. Run this once in your database console:
+
+```sql
+CREATE TABLE IF NOT EXISTS assistant_messages (
+  id TEXT PRIMARY KEY,
```
whoiskatrin: I think this should use session_id as part of the primary key.

Since this table is shared across sessions, id by itself makes message ids global to the whole db. If two sessions ever use the same message id, one of them gets silently skipped.

Could we make this use

```sql
PRIMARY KEY (session_id, id)
```

and update the insert conflict to match

whoiskatrin (commented):
Nice, I like this direction. The provider split feels right to me and I’m glad pg is only in the example, not in the package itself.

I do think there are a few things to fix before this lands though.

packages/agents/src/experimental/memory/session/providers/postgres.ts
```diff
@@ -0,0 +1,328 @@
...
+      parentId !== undefined
+        ? parentId
+        : (((await this.latestLeafRow())?.id as string | undefined) ?? null);
+    const json = JSON.stringify(message);
+    const text = this.extractSearchableText(json);
+
+    await this.conn.execute(
+      `INSERT INTO assistant_messages (id, session_id, parent_id, role, content, text_content)
+       VALUES (?, ?, ?, ?, ?, ?)
+       ON CONFLICT (id) DO NOTHING`,
```
whoiskatrin: I think this should match the schema change above and conflict on session_id plus id, not just id.

Otherwise one session can accidentally block another session from inserting a message with the same id.

packages/agents/src/experimental/memory/session/providers/postgres.ts
```diff
@@ -0,0 +1,328 @@
...
+  async appendMessage(
+    message: SessionMessage,
+    parentId?: string | null
+  ): Promise<void> {
+    // Honour the `SessionProvider` contract:
+    //   - `undefined` / omitted → auto-detect (attach to latest leaf)
+    //   - explicit `null`       → create a root message with no parent
+    // Using `??` here would collapse those two cases; `parentId !== undefined`
+    // preserves the distinction.
+    const parent =
```
whoiskatrin: Can we check parentId here like the SQLite provider does

Right now if parentId belongs to another session, we still store it. Then getHistory filters by this session_id while walking the tree, so the history can get cut off in a weird way.

Probably better to check the parent exists in this session, and use null if it does not.

whoiskatrin:
Looks like this needs a changeset.

This adds new public exports and changes some Think methods to async, so I think we should have release notes for it.

whoiskatrin:
One more thing, docs/sessions.md has a section that still reads like Session is mostly sync.

I think those examples should use await, and the note saying updateMessage deleteMessages and clearMessages are sync should be removed.

Maybe just document the rule as, Session methods are async. SQLite is fast, external storage may do IO.

threepointone:
looking into this PR today

whoiskatrin (approved):
...

packages/think/src/think.ts
```diff
@@ -697,11 +706,32 @@ export class Think<
...
+        switch (event.type) {
+          case "append":
+            if (!event.inserted || event.parentId !== undefined) {
+              await this._syncMessages();
+            } else {
+              this._upsertCachedMessage(event.message as UIMessage);
+            }
+            break;
+          case "update":
+            this._upsertCachedMessage(event.message as UIMessage);
```
whoiskatrin: wondering if this should only patch messages that are already in the live cache?

if an older branch gets updated, or `updateMessage()` is called for something not on the current history path, `_upsertCachedMessage` looks like it would append it into `this.messages`. since turns now read from that cache instead of `session.getHistory()`, could that make off-branch / missing messages show up in model context?

packages/agents/src/experimental/memory/session/providers/postgres.ts
```diff
@@ -0,0 +1,337 @@
...
+    const { rows } = await this.conn.execute(
+      "SELECT * FROM assistant_compactions WHERE session_id = ? ORDER BY created_at ASC",
+      [this.sessionId]
+    );
+    return rows.map((r) => ({
+      id: r.id as string,
+      summary: r.summary as string,
+      fromMessageId: r.from_message_id as string,
+      toMessageId: r.to_message_id as string,
+      createdAt: r.created_at as string
```
whoiskatrin: small question, does raw `pg.Client` return `TIMESTAMPTZ` columns as `Date` here?

if so this cast might leave `createdAt` as a runtime `Date` even though `StoredCompaction` says string. should we normalize this to an ISO string before returning it?

packages/think/src/think.ts
threepointone: good catch, fix incoming

packages/agents/src/experimental/memory/session/providers/postgres.ts
threepointone: good catch, fix incoming

PR #1358, opened 2026-04-22
fix(ai-chat): don't crash when useAgent()'s HTTP URL isn't ready yet

### LINKED ISSUE #1356: useAgent getHttpUrl() returns empty string before WebSocket connects, causing ai-chat to crash
## Bug

`useAgent()` from `agents/react` returns an agent object where `getHttpUrl()` returns `""` before the WebSocket connection is established. `@cloudflare/ai-chat`'s `useAgentChat()` then does `new URL(agent.getHttpUrl())` (line 516 of react.js) which throws `TypeError: Failed to construct 'URL': Invalid URL`.

This crashes the component on first render before the WebSocket has time to connect.

## Reproduction

```tsx
import { useAgent } from 'agents/react'
import { useAgentChat } from '@cloudflare/ai-chat/react'

function Chat() {
  const agent = useAgent({ agent: 'my-agent', name: 'test' })
  
  // This crashes immediately because agent.getHttpUrl() returns ""
  // and ai-chat does new URL("") internally
  const chat = useAgentChat({ agent })
  
  return <div>{/* ... */}</div>
}
```

## Root cause

In `agents@0.11.4`, `react.js` line 264:

```js
agent.getHttpUrl = () => {
  return (agent._url || agent._pkurl || "").replace("ws://", "http://").replace("wss://", "https://");
};
```

When neither `_url` nor `_pkurl` is set (before WebSocket connects), this returns `""`.

In `@cloudflare/ai-chat@0.4.6`, `react.js` line 516:

```js
const agentUrl = new URL(agent.getHttpUrl());
```

`new URL("")` throws `TypeError: Invalid URL`.

## Versions

- `agents@0.11.4`
- `@cloudflare/ai-chat@0.4.6`
- `@ai-sdk/react@3.0.170`

## Current workaround

We patch `getHttpUrl` in our `useAgent` wrapper to return a placeholder URL:

```ts
const nativeGetHttpUrlRef = useRef(theAgent.getHttpUrl)
theAgent.getHttpUrl = () => {
  const url = nativeGetHttpUrlRef.current?.() ?? ''
  return url || 'http://localhost'
}
```

## Suggested fix

Either in `agents`:
```js
agent.getHttpUrl = () => {
  const raw = (agent._url || agent._pkurl || "");
  if (!raw) return "about:blank"; // or throw a descriptive error
  return raw.replace("ws://", "http://").replace("wss://", "https://");
};
```

Or in `ai-chat`:
```js
const rawUrl = agent.getHttpUrl();
const agentUrl = rawUrl ? new URL(rawUrl) : null;
// guard against null agentUrl downstream
```

threepointone:
## Summary

Closes #1356.

`useAgentChat()` called `new URL(agent.getHttpUrl())` unconditionally during render. `useAgent()` builds `getHttpUrl()` from PartySocket internals that can legitimately still be `""` before the WebSocket connects, which happens most often when the agent is behind a proxy or reached via `basePath`/custom routing. That threw `TypeError: Failed to construct 'URL': Invalid URL` and crashed the component on first render, well before the handshake completed.

This PR treats "HTTP URL not ready yet" as a first-class state in `useAgentChat()`:

- Guards `new URL(...)` so an empty result just yields `null`.
- Defers the built-in `/get-messages` fetch until the URL is available, and the default fetcher returns `[]` if called with an empty URL.
- Keeps custom `getInitialMessages` callbacks runnable before the URL exists. `GetInitialMessagesOptions.url` is now `string | undefined`; callers that previously typed `url: string` should widen to `url?: string`.
- Stabilizes the `useChat` `id` across the URL-arrival transition so the AI SDK doesn't recreate the underlying `Chat` (and abandon in-flight resume) just because the URL materialized on render 2.
- Seeds initial messages exactly once when the load promise resolves after mount (only if the chat is empty), and marks the current cache key as seeded from `clearHistory()`, `CF_AGENT_CHAT_CLEAR`, and explicit `setMessages([])` so user-driven clears aren't re-hydrated.

No API break. Apps where `getHttpUrl()` was synchronously available on first render are unchanged.

## Test plan

- [x] Added `"should wait for a valid HTTP URL before fetching initial messages"` — starts with an empty URL, verifies no `fetch()` happens, then makes the URL available and confirms `/get-messages` is called exactly once with the normalized HTTP URL.
- [x] Added `"should allow custom initial message loaders before the HTTP URL is ready"` — verifies a user-provided `getInitialMessages` runs with `url: undefined` and seeds the chat.
- [x] `npm run test:react` in `packages/ai-chat` — 47/47 passing.
- [x] `npm run check` at the repo root — formatting, lints, exports, and all 75 typecheck projects pass.

## Changeset

A patch-level changeset for `@cloudflare/ai-chat` is included at `.changeset/ai-chat-pending-http-url.md`.

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1358" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/ai-chat-pending-http-url.md
+++ b/.changeset/ai-chat-pending-http-url.md
@@ -8,6 +8,7 @@ Fix `useAgentChat()` crashing on first render when `agent.getHttpUrl()` returns 
 
 - The built-in `/get-messages` fetch is deferred until the URL is known, and applied exactly once when it resolves (empty chats only — existing messages are never overwritten).
 - Custom `getInitialMessages` callbacks continue to run and are passed `url: undefined` so they can load from other sources if they don't need the socket URL. `GetInitialMessagesOptions.url` is now `string | undefined`; callers that previously typed `url: string` should widen to `url?: string`.
+- Initial messages are cached by agent identity (class + name) rather than by URL + identity, so the URL-arrival transition no longer invalidates the cache, re-invokes the loader, or re-triggers Suspense once the chat has already been populated.
 - The underlying `useChat` instance keeps a stable `id` across the URL-arrival transition, so in-flight stream resume and chat state are preserved.
 
 No API or behavior changes for apps where the URL was already available synchronously on first render.
--- a/.changeset/ai-chat-pending-http-url.md
+++ b/.changeset/ai-chat-pending-http-url.md
@@ -1,0 +1,13 @@
+---
+"@cloudflare/ai-chat": patch
+---
+
+Fix `useAgentChat()` crashing on first render when `agent.getHttpUrl()` returns an empty string. This happened in setups where the WebSocket handshake hadn't completed by the time React rendered — most commonly when the agent is reached through a proxy or custom-routed worker — because `@cloudflare/ai-chat` unconditionally called `new URL(agent.getHttpUrl())`. See [#1356](https://github.com/cloudflare/agents/issues/1356).
+
+`useAgentChat()` now treats a missing HTTP URL as "not ready yet":
+
+- The built-in `/get-messages` fetch is deferred until the URL is known, and applied exactly once when it resolves (empty chats only — existing messages are never overwritten).
+- Custom `getInitialMessages` callbacks continue to run and are passed `url: undefined` so they can load from other sources if they don't need the socket URL. `GetInitialMessagesOptions.url` is now `string | undefined`; callers that previously typed `url: string` should widen to `url?: string`.
+- The underlying `useChat` instance keeps a stable `id` across the URL-arrival transition, so in-flight stream resume and chat state are preserved.
+
+No API or behavior changes for apps where the URL was already available synchronously on first render.
--- a/packages/ai-chat/src/react-tests/use-agent-chat.test.tsx
+++ b/packages/ai-chat/src/react-tests/use-agent-chat.test.tsx
@@ -294,6 +294,62 @@ describe("useAgentChat", () => {
     expect(getInitialMessages).toHaveBeenCalledTimes(1);
   });
 
+  it("should invoke custom getInitialMessages only once across the HTTP URL transition", async () => {
+    let url = "";
+    const agent = createAgent({
+      name: "thread-custom-transition",
+      url
+    });
+    agent.getHttpUrl = () =>
+      url.replace("ws://", "http://").replace("wss://", "https://");
+
+    const testMessages = [
+      {
+        id: "1",
+        role: "assistant" as const,
+        parts: [{ type: "text" as const, text: "One call only" }]
+      }
+    ];
+
+    const getInitialMessages = vi.fn(async () => testMessages);
+
+    const TestComponent = () => {
+      const chat = useAgentChat({ agent, getInitialMessages });
+      return <div data-testid="messages">{JSON.stringify(chat.messages)}</div>;
+    };
+
+    const screen = await act(async () => {
+      const screen = render(<TestComponent />, {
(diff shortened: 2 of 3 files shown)


threepointone:
Good catch — confirmed the reproduction: with a custom `getInitialMessages`, the first render cached the promise under the identity-only key (URL was empty), and the second render (URL resolved) built a URL-prefixed cache key, missed the cache, re-invoked the loader, and `use(newPromise)` re-triggered Suspense after the chat was already populated.

Fixed in 5f6e794c by caching initial messages by agent identity only. The URL-aware key still exists as `resolvedInitialMessagesCacheKey` for the `stableChatIdRef` upgrade logic (which separately needs to tell "URL arrived" apart from "identity changed"), but `requestCache` and the late-seed guard both key on identity.

Added a regression test — `should invoke custom getInitialMessages only once across the HTTP URL transition` — that fails on the previous commit and passes on the fix. Changeset description is also updated to call out the identity-only caching.

threepointone:
Valid too — confirmed the re-hydration race with a failing test before the fix.

Reproduction: `getHttpUrl()` returns a valid URL on first render → `useChat({ messages: initialMessages })` seeds the Chat directly → `chatMessages.length > 0` by the time the late-seed effect first runs → the effect early-returned without calling `markInitialMessagesSeeded()`, so the seeded-key ref stayed `null` forever. Any later path that emptied `chatMessages` without going through the wrapper (notably the `CF_AGENT_CHAT_MESSAGES` server broadcast handler, which does `setMessages(preserveProtectedStreamingAssistant(data.messages))`) tripped all three guards and re-hydrated the stale initial messages.

Fixed in 37e0093a by marking the cache key seeded on every settled observation, not just when we actively inject messages. The chat-is-already-populated branch now marks and returns.

Regression test: `should not re-hydrate initial messages when a server broadcast empties the chat`. It fails on 5f6e794c and passes on 37e0093a.

threepointone:
Good observation, and the conclusion is correct. Two structural guarantees back it up beyond "works in practice":

1. **React Suspense blocks effect commit.** If `use(initialMessagesPromise)` suspends on first render, the component never commits, so neither the late-seed effect nor the `onAgentMessage` listener attach run. The listener can't see a broadcast before initial messages settle.

2. **Top-down effect declaration order.** Once the component does commit, React runs effects in top-down declaration order. The late-seed effect is declared at `react.tsx:984`; the `onAgentMessage` listener effect is declared at `react.tsx:1410`. So late-seed always marks the cache key seeded *before* the `message` listener is attached. Any subsequent `CF_AGENT_CHAT_MESSAGES` broadcast goes through the already-set guard.

So the ordering dependency isn't a timing race — it's enforced by React.

**One edge case that isn't protected**, not flagged in the comment but worth being honest about: two `useAgentChat` hooks mounted concurrently against the *same* agent identity (e.g. a chat widget rendered alongside a fullscreen view). `useChat({ id })` shares the `Chat` instance across hooks with identical ids, so if hook A clears and hook B mounts afterwards, hook B sees `chatMessages.length === 0` with its own fresh `seededInitialMessagesKeyRef === null` and will run the seed branch, undoing the clear. The seeded ref is per-hook, not per-identity.

Resolving that would require lifting the "seeded" record to module-level state keyed by identity. Given how niche the co-mounted-duplicate-hook pattern is, and that it interacts with the already-module-level `requestCache` in subtle ways (re-mounting should arguably re-fetch), I'd prefer to land the three fixes in this PR and address the multi-instance case separately if it surfaces as a real report.

Issue #1365, opened 2026-04-23
State tracking gap during client-defined tool round-trips


alexanderjacobsen:
## Context

This is specifically about **client-defined tools** — tools registered via the `tools` option on `useAgentChat`, with `execute` functions that run in the browser. This path is documented as the intended way for SDKs and platforms that need to register tools dynamically at runtime, where the server can't know the tool set at deploy time.

During a single user turn that involves one or more client tool calls, the exposed state (`status`, `isStreaming`, `isServerStreaming`) doesn't coherently represent what's actually happening. Two distinct issues, both rooted in the same area.

## Issue 1: State gap during client-tool execution

The sequence today:

```
AI emits tool-call → stream ends                                → status: 'ready', isStreaming: false
SDK invokes onToolCall(...)                                     → status: 'ready', isStreaming: false
  [consumer code awaits tool.execute(), can take seconds]       → status: 'ready', isStreaming: false
addToolOutput(result)                                           → deferred stream created, status: 'submitted', isStreaming: true
Server pushes continuation                                      → isServerStreaming: true
```

From the consumer's perspective, the entire turn should look like one continuous \"something is happening\" signal. Instead, there's a real window — the duration of the client-side `tool.execute` — where no state field reflects any activity. For fast tools this is imperceptible; for anything doing a fetch/API call, it's noticeable.

The root cause is that the deferred stream (which exists specifically to cover this gap) is only created on `addToolOutput`, not on `onToolCall` invocation. Since the SDK is the one calling `onToolCall`, it could easily bracket that invocation with the deferred stream creation. Proposed change:

```ts
// Conceptually, inside the SDK's tool dispatch
if (isClientTool(toolCall)) {
  this._createToolContinuationStream();  // moved earlier
  try {
    await options.onToolCall({ toolCall, addToolOutput });
  } catch { /* existing error handling */ }
}
```

## Issue 2: `status` conflates user-initiated and server-pushed submissions

The new separation between `status` (user-initiated request lifecycle) and `isServerStreaming` (server-pushed activity) is the right abstraction. But tool-call continuations currently flip `status` to `'submitted'` via the deferred stream, which breaks the separation — `'submitted'` can now mean either:

1. \"User just sent a new message, awaiting first token\" (a new turn)
2. \"Client tool result was just returned, awaiting server continuation\" (mid-turn)

Consumers who want to differentiate these (e.g. to show a typing indicator only for #1) have to inspect message history to tell them apart.

[...]

Issue #1157, opened 2026-03-23
With client side tool calls `status` remains `ready` after initial tool call.


fragglebob:
**Describe the bug**

We are making use of client side tools calls with `@cloudflare/ai-chat`. We are looking at adding a status indicator to the interface, but after the initial tool call comes through to the client. The `status` never seems to be anything other than `ready`, even when things are streaming back during the next turn.

**To Reproduce**

1. Have a client side tool call
2. LLM calls client side tool
3. User takes action to generate tool output - calls `addToolOutput`
4. `status` remains `ready`

**Expected behavior**

After calling `addToolOutput` the `status` would be `submitted`. Then when response begins streaming back, `status` would be `streaming`, until streaming is complete.

**Version:**
`agents`: 0.7.9
`@cloudflare/ai-chat`: 0.1.9

whoiskatrin:
@fragglebob do you mind giving this a spin? https://github.com/cloudflare/agents/pull/1161

this should fix it

fragglebob:
@whoiskatrin Working perfectly. 😄

PR #1368, opened 2026-04-23
feat(ai-chat): expose isToolContinuation to disambiguate submitted status (issue #1365)

threepointone:
## Summary

Closes the second half of #1365 — consumers who want a typing indicator *only* for new user messages had to inspect message history to tell a fresh submission apart from a mid-turn tool continuation.

Rather than re-designing `status` (which [#1157](https://github.com/cloudflare/agents/issues/1157) explicitly asked to track the whole tool round-trip, and which many loading-spinner UIs now rely on), this adds a purely additive disambiguation flag:

\`\`\`tsx
const { status, isStreaming, isToolContinuation } = useAgentChat({ ... });

const isLoading = isStreaming || status === \"submitted\";
const showTypingIndicator = status === \"submitted\" && !isToolContinuation;
\`\`\`

`isToolContinuation` is `true` from the moment `addToolOutput` / `addToolApprovalResponse` kicks off an auto-continuation until the continuation stream closes (or is aborted by `stop()`). It's `false` otherwise — including for cross-tab server broadcasts, which surface via `isServerStreaming` only.

Implementation mirrors the existing `resumingToolContinuationRef` (kept for its synchronous re-entry guard) as React state so the UI re-renders on flip.

## Why additive, not a redesign

#1157 asked for `status` to be `submitted`/`streaming`/`ready` across tool round-trips, and PR #1161 shipped that. The existing tests at \`useAgentChat tool continuation status (issue #1157)\` lock in that contract. Making `status` stay `ready` during continuations (as #1365 initially proposes) would regress #1157's users without warning. This PR gives #1365's author the semantics they need *without* touching the ones #1157's users depend on.

## Test plan

New describe block \`useAgentChat isToolContinuation (issue #1365)\` covers:
- [x] False on mount and while idle.
- [x] Full lifecycle after `addToolOutput` (true during handshake, true while chunks arrive, false after `done:true`). Asserts `status` still follows the #1157 contract throughout.
- [x] Stays false when `autoContinueAfterToolResult: false` (no continuation happens).
- [x] Flips true after `addToolApprovalResponse` and back to false when the continuation ends.
- [x] Resets to false when `stop()` is called mid-continuation.
- [x] Stays false for cross-tab server broadcasts (not our continuation) while `isServerStreaming` correctly goes true.

Verification:
- [x] \`npx vitest --project react --run\` — 55/55 pass (2 new-this-PR + the 2 existing #1157 tests still green)
- [x] \`tsc --noEmit\`, \`oxlint\`, \`oxfmt\` all clean
- [x] Changeset included (\`@cloudflare/ai-chat\` patch)

## Relationship to #1366

#1366 (\"surface busy signal while onToolCall is awaiting\") closed the first half of #1365 — the invisible gap during \`tool.execute()\`. This PR closes the second half — disambiguating the \`submitted\` status. The two PRs are independent; touches on the same file but easy to rebase whichever lands second.

Made-with: Cursor

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1368" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/ai-chat-is-tool-continuation.md
+++ b/.changeset/ai-chat-is-tool-continuation.md
@@ -4,7 +4,7 @@
 
 Add `isToolContinuation: boolean` to `useAgentChat()` so consumers can disambiguate a fresh user-initiated `status === "submitted"` from one driven by a server-pushed tool continuation. See [#1365](https://github.com/cloudflare/agents/issues/1365).
 
-`status` already tracks the whole tool round-trip (`submitted` → `streaming` → `ready`) after `addToolOutput` / `addToolApprovalResponse`, on purpose — that's what [#1157](https://github.com/cloudflare/agents/issues/1157) asked for and what many loading-spinner UIs now rely on. But some consumers want a typing indicator *only* for new user messages, not for mid-turn continuations, and previously had to inspect message history to tell them apart.
+`status` already tracks the whole tool round-trip (`submitted` → `streaming` → `ready`) after `addToolOutput` / `addToolApprovalResponse`, on purpose — that's what [#1157](https://github.com/cloudflare/agents/issues/1157) asked for and what many loading-spinner UIs now rely on. But some consumers want a typing indicator _only_ for new user messages, not for mid-turn continuations, and previously had to inspect message history to tell them apart.
 
 `isToolContinuation` is `true` from the moment `addToolOutput` / `addToolApprovalResponse` kicks off an auto-continuation until the continuation stream closes (or is aborted by `stop()`). It is `false` otherwise — including during cross-tab server broadcasts, which surface via `isServerStreaming` only.
 
--- a/.changeset/ai-chat-is-tool-continuation.md
+++ b/.changeset/ai-chat-is-tool-continuation.md
@@ -1,0 +1,18 @@
+---
+"@cloudflare/ai-chat": patch
+---
+
+Add `isToolContinuation: boolean` to `useAgentChat()` so consumers can disambiguate a fresh user-initiated `status === "submitted"` from one driven by a server-pushed tool continuation. See [#1365](https://github.com/cloudflare/agents/issues/1365).
+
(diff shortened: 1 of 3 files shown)


threepointone:
Follow-up note on the clear-path fixes, since the last two review comments were really the same class of issue and I ended up doing a deeper audit.

I walked every stateful bit in `useAgentChat()` and split them into three buckets:

1. prop mirrors / render-time refs (`onToolCallRef`, `onDataRef`, `messagesRef`, etc.) — no reset semantics
2. local chat-history state (`messages`, `clientToolResults`, `processedToolCalls`, `protectedStreamingAssistantRef`, tool-continuation state, etc.)
3. broadcast-only stream tracking (`streamStateRef`, `isServerStreaming`)

The two concrete bugs reviewers found were real:
- local `clearHistory()` didn't reset `isToolContinuation`
- server-pushed `CF_AGENT_CHAT_CLEAR` didn't do the equivalent reset either

Fixing those revealed two more asymmetries that weren't observable bugs today but were fragile and relied on the next-render cleanup effects to self-heal:
- `setClientToolResults(new Map())`
- `processedToolCalls.current.clear()`

So I extracted a shared `resetLocalChatState()` helper and now call it from both entry points:
- local `clearHistory()`
- server/other-tab `CF_AGENT_CHAT_CLEAR`

Deliberate non-change: I did **not** move `streamStateRef` / `isServerStreaming` into that shared helper.

Reason: they do *not* mean “this tab’s chat history is empty”; they mean “there is a broadcast/server stream active that this tab is observing.” A local `clearHistory()` cannot meaningfully cancel a stream another tab already started. If I zeroed those locally, the next broadcast chunk would just flip them back on again. So those stay owned by the broadcast handler, where the actual authoritative `CF_AGENT_CHAT_CLEAR` round-trip already resets them.

Net result:
- local clear and broadcast clear now share the full local-history reset surface
- cross-tab/server stream tracking remains broadcast-owned by design
- all 59 react tests still pass

Wanted to leave this rationale here so future review / maintenance has the model spelled out explicitly.

Issue #1359, opened 2026-04-22
subagents: cannot start workflows


kamranayub:
**Describe the bug**

I was trying to build off the experimental/gadgets-subagents and examples/workflows to combine them. I basically wanted a sub agent to kick off and own a workflow.

I had an `AgentWorkflow` named `FlightSearchWorkflow` and a `TripPlannerAgent` parent agent.

I defined a `FlightSearchAgent` and had code like this in a method:

```ts
async runFlightSearch(task: string) {
  console.log("Running flight search workflow...");
  const instanceId = await this.runWorkflow("FLIGHT_SEARCH_WORKFLOW", { task });
  console.log("Started flight search workflow with instanceId:", instanceId);

  // ...
}
```

It never got to the second log, instead it threw an error:

```
Could not detect Agent binding name from class name. Pass it explicitly via options.agentBinding
```

**To Reproduce**

1. Create and export an `AgentWorkflow` that does a step
1. Create an `Agent`
1. Create a subagent with a `doSomething` method that calls `this.runWorkflow` and export it
1. In the parent agent, call `this.subAgent` to get it
1. Invoke the `doSomething` method
1. See error

(Assuming all Wrangler setup is there)

**Expected behavior**

Subagents should be able to a workflow AgentFlow functionality the same way a top-level Agent does.

**Version:**
0.11.4

**Additional context**

```
  "dependencies": {
    "@cloudflare/ai-chat": "*",
    "@cloudflare/kumo": "^1.19.0",
    "@phosphor-icons/react": "^2.1.10",
    "agents": "^0.11.4",
    "ai": "^6.0.168",
    "react": "^19.2.5",
    "react-dom": "^19.2.5",
    "workers-ai-provider": "^3.1.11"
  },
  "devDependencies": {
    "@cloudflare/vite-plugin": "^1.33.1",
    "@cloudflare/workers-types": "^4.20260422.1",
    "@tailwindcss/vite": "^4",
    "@types/node": "^25.6.0",
    "@types/react": "^19.2.14",
    "@types/react-dom": "^19.2.3",
    "@vitejs/plugin-react": "^6.0.1",
    "tailwindcss": "^4.2.2",
    "vite": "^8.0.9"
  }
```

threepointone:
looking into this

threepointone:
work in progress here, but I think this might a while to land (I'm going out of town for the next week, can get to it when I'm back)

threepointone:
[...]

PR #1430, opened 2026-04-30
Fix sub-agent workflow origins

threepointone:
## Summary

Fixes workflows started from sub-agent facets by preserving the exact originating Agent path and routing workflow callbacks/RPC back to that facet instead of assuming a top-level Agent binding.

This addresses the failure mode from cloudflare/agents#1359 where `runWorkflow()` could not detect a binding for sub-agents and gives facet-started workflows the same tracking, callback, and Agent RPC semantics as top-level workflows.

## What changed

- Adds a versioned `AgentWorkflowOrigin` internal payload for workflow starts:
  - top-level Agents use `{ kind: "agent", binding, name }`
  - sub-agent facets use `{ kind: "facet", rootBinding, path }`
- Updates `Agent.runWorkflow()` to preserve root binding + `selfPath` for facet callers.
- Adds path-based internal RPC dispatch via `_cf_invokeAgentPath()` so `AgentWorkflow.agent`, progress, completion, error, event, and state callbacks land on the originating facet.
- Makes facet workflow `AgentWorkflow.agent` explicitly RPC-only and throws a clear error for `.fetch()`.
- Registers sub-agents before facet startup so workflows started in `onStart()` can route callbacks immediately, with rollback if initialization fails.

## Hardening (post-rebase review)

After rebasing onto `main`, the change was reviewed for restart safety, edge cases, and unusual usage. Follow-up fixes:

- **RPC dispatch guard.** `_cf_invokeAgentPath()` now matches real Durable Object stub RPC semantics: it refuses JS-internal probes (`constructor`, `toString`, symbol keys, thenable checks) and anything inherited from `Object.prototype`, so a facet-origin workflow can't reach a method surface a top-level workflow's stub would deny. Error messaging is aligned.
- **Origin version validation.** A mismatched/unknown `AgentWorkflowOrigin.version` now fails with a clear "upgrade the agents package" error instead of silently misreading a future payload shape.
- **Docs.** Documented the callback routing constraints: callbacks resolve the originating Agent **by name** (`idFromName`, not raw DO ids); sub-agent workflow tracking is **facet-local** (`getWorkflows()`/`getWorkflowById()` don't see child runs — aggregate yourself); class names must survive bundling (`keepNames: true`). Fixed a stale `onWorkflowCallback` doc comment that referenced a non-existent `/_workflow/callback` HTTP endpoint.

## Tests

Regression coverage for direct facets, nested facets, `onStart()` workflow starts, approval/rejection, errors, durable events, state updates, and facet-local tracking. New coverage from the hardening pass:

- callbacks route correctly to a facet that was **restarted mid-workflow** (explicitly aborted between start and the approval-resumed completion callback, then re-initialized from durable storage);
- a **deleted-mid-flight** sub-agent surfaces a clear "no longer exists" error, non-descendant paths are rejected, and built-in/prototype method names are refused;
- the documented HTTP escape hatch (`routeSubAgentRequest()` / nested `/sub/...` URL) reaches a facet **while its workflow is running**;
- a workflow started by an older SDK without `__agentOrigin` still resumes through the legacy `__agentName` + `__agentBinding` path.

## Test plan

- `pnpm --filter agents test -- workflow-sub-agent` (11 tests pass; the logged exceptions are intentional thrown/rejected workflow cases)
- `pnpm --filter agents test -- workflow` (86 workflow tests pass)
- `pnpm run typecheck` (all 113 projects)
- `pnpm lint` clean; `oxfmt --check` clean

### DIFF
--- a/.changeset/chat-protected-assistant-terminal-snapshot.md
+++ b/.changeset/chat-protected-assistant-terminal-snapshot.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
+Fix `useAgentChat` reordering a terminal/authoritative `CF_AGENT_CHAT_MESSAGES` snapshot when a protected streaming assistant is followed by a later assistant message (#1778). The protected-tail merge is still applied for stale mid-stream snapshots, but when the incoming snapshot already contains the protected assistant followed by a newer assistant (e.g. a Think HITL denial that persists the denied tool message and then appends a follow-up assistant response), protection is cleared and the snapshot is rendered in its authoritative order instead of moving the protected assistant to the end.
--- a/.changeset/facet-workflow-origins.md
+++ b/.changeset/facet-workflow-origins.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
+Support starting Agent workflows from sub-agent facets by preserving the originating facet path for callbacks and workflow Agent RPC.
--- a/.changeset/facet-workflow-origins.md
+++ b/.changeset/facet-workflow-origins.md
@@ -3,3 +3,9 @@
 ---
 
 Support starting Agent workflows from sub-agent facets by preserving the originating facet path for callbacks and workflow Agent RPC.
+
+- Route workflow callbacks, `this.agent` RPC, progress/completion/error, durable events, and state updates back to the exact originating facet via path-based dispatch.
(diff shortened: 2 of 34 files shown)


mattzcarey:
I'm pretty happy with this. will merge tomorrow

mattzcarey:
can you make not draft @threepointone

PR #1434, opened 2026-04-30
Fix large MCP streamable HTTP payloads

### LINKED ISSUE #1433: MCP streamable-HTTP transport: large JSON-RPC payloads (>16 KB) fail with TLS record_overflow due to header-encoding
## Summary
MCP tool calls with payloads larger than ~16 KB raw, routed through a Cloudflare One MCP Server Portal, fail with `TLS Alert: record_overflow` ~700–900 ms into the request, before the upstream Worker handler runs. Direct calls to the same upstream Worker with the same body succeed.
## Reproduction
1. Configure an MCP Server Portal pointing at any upstream MCP server.
2. Call any tool with content/argument >18 KB raw via the Portal endpoint.
3. Observe: ~700–900 ms then connection terminated with `TLS Alert: level=2 (Fatal), description=22 (record_overflow)`.
The upstream Worker handler is never invoked. Confirmed via `DEBUG_HEADERS` middleware on a receiving Worker — for 18+ KB content, only `initialize` and a small notification arrive; the `tools/call` request never reaches the Worker.
## Threshold (empirical)
| Raw payload size | Result |
| --- | --- |
| 12 KB | ✓ pass |
| 14 KB | ✓ pass |
| 16 KB | ✓ pass |
| 18 KB | ✗ FAIL (`record_overflow`) |
| 20 KB | ✗ FAIL |
| 24 KB | ✗ FAIL |
Direct `curl` to the same upstream endpoint with the same 24 KB body + bearer succeeds (200 OK in 384 ms). Difference is the Portal in the path.
## Localisation
9-row repro matrix at https://github.com/bkalytta-wq/mcp-1433-repro (`EVIDENCE.md`) — same upstream Worker, same Access auth, same `'A' × 24576` payload — eight non-Portal setups pass at 24 KB, including a custom bridge built on `@cloudflare/codemode@0.3.4` against the same upstream. Only the Portal-routed row fails, and it fails at 18 KB.
| # | Setup | 16 KB | 18 KB | 24 KB |
| - | ----- | ----- | ----- | ----- |
| 1 | direct SDK on workers.dev (`0.11.9`, `0.12.3`, `pkg.pr.new/1434`) | ✓ | ✓ | ✓ |
| 4 | direct SDK on custom zone | ✓ | ✓ | ✓ |
| 5 | direct SDK on custom zone + Access (service-token) | ✓ | ✓ | ✓ |
| 7 | direct SDK + Hono + Bearer wrapper (mirrors my upstream) | ✓ | ✓ | ✓ |
| 8 | own codemode bridge (`@cloudflare/codemode@0.3.4`) → row 7 upstream | ✓ | ✓ | ✓ |
| 9 | **Cloudflare One MCP Server Portal → my upstream** | ✓ | ✗ 869 ms | ✗ |

The symptom is localised to the Portal hop. The specific mechanism inside the Portal is not identified — earlier framing of this issue hypothesised an `agents-sdk` header-encoding interaction (`cf-mcp-message` header on the Worker → DO hop). That hypothesis has been corrected per maintainer feedback: the Portal is not built on `@cloudflare/agents`, and there is no `cf-mcp-message` header in the Portal's gateway codebase.
Filing the symptom + matrix evidence here for visibility; root-cause investigation belongs with the Portal team.

threepointone:
## Summary
- Fixes https://github.com/cloudflare/agents/issues/1433.
- Stops encoding streamable HTTP JSON-RPC POST bodies into the internal `cf-mcp-message` upgrade header.
- Keeps the existing Worker-to-Durable Object WebSocket bridge so long-lived streamable HTTP sessions continue to benefit from Durable Object hibernation.
- Sends the POST payload as the first internal WebSocket data frame, then lets the DO process it through the existing `StreamableHTTPServerTransport` path.

## Why
The reported `TLS Alert: record_overflow` failure was caused by large MCP JSON-RPC payloads being base64-encoded into `cf-mcp-message` during the internal Worker-to-Durable Object WebSocket upgrade. That made the request body part of an HTTP header, where it could exceed Cloudflare's per-header limit before the Durable Object ever saw the request.

A direct HTTP forwarding path would avoid the header limit too, but it changes the lifecycle story for streamable HTTP: the current bridge uses WebSockets specifically so idle/long-lived streams can hibernate with the Durable Object runtime. This PR keeps that architecture and only moves the large data out of headers and into the WebSocket data channel, which is the smaller behavioral change for a published package bug fix.

## Details
- `cf-mcp-method` remains as the small internal routing signal for POST vs GET upgrade requests.
- `cf-mcp-message` is no longer populated with the serialized payload and is stripped from forwarded request metadata.
- POST bridge state is recorded on connect, and the first WebSocket message is parsed as the JSON-RPC body.
- Notification/result/error-only POSTs still return `202`; the internal bridge is closed after the DO receives and processes the frame.

## Test plan
- `npm run test:workers -- src/tests/mcp/transports/streamable-http.test.ts`
- `npm run typecheck`
- `npm run check`

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1434" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/fix-mcp-large-streamable-http-payloads.md
+++ b/.changeset/fix-mcp-large-streamable-http-payloads.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
+Fix MCP streamable HTTP transport failures for large JSON-RPC request bodies by sending Worker-to-Durable Object payloads over the internal WebSocket data channel instead of encoding them into headers.
--- a/packages/agents/src/mcp/index.ts
+++ b/packages/agents/src/mcp/index.ts
@@ -8,10 +8,11 @@ import type {
 import {
   JSONRPCMessageSchema,
   isJSONRPCErrorResponse,
+  isJSONRPCRequest,
   isJSONRPCResultResponse,
   type ElicitResult
 } from "@modelcontextprotocol/sdk/types.js";
-import type { Connection, ConnectionContext } from "../";
+import type { Connection, ConnectionContext, WSMessage } from "../";
 import { Agent } from "../index";
 import type { BaseTransportType, MaybePromise, ServeOptions } from "./types";
 import {
@@ -25,6 +26,14 @@ import {
 } from "./utils";
 import { McpSSETransport, StreamableHTTPServerTransport } from "./transport";
 import { RPCServerTransport, type RPCServerTransportOptions } from "./rpc";
+
+type StreamableHttpPostState = {
+  _mcpHttpMethod?: "POST";
+  _mcpRequest?: {
+    headers: Record<string, string>;
+    url: string;
+  };
+};
 
 export abstract class McpAgent<
   Env extends Cloudflare.Env = Cloudflare.Env,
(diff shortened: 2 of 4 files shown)


threepointone:
putting into draft till we get a repro

PR #1456, opened 2026-05-04
Remove default pruneMessages call from Think to preserve client-side tool results

### LINKED ISSUE #1455: Think: pruneMessages is too agressive "before-last-2-messages", breaks client-side tool workflows
Client-side tool outputs (tools without execute — user provides output via addToolOutput) lose context after just 5 messages. The AI has zero memory of what the user chose and starts the flow
from scratch.

Root Cause

```js
_runInferenceLoop hardcodes:
const messages = pruneMessages({
    messages: await convertToModelMessages(truncateOlderMessages(this.session.getHistory()), { tools }),
    toolCalls: "before-last-2-messages"
});
```

Verified in both `@cloudflare/think@0.4.1` and `@0.5.3 (latest)`. There is no subclass hook to configure the pruning options.

With 5 UIMessages → ~7 model messages, all `tool-call` and `tool-result` parts from earlier tools get stripped. For client-side tools, user decisions (e.g., clarification answers, setup config) live only in those tool-result messages — so they're permanently lost.

Why client-side tools are uniquely affected

Server-side tools (with execute) resolve in the same turn — their results are always in the last 2 model messages and survive pruning. Client-side tools span multiple turns by design (LLM
proposes → user provides output → auto-continuation), so their results naturally fall outside the 2-message window on any follow-up message.

Reproduction

1. Define 2+ client-side tools (no execute)
2. User sends a message → AI calls tool 1 → user provides output via addToolOutput → AI auto-continues → calls tool 2 → user provides output → AI auto-continues
3. User sends any follow-up text message
4. AI has no memory of tool 1 or tool 2 outputs

Expected behavior

Tool results from client-side tools should not be silently stripped from a 5-message conversation.

Current workaround

1. The only escape hatch is overriding messages via `TurnConfig` in beforeTurn, but `ctx.messages` is already pruned at that point, so you have to rebuild from `this.messages `
2. In the client, we send the user answers as separate user messages so they stay in the conversation
3. Keep adding to context blocks, but context blocks feel like they're designed for high-signal, long-term facts —  "this project uses Python 3.12. and the user always uses Mac." 

Environment

- @cloudflare/think: 0.4.1 (also verified in 0.5.3)
- ai: 6.0.57

mattzcarey:
## Summary

Fixes #1455.

Think hardcoded `pruneMessages({ toolCalls: "before-last-2-messages" })` in the inference loop, which silently stripped client-side tool results (no `execute`, output supplied via `addToolOutput`) from any turn beyond the second. The user's choices live in those tool results, so the conversation effectively lost its memory.

This PR drops the `pruneMessages` call from `Think._runInferenceLoop` entirely. `truncateOlderMessages` still runs (size-trimming tool outputs >500 chars and text >10k chars in messages older than the last 4), so context cost stays bounded.

## What if you want the old behavior?

Apply `pruneMessages` yourself in `beforeTurn`:

```typescript
import { pruneMessages } from "ai";

beforeTurn(ctx: TurnContext) {
  return {
    messages: pruneMessages({
      messages: ctx.messages,
      toolCalls: "before-last-2-messages"
    })
  };
}
```

Or scope it per-tool with the array form so client-side tool results survive across turns:

```typescript
beforeTurn(ctx: TurnContext) {
  return {
    messages: pruneMessages({
      messages: ctx.messages,
      toolCalls: [
        { type: "before-last-2-messages", tools: ["read_file", "search"] }
      ]
    })
  };
}
```

Both snippets are added to `docs/think/lifecycle-hooks.md` next to the existing `beforeTurn` examples.

## Behavior change

This is a behavior change — the changeset is marked `minor` and explains the diff. Existing agents that benefit from the previous aggressive pruning will see longer model contexts; the size cap from `truncateOlderMessages` keeps the increase bounded.

## Test plan

- [x] `npm run typecheck` — all 80 projects pass
- [x] `npm run check:exports` — all 8 packages valid
- [x] `npx sherif` — no issues
- [x] `oxfmt --check .` — clean
- [x] `oxlint examples/ packages/ guides/ openai-sdk/ site/` — 0 warnings, 0 errors
- [x] New regression test in `packages/think/src/tests/think-session.test.ts` confirms client-side tool outputs (`user-choice-0/1/2`) all reach the model after follow-up turns
- [x] Full think test suite passes (290/290)
- [ ] Manual verification by the issue reporter against their client-side-tool flow

### DIFF
--- a/.changeset/calm-cups-reflect.md
+++ b/.changeset/calm-cups-reflect.md
@@ -1,0 +1,6 @@
+---
+"agents": patch
+"@cloudflare/ai-chat": patch
+---
+
+Avoid throwing when chat stream resume/replay races with a closed WebSocket connection.
--- a/.changeset/configurable-prune-options.md
+++ b/.changeset/configurable-prune-options.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/think": minor
+---
+
+Add `getPruneOptions()` method to `Think`, making the AI SDK's `pruneMessages` step configurable. Default behavior is unchanged (`{ toolCalls: "before-last-2-messages" }`); subclasses can override to relax pruning for client-side tools or return `null` to skip the prune step entirely. Fixes #1455 for agents that rely on multi-turn client-side tool flows where earlier tool results were silently dropped before reaching the model.
--- a/.changeset/configurable-prune-options.md
+++ b/.changeset/configurable-prune-options.md
@@ -2,4 +2,19 @@
 "@cloudflare/think": minor
 ---
 
-Add `getPruneOptions()` method to `Think`, making the AI SDK's `pruneMessages` step configurable. Default behavior is unchanged (`{ toolCalls: "before-last-2-messages" }`); subclasses can override to relax pruning for client-side tools or return `null` to skip the prune step entirely. Fixes #1455 for agents that rely on multi-turn client-side tool flows where earlier tool results were silently dropped before reaching the model.
+Stop applying `pruneMessages({ toolCalls: "before-last-2-messages" })` to the model context by default. The previous default silently stripped client-side tool results (no `execute`, output supplied via `addToolOutput`) from any turn beyond the second, breaking multi-turn flows where the user's choices live in those tool results (see #1455). `truncateOlderMessages` still runs as before, so context cost stays bounded.
+
+This is a behavior change. Subclasses that relied on the old aggressive pruning can opt back in from `beforeTurn`:
+
+```typescript
+import { pruneMessages } from "ai";
+
+beforeTurn(ctx) {
+  return {
+    messages: pruneMessages({
+      messages: ctx.messages,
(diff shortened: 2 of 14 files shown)


threepointone:
Feels like we need a better fix than making this configurable now when we'll likely remove it later?

mattzcarey:
> Feels like we need a better fix than making this configurable now when we'll likely remove it later?

id would prefer to remove it cause it bit me before. @threepointone I cant see it being a good default.

mattzcarey:
Pivoted per your feedback — dropped the default `pruneMessages` call entirely instead of making it configurable. `truncateOlderMessages` still runs so context cost stays bounded, and the `beforeTurn` docs now include opt-in pruning snippets for users who want the old behavior. Force-push avoided; updated PR description and changeset accordingly.

whoiskatrin (approved):

aron-cf (approved):
LGTM

Issue #1352, opened 2026-04-21
voice ⨉ think/ai-chat


threepointone:
make sure tool calling works, vad, etc etc

threepointone:
from internal chat:  

---

  Hey team,
  
  Building an Agent based on Think and VoiceAgent and I'd like the the conversation to fluidly swap between text and voice. 
  
  Here are some things I just bumped into that I'd like to validate with you all:
  
  1. VoiceAgent can't use client side tools.
   
  2. On the client side, useVoiceAgent represents messages as TranscriptMessage[] while useAgentChat does UIMessage<unknown, UIDataTypes, UITools>[] - I'm struggling with displaying a single panel showing the exchange of messages independent of whether it's a text or voice conversations.
  
  3. Is using a Think agent as a sub-agent from within a VoiceAgent via VoiceBase's onTurn and Think's chat combined with a callback which is a StreamCallback a good pattern?
  
  4. What's a good way to insert a "let me look that up" message into the VoiceAgent when a tool needs to be called?

urjitc:
Would love this

threepointone:
I took a pass through the current repo state. My read is: we already have most of the ingredients for this, but not yet the clean flagship proof point this issue is asking for.

What exists today:

- `@cloudflare/voice` already has the real voice pipeline: continuous STT, model turn detection, streaming TTS, text fallback via `sendText()`, interruption/barge-in handling, per-call keepalive, and simple voice conversation persistence.
- `@cloudflare/ai-chat` already has the mature chat path: persisted AI SDK `UIMessage`s, server tools, client tools, `needsApproval`, approval continuation, stream recovery, MCP/browser tools, and the React `useAgentChat` state machine.
- `examples/playground/src/demos/voice/voice-agent.ts` proves the `withVoice(Agent)` path with Workers AI + AI SDK `result.fullStream` + spoken output.
- `examples/ai-chat` proves tools, approvals, MCP/browser tools, recovery, and complete chat UI rendering.
- `examples/elevenlabs-starter` is the closest existing hybrid: `VoiceChatAgent extends AIChatAgent`, audio is transcribed via ElevenLabs, transcript is fed into the normal chat UI, and assistant messages are spoken afterwards via callable TTS. That proves a voice-flavored AI chat UX, but it is not the `@cloudflare/voice` pipeline and does not prove `withVoice` + AI chat protocol integration.

The important gap is that voice and AI chat currently own different turn protocols and different persistence shapes:

[...]

PR #1462, opened 2026-05-05
Fix voice TTS for AI SDK textStream responses

### LINKED ISSUE #1364: `withVoice`: `onTurn` returning AI SDK `result.textStream` produces no TTS audio
## Summary

When `onTurn` returns the `textStream` from AI SDK v6's `streamText()`, the voice pipeline runs the LLM and saves the assistant message but produces **no TTS audio**. Wrapping the same stream in a plain `async function*` makes audio play correctly.

## Environment

- `@cloudflare/voice@0.1.2`
- `agents@0.11.4`
- `@cloudflare/ai-chat@0.4.6`
- `ai@6.0.168`
- `workers-ai-provider@3.1.11`
- Workers AI providers: `WorkersAIFluxSTT` (STT) + `WorkersAITTS` (TTS, `@cf/deepgram/aura-1`)
- Local dev via `wrangler dev` on macOS, Chrome

## Repro

`withVoice(AIChatAgent)`-mixed agent with this `onTurn`:

```ts
async onTurn(transcript: string) {
  const workersai = createWorkersAI({ binding: this.env.AI });
  const result = streamText({
    model: workersai("@cf/google/gemma-4-26b-a4b-it"),
    system: "You are a helpful voice assistant. Keep replies short.",
    messages: [{ role: "user", content: transcript }]
  });
  return result.textStream; // ← no audio plays
}
```

Observed:
- STT works (interim transcript shows correctly).
- LLM runs (text appears in the conversation history, `saveMessage("assistant", …)` fires).
- **No TTS request is made** — Aura is never called, no audio frames sent over the WebSocket.
- No errors logged on server or client.

## Workaround

Wrap the stream so it presents as a pure `AsyncIterable<string>`:

```ts
async function* toStrings() {
  for await (const chunk of result.textStream) yield chunk;
}
return toStrings(); // ← audio plays correctly
```

## Suspected cause

In `dist/voice.js` `iterateText` checks `instanceof ReadableStream` before `Symbol.asyncIterator`:

```js
if (source instanceof ReadableStream) {
  const reader = source.getReader();
  const first = await reader.read();
  if (first.done || first.value === void 0) return;
  if (typeof first.value === "string") { /* string branch */ }
  else { /* NDJSON branch */ }
}
if (Symbol.asyncIterator in source) { /* string branch — works */ }
```

AI SDK v6's `result.textStream` satisfies both protocols. Going through `getReader()` either reads a non-string first value (sending it down the NDJSON branch, which then yields nothing for raw string chunks) or returns `done: true` before the LLM has produced a delta — in either case, no text reaches `SentenceChunker` and no TTS happens. Iterating via `Symbol.asyncIterator` works because the AI SDK's async iterator yields raw string deltas.

## Suggested fix

Reorder the checks in `iterateText` so `Symbol.asyncIterator` is tried before `instanceof ReadableStream`, **or** when the first peeked value isn't a string, also check whether the source is async-iterable and prefer that path before falling back to NDJSON parsing.

A documentation note in the README that `streamText({...}).textStream` should be wrapped (or that `toTextStreamResponse()` etc. is preferred) would also help, since the AI SDK is the obvious LLM choice for a Workers-based agent.

## Related

The dual `ReadableStream<string>` + `AsyncIterable<string>` shape is standard for AI SDK v5+. This bug will likely affect anyone wiring the voice mixin to AI SDK out-of-the-box.

whoiskatrin:
## Summary

Fixes `withVoice` handling of AI SDK `streamText(...).textStream` values returned from `onTurn()`.

AI SDK `textStream` values are dual-protocol objects: they are `ReadableStream`s and also expose a custom `Symbol.asyncIterator` that yields string deltas. `iterateText()` previously checked `instanceof ReadableStream` first, so these streams could be consumed through the generic stream parser instead of the AI SDK string iterator. That meant assistant text could be persisted while no text reached the sentence chunker/TTS pipeline, producing no spoken audio.

This updates `iterateText()` to prefer a custom/source-owned async iterator before falling back to `ReadableStream` parsing, while preserving the existing native `ReadableStream<string>` and `ReadableStream<Uint8Array>` SSE/NDJSON paths.

Fixes #1364
Refs #1352

## Changes

- Prefer custom async iterators for dual-protocol text streams in `@cloudflare/voice`.
- Keep native `ReadableStream` parsing behavior intact for string streams and SSE/NDJSON byte streams.
- Add a regression test for an AI SDK-like dual-protocol stream.
- Document that AI SDK `result.textStream` can be returned directly from `onTurn()`.
- Add a patch changeset for `@cloudflare/voice`.

## Testing

- `npm test -w @cloudflare/voice`
- `npm run build`
- `npm run check`


mattzcarey (approved):
looks good and makes sense to me

PR #1470, opened 2026-05-06
fix(codemode): resolve OpenAPI specs inside sandbox

### LINKED ISSUE #1466: resolveRefs in openApiMcpServer overloads Worker Loader RPC boundary limit for deeply-nested API surfaces
**Describe the bug**

`openApiMcpServer` (in `@cloudflare/codemode/mcp`) calls `resolveRefs(spec, spec)` at startup (`packages/codemode/src/mcp.ts`, around the `openApiMcpServer` definition; in `dist/mcp.js@0.3.4` the call site is line 235, the resolver is at line 135). It walks the spec and inlines every `#/`-prefixed `$ref` recursively at every site. The resolved object is then exposed to the sandboxed sub-worker via `codemode.spec()` on each `search` call, and that value crosses the Worker Loader RPC boundary, which is capped at 32 MiB per argument or return value.

For sufficiently rich enterprise specs this is unworkable. Concrete numbers from a real provider integration (Xero accounting + payroll AU OpenAPI):

| Variant | On-disk | After `resolveRefs` (JSON bytes) | Runtime serialized (observed) |
|---|---|---|---|
| Xero (full)        |   778 KiB | 71.04 MiB | 141.9 MiB (`148,779,506 bytes`) |
| Gmail (control)    |   129 KiB |   250 KiB | (under cap, fine) |

At runtime the `search` tool fails with:

```
Error: Serialized RPC arguments or return values are limited to 32MiB,
but the size of this value was: 148779506 bytes.
```

**To Reproduce**

1. Instantiate `openApiMcpServer` with a sufficiently large OpenAPI spec; specs with deep cross-referencing `$ref` graphs blow up most. The Xero accounting + payroll AU OpenAPI (~778 KiB on disk) is the example where I encountered the error.
2. Connect an MCP client and invoke the `search` tool.
3. The call fails with the RPC size error shown above.

[measure-spec-resolution.mjs.txt](https://github.com/user-attachments/files/27420395/measure-spec-resolution.mjs.txt)

I've attached a reproducer script (`measure-spec-resolution.mjs`) that copies the `resolveRefs` function verbatim from `dist/mcp.js@0.3.4` and runs it against bundled specs, so the resolved-size blowup can be confirmed without deploying a worker.

**Expected behavior**

`search` should work with reasonably-sized real-world OpenAPI specs without the resolved-spec payload exceeding the 32 MiB Worker Loader RPC argument/return limit.

**Screenshots**

N/A -- the failure is a runtime error string, included verbatim above.

**Version:**

`@cloudflare/codemode@0.3.4`

**Additional context**

The attached script also tries the obvious mitigations:

| Variant | After `resolveRefs` (JSON bytes) | Under 32 MiB JSON? |
|---|---|---|
| Xero pruned to ~60% of operations               | 46.86 MiB | no |
| Xero full with `description`/`examples` stripped | 30.98 MiB | yes (1 MiB margin) |
| Xero pruned + descriptions stripped              | 20.47 MiB | yes (11.5 MiB margin) |

JSON-byte size is a lower bound on what the runtime serializes -- the observed runtime size is roughly 2x the JSON-byte count for the same input (71 MiB JSON -> 142 MiB runtime). Applying that lens, even the most aggressive of the three mitigations (prune + strip) is in the danger zone at ~40 MiB.

A natural fix would be to pass the unresolved spec across the RPC boundary and expose a `codemode.resolveRef(refString)` helper to the sandbox so search code can follow refs on demand. The `SPEC_TYPES` doc string would need updating to reflect the new contract (currently it promises "All $refs are pre-resolved inline").

Related but distinct: #1268 proposes operation-level filtering for `openApiMcpServer`. That's complementary -- it reduces *what* is exposed; this issue is about *how* the exposed surface is transmitted. Even with #1268's filter, large remaining schemas would still blow up due to ref expansion (as shown above with Xero).

mattzcarey:
## Summary

Fixes #1466.

`openApiMcpServer` previously resolved every internal OpenAPI `$ref` on the host during setup:

```ts
const resolved = resolveRefs(options.spec, options.spec);
```

That preserved a convenient LLM-facing contract — `await codemode.spec()` returned a fully inlined spec — but it also meant the fully expanded object had to cross the Worker Loader RPC boundary whenever sandbox code called `codemode.spec()`.

For heavily-referenced real-world specs, resolving refs inline can expand a compact OpenAPI document by orders of magnitude. In the issue report, a ~778 KiB Xero spec expanded to ~71 MiB JSON / ~142 MiB runtime serialization, which exceeds the Worker Loader 32 MiB argument/return value limit before the generated search code can do anything useful.

This PR keeps the simple LLM-facing API but changes where the work happens:

- embed the compact unresolved OpenAPI spec into the sandbox wrapper
- resolve `$ref`s inside the sandbox when `codemode.spec()` is called
- keep `codemode.spec()` returning a fully resolved spec, preserving existing generated-code behavior
- keep `codemode.request()` as the only host-dispatched RPC call, since API requests need host-side auth/network access
- stringify/truncate search and execute results inside the sandbox before they cross back over RPC

This hides `$ref` complexity from the LLM while avoiding the huge pre-expanded spec crossing the host → sandbox RPC boundary.

## Why this shape

We considered exposing an explicit `codemode.resolveRef(ref)` helper and returning an unresolved spec from `codemode.spec()`. That would avoid the oversized RPC payload too, but it pushes OpenAPI reference-management complexity into the generated code. Since LLM-generated code should be able to one-shot common search tasks with a simple mental model, this PR preserves the existing `codemode.spec()` behavior instead.

The remaining RPC boundary is now reserved for genuinely host-owned behavior:

```ts
await codemode.request(...)
```

Static OpenAPI data and `$ref` expansion stay local to the sandbox execution.

## Tests

Added a regression test that builds a synthetic OpenAPI spec with many operations referencing one large shared schema. Under the old eager host-side resolution this expands beyond the 32 MiB Worker Loader RPC limit. With this change, `codemode.spec()` resolves inside the sandbox and the search completes successfully.

Verified locally:

```bash
npx nx run @cloudflare/codemode:test -- mcp.test.ts -u
npm run build
npm run check
```

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1470" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/codemode-openapi-spec-sandbox.md
+++ b/.changeset/codemode-openapi-spec-sandbox.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/codemode": patch
+---
+
+Resolve OpenAPI specs inside the codemode sandbox to avoid Worker Loader RPC size limits for heavily-referenced specs.
--- a/packages/codemode/src/mcp.ts
+++ b/packages/codemode/src/mcp.ts
@@ -6,6 +6,7 @@ import {
   generateTypesFromJsonSchema,
   type JsonSchemaToolDescriptors
 } from "./json-schema-types";
+import { normalizeCode } from "./normalize";
 import { sanitizeToolName } from "./utils";
 import type { Executor } from "./executor";
 
@@ -247,48 +248,6 @@ export interface OpenApiMcpServerOptions {
   description?: string;
 }
 
-/**
- * Resolve internal $ref pointers in a JSON object against the root document.
- * Only handles `#/` internal refs. External file refs are left as-is.
- */
-function resolveRefs(
-  obj: unknown,
-  root: Record<string, unknown>,
-  seen = new Set<string>()
-): unknown {
-  if (obj === null || obj === undefined) return obj;
-  if (typeof obj !== "object") return obj;
-  if (Array.isArray(obj))
-    return obj.map((item) => resolveRefs(item, root, seen));
-
-  const record = obj as Record<string, unknown>;
-
-  if ("$ref" in record && typeof record.$ref === "string") {
-    const ref = record.$ref;
-    if (seen.has(ref)) return { $circular: ref };
-    if (!ref.startsWith("#/")) return record;
-    seen.add(ref);
-
-    const parts = ref
-      .slice(2)
-      .split("/")
-      .map((s) => s.replace(/~1/g, "/").replace(/~0/g, "~"));
-    let resolved: unknown = root;
-    for (const part of parts) {
-      resolved = (resolved as Record<string, unknown>)?.[part];
-    }
-    const result = resolveRefs(resolved, root, seen);
-    seen.delete(ref);
-    return result;
-  }
-
-  const result: Record<string, unknown> = {};
-  for (const [key, value] of Object.entries(record)) {
-    result[key] = resolveRefs(value, root, seen);
-  }
-  return result;
-}
-
 const SPEC_TYPES = `
 // OpenAPI 3.x spec with $refs resolved inline.
 // The spec object follows the standard OpenAPI 3.x structure.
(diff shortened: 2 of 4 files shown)


aron-cf (approved):
👏 Nice fix!

PR #1476, opened 2026-05-07
Fix agent-tool recovery finalization

### LINKED ISSUE #1475: _reconcileAgentToolRuns updates SQL terminal status but does not invoke onAgentToolFinish or broadcast — causes silent state desync after DO eviction
# Cloudflare Agents SDK — Bug Report

**Title:** `_reconcileAgentToolRuns` updates `cf_agent_tool_runs` to terminal status but does not invoke `onAgentToolFinish` hook or call `_broadcastAgentToolTerminal`, causing application state to silently desync from framework state after any DO eviction or hibernation.

**Severity:** High. Affects any application that mirrors agent-tool run state (database tables, structured logs, message-part state, dashboards) — which is the standard pattern from the SDK's own examples. Symptom is silent and progressive: state drifts further from reality with each parent DO eviction, eventually causing the chat UI to show forever-spinning tool calls and orchestrators to repeat-dispatch sub-agents that already completed.

**SDK version:** `agents@0.12.3` (also via `@cloudflare/ai-chat@0.6.2` and `@cloudflare/think@0.5.3`). Confirmed against the source at `node_modules/agents/dist/index.js` lines 3710–3743 in `Agent._reconcileAgentToolRuns`.

---

### Setup

A standard Cloudflare Agents app with a parent agent class (`Agent` extension) that uses `agentTool(ChildAgent, {...})` to dispatch a sub-agent. The application:

1. Mirrors `cf_agent_tool_runs` rows into a separate table (e.g. for a sidebar UI showing all runs across chats) by overriding `onAgentToolStart` and `onAgentToolFinish` to write to the mirror.
2. Maintains structured per-step logging via `onStepFinish` for observability.
3. Renders the parent's `tool-X` message-part state in a chat UI; relies on `_broadcastAgentToolTerminal` frames to flip `tool-X` from `input-available` to `output-available`.

Concrete repro pattern from spk-agents (single-tenant chat app on Workers AI):

```ts
class MyAssistant extends Agent<Env> {
  override async onAgentToolStart(run: AgentToolRun) {
    void this.parentAgent(AssistantDirectory).recordSubSession({
      runId: run.runId, agentType: run.agentType, status: "running", ...
    });
  }
  override async onAgentToolFinish(run: AgentToolRun, result: AgentToolResult) {
    void this.parentAgent(AssistantDirectory).recordSubSession({
      runId: run.runId, status: result.status, ...
    });
  }
}
```

### Steps

1. Trigger a long-running sub-agent dispatch (e.g. `runAgentTool(ChildAgent, ...)` with input that takes >30s of model + tool work).
2. Force-evict or hibernate the parent `MyAssistant` DO before the child completes. Practical ways to trigger:
   - Run `wrangler deploy` while the child run is in flight (DO is reset on script update).
   - Wait for natural DO hibernation if the parent has no active client connection.
3. The child run continues to completion in its own DO and writes terminal status into the parent's `cf_agent_tool_runs` table (via the SDK's child→parent state propagation).
4. The parent DO wakes up on the next request. `Agent.onStart` invokes `_reconcileAgentToolRuns`.

### Expected

- `_reconcileAgentToolRuns` detects the terminal child status, updates the framework's `cf_agent_tool_runs` row, **and** invokes `onAgentToolFinish(run, result)` so application code can mirror the transition.
- A `_broadcastAgentToolTerminal` frame is emitted so any subscribed clients (chat UI, message-part state machine) see the tool-part flip to `output-available`.
- The mirror table and the framework table are in sync.

### Actual

`_reconcileAgentToolRuns` at `node_modules/agents/dist/index.js:3710`:

```js
async _reconcileAgentToolRuns() {
  const rows = this.sql`SELECT * FROM cf_agent_tool_runs WHERE status IN ('starting', 'running')`;
  for (const row of rows) {
    const child = await this.subAgent(this._agentToolClassByName(row.agent_type), row.run_id);
    const inspection = await child.inspectAgentToolRun(row.run_id);
    if (!inspection || inspection.status === 'running' || inspection.status === 'starting') {
      this.sql`UPDATE cf_agent_tool_runs SET status = 'interrupted', summary = '...' WHERE run_id = ${row.run_id}`;
      // ❌ no this.onAgentToolFinish(...) call
      // ❌ no this._broadcastAgentToolTerminal(...) call
    } else {
      this.sql`UPDATE cf_agent_tool_runs SET status = ${inspection.status}, summary = ${inspection.summary} WHERE run_id = ${row.run_id}`;
      // ❌ no this.onAgentToolFinish(...) call
      // ❌ no this._broadcastAgentToolTerminal(...) call
    }
  }
}
```

Compare to `Agent.runAgentTool` (around `index.js:3367`) which DOES call both:

```js
// inside runAgentTool, after a child reaches terminal status:
await this._updateAgentToolTerminal(run, result);
this._broadcastAgentToolTerminal(run, result);
await this.onAgentToolFinish(run, result);  // ← the missing piece in _reconcileAgentToolRuns
```

### Observable consequences in the affected application

1. The application's mirror table (`sub_sessions` in our case) shows `status: 'running'` indefinitely for a child that the framework already marked terminal. Sidebar UI, dashboards, watchdog logic that depend on this all desync.
2. The chat UI's `tool-X` message-part stays in its pre-terminal state (`input-streaming` or `input-available`), making the call appear to spin forever to the user.
3. Application-level orchestration (e.g. an LLM that decides whether to re-dispatch based on whether the previous call "completed") may repeat-dispatch the same sub-agent. We've observed 11 stacked dispatches of the same agent-tool from a single user message because each new invocation reads the previous as still in flight.

---

## Application-side workarounds we deployed

In case it helps another team running into this:

1. **Heartbeat + watchdog.** Override `onStepFinish` to bump a `last_activity_at` column on our mirror table via cross-DO RPC. A periodic alarm-driven sweep marks any row silent for 10+ min as `interrupted` and rewrites the parent's stuck `tool-X` part to `output-error` via `session.updateMessage`. This bypasses the framework's reconciliation entirely. ~150 lines of code, plus an extra DO RPC per step.
2. **`onChatRecovery` repair.** Override the recovery hook to scan the last assistant message for `tool-X` parts whose framework row is now terminal and rewrite them in place before recovery resumes the turn.
3. **Auto-investigation.** When the watchdog force-interrupts, fire-and-forget an Analyst agent with the run context to investigate cause. Without this, every desync looked like a fresh model failure.

These workarounds eliminated the user-visible symptoms but added significant complexity and observability surface. A SDK-level fix would let us delete most of it.

---

## Suggested fix

In `Agent._reconcileAgentToolRuns`, after each terminal-status update, reconstruct an `AgentToolRun` and `AgentToolResult` from the row and call:

```js
await this.onAgentToolFinish(run, result);
this._broadcastAgentToolTerminal(run, result);
```

Mirroring what `runAgentTool` does at the synchronous-completion path (line 3367). This is a ~5-line change.

Edge case: `onAgentToolFinish` is documented as part of the lifecycle hook set; calling it from reconciliation matches user expectations.

---

## Repro repo

If helpful: https://github.com/smartphonekey/spk-agents — the relevant code paths are in `src/server.ts` (search for `_sweepStaleSubSessions`, `recordSubSession`, `onAgentToolStart`/`onAgentToolFinish` overrides). The watchdog workaround is at `AssistantDirectory._sweepStaleSubSessions`.

Happy to provide a tighter minimal repro if useful.

whoiskatrin:
## Summary

Fixes #1475.

This PR makes recovered agent-tool runs go through the same terminal finalization path as live `runAgentTool()` executions.

Previously, `_reconcileAgentToolRuns()` updated `cf_agent_tool_runs` when a parent Agent woke up and found a child run had reached a terminal state, but it skipped two observable parts of the framework lifecycle:

- `onAgentToolFinish(...)`
- terminal `agent-tool-event` broadcasts via `_broadcastAgentToolTerminal(...)`

That meant application state mirrors, dashboards, structured logs, and chat UI state could silently drift from the framework’s durable run registry after parent DO eviction / hibernation.

### Shared terminal finalization path

Added a private `_finishAgentToolRun(...)` helper that centralizes the terminal lifecycle invariant:

1. update `cf_agent_tool_runs`
2. broadcast terminal agent-tool event when a sequence is available
3. invoke `onAgentToolFinish(...)`

Live `runAgentTool()` terminal paths now use this helper too, so recovery and live execution share the same behavior.

### Reconciliation now restores observable lifecycle

`_reconcileAgentToolRuns()` now:

- selects the full stored agent-tool row, not only `run_id` / `agent_type`
- reconstructs `AgentToolRunInfo` from durable state
- resolves the child by stored `agent_type`
- inspects the child’s terminal state
- best-effort replays stored child chunks before terminal broadcast
- finalizes the run through `_finishAgentToolRun(...)`

### Stored chunk replay helper

Added `_broadcastAgentToolStoredChunks(...)` to keep chunk replay behavior local and shared between:

- recovery reconciliation
- replay to connecting clients
- interrupted replay path

This keeps sequencing behavior consistent: terminal events are emitted after any replayed stored chunks.

## Why this architecture

The bug came from terminal finalization being spread across call sites. `runAgentTool()` knew to update durable state, broadcast, and call the lifecycle hook, while `_reconcileAgentToolRuns()` only updated durable state.

This PR deepens the internal finalization module: callers provide a run and terminal result, and the implementation owns the lifecycle ordering. That improves locality and makes it less likely future recovery paths drift from live execution.

## Tests

Added regression coverage for parent recovery reconciliation using `AIChatAgentToolParent` / `AIChatAgentToolChild`.

The test simulates:

1. a parent registry row stuck in `running`
2. a child agent-tool run that has completed durably
3. parent recovery reconciliation
4. verification that:
   - `onAgentToolFinish(...)` fires
   - the terminal `finished` event is broadcast
   - reconstructed run metadata is passed to the lifecycle hook

## Verification

Ran:

```bash
cd packages/ai-chat && npm run test:workers -- src/tests/agent-tools.test.ts
npm run check
```

### DIFF
--- a/.changeset/fix-agent-tool-reconcile-finish.md
+++ b/.changeset/fix-agent-tool-reconcile-finish.md
@@ -2,4 +2,4 @@
 "agents": patch
 ---
 
-Ensure recovered agent-tool runs go through the same terminal lifecycle path as live runs. Parent recovery reconciliation now replays stored child chunks, broadcasts terminal agent-tool events, and invokes `onAgentToolFinish` after updating the parent run registry.
+Fixed a bug that could cause client state to drift from internal Durable Object state when agent tool calls spanned a Durable Object restart.
--- a/.changeset/fix-agent-tool-reconcile-finish.md
+++ b/.changeset/fix-agent-tool-reconcile-finish.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
+Ensure recovered agent-tool runs go through the same terminal lifecycle path as live runs. Parent recovery reconciliation now replays stored child chunks, broadcasts terminal agent-tool events, and invokes `onAgentToolFinish` after updating the parent run registry.
--- a/packages/agents/src/index.ts
+++ b/packages/agents/src/index.ts
@@ -5740,7 +5740,6 @@ export class Agent<
         return this._resultFromAgentToolRow<Output>(existing);
       }
       return await this._replayAndInterruptAgentToolRun<Output>(
-        cls,
         existing,
         "Agent tool run was still running, but live-tail reattachment is not supported in this runtime."
       );
@@ -5836,16 +5835,7 @@ export class Agent<
           status: "aborted",
           error: reason
         };
-        this._updateAgentToolTerminal(runId, result);
-        this._broadcastAgentToolTerminal(
-          options.parentToolCallId,
-          sequence,
-          result
-        );
-        await this.onAgentToolFinish(
-          { ...runInfo, status: "aborted", completedAt: Date.now() },
-          result
-        );
+        await this._finishAgentToolRun(runInfo, result, { sequence });
         return result;
       } else {
         parentAbortListener = () => {
@@ -5891,16 +5881,7 @@ export class Agent<
           status: "aborted",
           error: reason
         };
-        this._updateAgentToolTerminal(runId, result);
-        this._broadcastAgentToolTerminal(
-          options.parentToolCallId,
-          sequence,
-          result
-        );
-        await this.onAgentToolFinish(
-          { ...runInfo, status: "aborted", completedAt: Date.now() },
-          result
-        );
+        await this._finishAgentToolRun(runInfo, result, { sequence });
         return result;
       }
 
@@ -5910,16 +5891,10 @@ export class Agent<
         agentType,
         inspection
       );
-      this._updateAgentToolTerminal(runId, result, inspection.completedAt);
-      this._broadcastAgentToolTerminal(
-        options.parentToolCallId,
+      await this._finishAgentToolRun(runInfo, result, {
         sequence,
-        result
-      );
-      await this.onAgentToolFinish(
-        { ...runInfo, status: result.status, completedAt: Date.now() },
-        result
-      );
+        completedAt: inspection.completedAt
+      });
       return result;
     } catch (error) {
       if (options.signal?.aborted) {
@@ -5934,16 +5909,7 @@ export class Agent<
           status: "aborted",
           error: reason
         };
-        this._updateAgentToolTerminal(runId, result);
-        this._broadcastAgentToolTerminal(
-          options.parentToolCallId,
-          sequence,
-          result
-        );
-        await this.onAgentToolFinish(
-          { ...runInfo, status: "aborted", completedAt: Date.now() },
-          result
-        );
+        await this._finishAgentToolRun(runInfo, result, { sequence });
         return result;
       }
       const message = error instanceof Error ? error.message : String(error);
@@ -5953,16 +5919,7 @@ export class Agent<
         status: "error",
         error: message
       };
-      this._updateAgentToolTerminal(runId, result);
-      this._broadcastAgentToolTerminal(
-        options.parentToolCallId,
-        sequence,
-        result
-      );
-      await this.onAgentToolFinish(
-        { ...runInfo, status: "error", completedAt: Date.now() },
-        result
-      );
+      await this._finishAgentToolRun(runInfo, result, { sequence });
       return result;
     } finally {
       if (parentAbortListener && options.signal) {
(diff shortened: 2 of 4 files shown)


.changeset/fix-agent-tool-reconcile-finish.md
```diff
@@ -0,0 +1,5 @@
+---
+"agents": patch
+---
+
+Ensure recovered agent-tool runs go through the same terminal lifecycle path as live runs. Parent recovery reconciliation now replays stored child chunks, broadcasts terminal agent-tool events, and invokes `onAgentToolFinish` after updating the parent run registry.
```
aron-cf: This doesn't clearly describe to the reader what actually changed from their perspective. I think it's something like "Fixed bug causing client state to drift from internal Durable Object state when tool calls spanned a Durable Object restart."

aron-cf (approved):
One comment regarding changeset.

.changeset/fix-agent-tool-reconcile-finish.md
whoiskatrin: Good point, fixed

PR #1478, opened 2026-05-07
Add enabled gate to useVoiceAgent

### LINKED ISSUE #1474: Add enabled option to @cloudflare/voice useVoiceAgent
## Summary

`useVoiceAgent` currently connects as soon as the hook mounts and does not expose an `enabled` option. Apps that need to wait for an async, user/session-scoped credential before opening the WebSocket have to mount the hook with a dormant local transport first, then swap to the real connection options after auth is ready.

## Use case

In an anonymous marketing-demo voice lane, the browser must first obtain a short-lived capability token after a Turnstile solve. The token is bound to `{ sessionToken, caseStudyId }` and is required on the `useVoiceAgent` WebSocket query. Opening the WebSocket before that token exists should not happen: the server correctly rejects the connection, and the UI would show avoidable transport churn before the user can start a call.

The current local workaround keeps the official hook mounted, but passes a no-op/dormant transport until the capability bootstrap resolves. That preserves hook ordering, but it is extra adapter code solely to represent "do not connect yet."

## Current API evidence

`@cloudflare/voice@0.1.2` has:

```ts
export interface UseVoiceAgentOptions extends VoiceClientOptions {
  onReconnect?: () => void;
}
```

The hook's effect always constructs `new VoiceClient(options)` and calls `client.connect()` when the connection key changes. There is no `enabled?: boolean` gate in the options type.

## Requested API

Add an optional gate such as:

```ts
export interface UseVoiceAgentOptions extends VoiceClientOptions {
  enabled?: boolean;
  onReconnect?: () => void;
}
```

Expected behavior:

- default `enabled` to `true` for backward compatibility
- when `enabled === false`, do not construct/connect a `VoiceClient`
- reset exposed state to the idle/disconnected shape while disabled
- when `enabled` flips `false -> true`, construct/connect using the current options
- when `enabled` flips `true -> false`, disconnect/cleanup the current client
- include `enabled` in the connection identity/effect dependencies without firing `onReconnect` for the initial disabled mount

## Why this belongs upstream

Credential-gated voice connections are a common app shape, and the cleanest representation is a hook-level connection gate. Without it, every app with async auth/capability/bootstrap needs its own dormant `VoiceTransport` adapter, which is easy to get subtly wrong and obscures the intended lifecycle.

whoiskatrin:
## Summary

Fixes #1474 by adding an `enabled` lifecycle gate to `useVoiceAgent`.

Apps that need to wait for async connection prerequisites, such as a Turnstile-backed capability token, can now keep the hook mounted without constructing or connecting a `VoiceClient` until those prerequisites are ready.

## What changed

- Adds `enabled?: boolean` to `UseVoiceAgentOptions`, defaulting to `true` for backwards compatibility.
- Keeps `useVoiceAgent` in the idle/disconnected state while disabled.
- Avoids constructing or connecting `VoiceClient` when `enabled === false`.
- Connects with the current options when `enabled` flips from `false` to `true`.
- Disconnects and resets state when `enabled` flips from `true` to `false`.
- Makes hook action callbacks safe no-ops while disabled.
- Ensures the initial disabled mount and first enable do not fire `onReconnect`.
- Documents the new option and records the lifecycle decision in the voice design notes.
- Adds a changeset for `@cloudflare/voice`.

## Architecture note

The gate lives at the React hook seam rather than in `VoiceClient` or a new transport adapter. This keeps auth/bootstrap readiness local to the React lifecycle module, and avoids forcing users to encode “do not connect yet” with dormant `VoiceTransport` adapters.

## Tests

- `npm --workspace @cloudflare/voice run build`
- `npm --workspace @cloudflare/voice run test:react`
- `npm run check`

### DIFF
--- a/.changeset/bright-voices-wait.md
+++ b/.changeset/bright-voices-wait.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/voice": minor
+---
+
+Add an `enabled` option to `useVoiceAgent` so React apps can delay creating and connecting a `VoiceClient` until async prerequisites such as capability tokens are ready.
--- a/design/voice.md
+++ b/design/voice.md
@@ -387,6 +387,8 @@ JSON message handling narrows the `try/catch` to only cover `JSON.parse`. Listen
 - `useVoiceAgent(options)` — wraps `VoiceClient` for `withVoice` agents
 - `useVoiceInput(options)` — wraps `VoiceClient` for `withVoiceInput` agents
 
+`useVoiceAgent` accepts `enabled?: boolean` as a React lifecycle gate. While disabled, the hook keeps the exposed state in the idle/disconnected shape and does not construct or connect a `VoiceClient`. Enabling the hook creates a client with the current options; disabling it tears down the active client. This keeps credential/bootstrap readiness in the React hook module instead of forcing callers to provide dormant `VoiceTransport` adapters.
+
 Both include tuning knobs (`silenceThreshold`, `silenceDurationMs`, `interruptThreshold`, `interruptChunks`) in the connection key so changing them triggers client recreation.
 
 ## Tradeoffs
--- a/docs/voice.md
+++ b/docs/voice.md
@@ -335,9 +335,12 @@ const {
 } = useVoiceAgent({
   agent: "MyAgent", // Required: Durable Object class name
   name: "default", // Instance name (default: "default")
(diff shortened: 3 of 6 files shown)


sort2f:
Thanks, this is exactly the lifecycle gate I need. Hope it gets merged/released soon!

aron-cf (approved):
Looks good 👍

PR #1484, opened 2026-05-08
Add durable cancellation mode for useAgentChat

### LINKED ISSUE #1471: useAgentChat should not transparently cancel durable turns on client stream disposal
## Summary

`useAgentChat` currently maps client-side stream abort/cancel lifecycle to server-side turn cancellation. For Durable Object backed agents, that is the wrong default contract: browser tabs are flaky observers of a durable turn, and page lifecycle / React cleanup / AI SDK stream disposal should be recoverable through stream resumption unless the application explicitly chooses to stop the server turn.

To be precise: this issue is **not** claiming that a raw WebSocket close directly calls `onAbort`. A plain disconnect should be resume-only. The problematic path is that `useChat` owns an active client response with an `AbortController`, and the Cloudflare `WebSocketChatTransport` maps that abort signal and `ReadableStream.cancel()` to the same `CF_AGENT_CHAT_REQUEST_CANCEL` frame used for explicit stop.

Server-side cancellation should be an explicit user/app action, e.g. `stop()`, or transparent cancellation on client abort should be opt-in via configuration.

This is not about removing explicit stop. Explicit stop is useful. The bug is that the current implementation does not give apps a way to distinguish explicit user stop from generic client stream abort/cancel lifecycle.

## Relevant code

Pinned to current `main` at `5f6214dccfe3ba9bf243ec15d291b37b9659a54c`.

`WebSocketChatTransport.sendMessages()` defines a single `onAbort` path that sends `CF_AGENT_CHAT_REQUEST_CANCEL` to the server:

https://github.com/cloudflare/agents/blob/5f6214dccfe3ba9bf243ec15d291b37b9659a54c/packages/ai-chat/src/ws-chat-transport.ts#L197-L214

That same `onAbort` path is used for both `ReadableStream.cancel()` and the caller's `options.abortSignal`:

https://github.com/cloudflare/agents/blob/5f6214dccfe3ba9bf243ec15d291b37b9659a54c/packages/ai-chat/src/ws-chat-transport.ts#L263-L271

`useAgentChat` constructs `WebSocketChatTransport` internally and passes it into `useChat`, so consumers cannot supply a different transport or cancellation policy:

https://github.com/cloudflare/agents/blob/5f6214dccfe3ba9bf243ec15d291b37b9659a54c/packages/ai-chat/src/react.tsx#L886-L943

The public `UseAgentChatOptions` exposes `resume`, `body`, `prepareSendMessagesRequest`, etc., but no cancel policy / `cancelOnClientAbort` / transport escape hatch:

https://github.com/cloudflare/agents/blob/5f6214dccfe3ba9bf243ec15d291b37b9659a54c/packages/ai-chat/src/react.tsx#L366-L493

On the server side, that frame is treated as a real cancellation, not as a client-only cleanup signal. In `AIChatAgent`:

https://github.com/cloudflare/agents/blob/5f6214dccfe3ba9bf243ec15d291b37b9659a54c/packages/ai-chat/src/index.ts#L828-L832

In the shared chat protocol parser:

https://github.com/cloudflare/agents/blob/5f6214dccfe3ba9bf243ec15d291b37b9659a54c/packages/agents/src/chat/parse-protocol.ts#L89-L90

And in `Think`:

https://github.com/cloudflare/agents/blob/5f6214dccfe3ba9bf243ec15d291b37b9659a54c/packages/think/src/think.ts#L2555-L2556

## How this can happen without Cloudflare explicitly aborting on WebSocket close

This is the part that is easy to misunderstand. #1164 says Cloudflare's WebSocket reconnect/page navigation path should not itself fire the abort signal. I agree with that contract. The concern here is the layer above it: the AI SDK chat client has its own active response lifecycle and explicitly documents that abort and resumable streams currently conflict.

Pinned to `ai@6.0.174` at `0aed63e156d141b8090f87da6db763454c554317`, the AI SDK troubleshooting docs say that with `useChat({ resume: true })`, closing a tab, refreshing, or calling `stop()` can trigger an abort signal that interferes with resumption:

https://github.com/vercel/ai/blob/0aed63e156d141b8090f87da6db763454c554317/content/docs/09-troubleshooting/15-abort-breaks-resumable-streams.mdx#L10-L28

The AI SDK `Chat` object creates an `AbortController` per active response and passes that signal to the configured transport's `sendMessages()` call:

https://github.com/vercel/ai/blob/0aed63e156d141b8090f87da6db763454c554317/packages/ai/src/ui/chat.ts#L656-L684

Its explicit `stop()` implementation aborts that same active response controller:

https://github.com/vercel/ai/blob/0aed63e156d141b8090f87da6db763454c554317/packages/ai/src/ui/chat.ts#L584-L592

So there are two distinct things to keep separate:

1. **Cloudflare WebSocket connection close**: should be resume-only and should not directly abort the server turn.
2. **AI SDK active response abort / stream cancellation**: can happen as part of the client response lifecycle and is passed into Cloudflare's transport as `options.abortSignal` or `ReadableStream.cancel()`.

The bug is that Cloudflare currently treats (2) as server cancellation unconditionally.

## Why this is a bug

The Agents/ai-chat architecture already documents and implements resumable streams. For durable agents, a browser client should be able to disappear and reconnect without taking down the underlying turn.

The current transport conflates two different events:

1. **Explicit app/user cancellation**: the user clicked a Stop button or the app deliberately called `stop()` / cancelled the run.
2. **Client response disposal**: the browser tab, React tree, route, socket, or AI SDK client stream was torn down; the local client stream should clean up, but the durable server turn should continue and be resumable.

Only the first should send `CF_AGENT_CHAT_REQUEST_CANCEL` by default. The second should be local cleanup unless the app explicitly opted into tying server execution to client stream lifetime.

## Related upstream context

- #105 introduced `cf_agent_chat_request_cancel` for explicit stop propagation. That PR also calls out the desired durable property: generations are not interrupted if the browser refreshes or loses connection, while explicit stop should interrupt generation.
- #1164 is the closest prior discussion. It was closed with the explanation that WebSocket reconnect/page navigation should not fire abort and that abort should only happen for explicit cancellation, `abortActiveTurn()`, or clear. This issue is about the client transport implementation path where `ReadableStream.cancel()` / `abortSignal` are mapped to the same server cancel frame.
- #985 / #986 fixed the client getting stuck after explicit abort by ensuring the transport terminates the local stream and sends cancel.
- #1233 / #1234 fixed explicit stop for tool continuation streams by targeting the right continuation request id.

Those fixes make sense for explicit stop. They do not provide an application-level way to say: "this is a durable session; client response disposal is not server cancellation."

## Expected behavior

One of these shapes would work:

1. Make server cancellation explicit-only by default: `stop()` sends `CF_AGENT_CHAT_REQUEST_CANCEL`, while internal stream disposal / `ReadableStream.cancel()` / generic client abort only cleans up the local client stream.
2. Or add a `useAgentChat` option such as:

```ts
useAgentChat({
  agent,
  resume: true,
  cancelOnClientAbort: false, // durable-client policy
})
```

or a more explicit policy:

```ts
useAgentChat({
  agent,
  clientAbortPolicy: "local-only" | "cancel-server-turn",
})
```

The key requirement is that transparent server cancellation from client lifecycle should be opt-in, or there should be an explicit durable/no-client-cancel mode. `stop()` should still be able to send server cancellation.

## Observed impact

In a production app using `useAgentChat`, we observed assistant turns ending as `aborted` even though our application code did not call `stop()`, did not call `abortActiveTurn()`, and did not manually send `cf_agent_chat_request_cancel`. The app treats the browser as a reconnectable observer and expects the underlying Durable Object turn to continue; the client should reconnect and catch up from persisted stream chunks.

Current workaround is to wrap the `agent` passed to `useAgentChat` and suppress only `cf_agent_chat_request_cancel` frames. That restores durable-session semantics, but it is an application-level protocol shim that should be unnecessary.

## Environment

Direct package versions in the affected app:

- `@cloudflare/ai-chat`: `^0.6.2`
- `agents`: `^0.12.2`
- `@cloudflare/think`: `^0.5.3`
- `ai`: `^6.0.174`

whoiskatrin:
## Summary

Changes `useAgentChat` cancellation semantics so browser/client stream cleanup is local-only by default:

```tsx
useAgentChat({
  agent
});
```

By default, the browser is treated as a reconnectable observer of a server turn. Generic client-side stream abort/cleanup does **not** send `cf_agent_chat_request_cancel`, so the server turn can continue and be resumed later. Explicit `stop()` still sends `cf_agent_chat_request_cancel` and cancels the server turn.

For apps that intentionally want request-lifetime behavior, this adds:

```ts
cancelOnClientAbort?: boolean;
```

Set `cancelOnClientAbort: true` to make generic client abort/stream cleanup cancel the server turn.

## What the issue actually was

This is not about a raw WebSocket close directly calling server-side abort.

The problematic path is one layer higher:

1. `useAgentChat` wraps AI SDK `useChat`.
2. AI SDK `useChat` owns a local active response with an `AbortController`.
3. `WebSocketChatTransport.sendMessages()` receives that signal as `options.abortSignal`, and also exposes a `ReadableStream` whose `cancel()` method can be called by the client stream lifecycle.
4. Before this PR, both paths used the same `onAbort()` handler.
5. That handler always sent `cf_agent_chat_request_cancel`.
6. The server correctly treated that frame as real server-turn cancellation.

So the server could not distinguish these two cases:

- the user/app explicitly clicked Stop and wanted to cancel the server turn
- the client-side response/stream was aborted or disposed as part of browser/React/AI SDK lifecycle

For Durable Object backed agents, the second case should often be recoverable via stream resumption. The browser can disappear or detach while the durable server turn continues and buffers chunks for a later reconnect.

## What changed

- Added `cancelOnClientAbort?: boolean` to `useAgentChat` and `WebSocketChatTransport`.
  - Default is `false`: generic client abort/stream cancel is local-only.
  - `true` opts into request-lifetime behavior where generic client abort cancels the server turn.
- Removed the earlier `durable` / `serverTurnCancellation` API shape in favor of the clearer boolean toggle.
- Kept explicit `useAgentChat().stop()` as server-side cancellation regardless of `cancelOnClientAbort`.
- Updated docs and README for the default resumable semantics and the opt-in request-lifetime mode.

## Follow-up changes from review

- Separated "active server turn is cancellable" from "local stream is attached" inside `WebSocketChatTransport`.
- Kept detached server turns explicitly cancellable after local-only client abort.
- Made explicit cancellation work for resumed streams created by `reconnectToStream()`.
- Added support for explicitly cancelling fallback-observed server turns via `observeServerTurn()`.
- Cleared stored cancellation state when the server later sends `done` for a detached turn.
- Avoided starting a new server turn when `sendMessages()` is called with an already-aborted signal.
- Added edge-case coverage for local-only abort, explicit stop after detach, resumed-stream stop, fallback-observed stop, and `cancelOnClientAbort: true` with both abort signals and `stream.cancel()`.

## Tests

- `npm run test:workers -w @cloudflare/ai-chat -- ws-transport-abort.test.ts ws-transport-cancellation-policy.test.ts ws-transport-resume.test.ts`
- `npm run test:react -w @cloudflare/ai-chat -- use-agent-chat.test.tsx`
- `npm run check`

## Related

Fixes #1471

### DIFF
--- a/.changeset/durable-chat-cancellation.md
+++ b/.changeset/durable-chat-cancellation.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/ai-chat": minor
+---
+
+Add `durable` and `serverTurnCancellation` options to `useAgentChat`. `durable: true` treats browser/client stream cleanup as local-only while preserving explicit `stop()` as server-side turn cancellation.
--- a/.changeset/durable-chat-cancellation.md
+++ b/.changeset/durable-chat-cancellation.md
@@ -2,4 +2,4 @@
 "@cloudflare/ai-chat": minor
 ---
 
-Add `durable` and `serverTurnCancellation` options to `useAgentChat`. `durable: true` treats browser/client stream cleanup as local-only while preserving explicit `stop()` as server-side turn cancellation.
+Add `cancelOnClientAbort` to `useAgentChat`. Generic browser/client stream cleanup is now local-only by default so server turns can continue and resume; explicit `stop()` still cancels the server turn. Set `cancelOnClientAbort: true` to make generic client aborts cancel the server turn.
--- a/docs/chat-agents.md
+++ b/docs/chat-agents.md
@@ -775,18 +775,17 @@ function Chat() {
 
 ### Options
 
-| Option                        | Type                                            | Default             | Description                                                                                                                                             |
-| ----------------------------- | ----------------------------------------------- | ------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `agent`                       | `ReturnType<typeof useAgent>`                   | Required            | Agent connection from `useAgent`                                                                                                                        |
-| `onToolCall`                  | `({ toolCall, addToolOutput }) => void`         | —                   | Handle client-side tool execution                                                                                                                       |
-| `autoContinueAfterToolResult` | `boolean`                                       | `true`              | Auto-continue conversation after client tool results and approvals                                                                                      |
-| `resume`                      | `boolean`                                       | `true`              | Enable automatic stream resumption on reconnect                                                                                                         |
-| `durable`                     | `boolean`                                       | `false`             | Treat the browser as a reconnectable observer: generic client stream abort/cleanup is local-only, while explicit `stop()` still cancels the server turn |
-| `serverTurnCancellation`      | `"on-client-abort" \| "explicit-only"`          | `"on-client-abort"` | Advanced cancellation policy. `durable: true` defaults this to `"explicit-only"` unless explicitly overridden                                           |
-| `body`                        | `object \| () => object`                        | —                   | Custom data sent with every request                                                                                                                     |
-| `prepareSendMessagesRequest`  | `(options) => { body?, headers? }`              | —                   | Advanced per-request customization                                                                                                                      |
-| `tools`                       | `Record<string, AITool>`                        | —                   | Dynamic client-defined tools for SDK/platform use cases. Schemas are sent to the server automatically                                                   |
-| `getInitialMessages`          | `(options) => Promise<ChatMessage[]>` or `null` | —                   | Custom initial message loader. Set to `null` to skip the HTTP fetch entirely (useful when providing `messages` directly)                                |
+| Option                        | Type                                            | Default  | Description                                                                                                              |
+| ----------------------------- | ----------------------------------------------- | -------- | ------------------------------------------------------------------------------------------------------------------------ |
+| `agent`                       | `ReturnType<typeof useAgent>`                   | Required | Agent connection from `useAgent`                                                                                         |
+| `onToolCall`                  | `({ toolCall, addToolOutput }) => void`         | —        | Handle client-side tool execution                                                                                        |
+| `autoContinueAfterToolResult` | `boolean`                                       | `true`   | Auto-continue conversation after client tool results and approvals                                                       |
+| `resume`                      | `boolean`                                       | `true`   | Enable automatic stream resumption on reconnect                                                                          |
+| `cancelOnClientAbort`         | `boolean`                                       | `false`  | Cancel the server turn when generic client stream abort/cleanup occurs. Explicit `stop()` always cancels the server turn |
+| `body`                        | `object \| () => object`                        | —        | Custom data sent with every request                                                                                      |
+| `prepareSendMessagesRequest`  | `(options) => { body?, headers? }`              | —        | Advanced per-request customization                                                                                       |
+| `tools`                       | `Record<string, AITool>`                        | —        | Dynamic client-defined tools for SDK/platform use cases. Schemas are sent to the server automatically                    |
+| `getInitialMessages`          | `(options) => Promise<ChatMessage[]>` or `null` | —        | Custom initial message loader. Set to `null` to skip the HTTP fetch entirely (useful when providing `messages` directly) |
 
 ### Return Values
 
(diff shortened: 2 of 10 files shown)


advait:
@whoiskatrin LGTM thank you!

threepointone:
Very nice. I like this PR, but I think it could go further. This new toggle should probably be the default behaviour, since it's the overwhelming actual desired usecase (navigating to a new chat session shouldn't kill an existing stream), let's also rename the flag to `cancelOnClientAbort`, and default it to false. Pushed up a commit with these changes.

PR #1523, opened 2026-05-14
refactor(codemode): omit source code from tool results

mattzcarey:
## Summary

This PR removes the echoed source `code` field from codemode tool results.

Successful sandbox executions now return only:

```ts
{
  result: unknown;
  logs?: string[];
}
```

Instead of:

```ts
{
  code: string;
  result: unknown;
  logs?: string[];
}
```

The submitted code is already present in the tool call, so echoing it back in the tool result adds hella context without providing useful information to the model.

### DIFF
--- a/.changeset/silent-pandas-learn.md
+++ b/.changeset/silent-pandas-learn.md
@@ -1,5 +1,5 @@
 ---
-"@cloudflare/codemode": minor
+"@cloudflare/codemode": patch
 ---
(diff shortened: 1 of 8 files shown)


threepointone (approved):
I'm fine if the changeset is a patch for this, but your call

mattzcarey:
I'll patch this

PR #1535, opened 2026-05-15
chore(ci): use nx affected for tests + nrwl/nx-set-shas

mattzcarey:
## Summary
- Switch PR test step from `nx run-many -t test` to `nx affected -t test` so unchanged projects skip testing entirely.
- Add `nrwl/nx-set-shas@v5` to derive `NX_BASE`/`NX_HEAD` against the last successful main-branch run (handles PR and push-to-main correctly without explicit `--base`/`--head`).
- Bump checkout to `fetch-depth: 0` with `filter: tree:0` (blobless clone) since `nx-set-shas` walks history. `filter: tree:0` keeps it fast — full history, no blob content.

Build is intentionally left as full `npm run build` (still `nx run-many -t build`) because the trailing `pkg-pr-new publish --peerDeps ./packages/*` step requires every package's `dist/` to be present. The existing Nx cache (already keyed on `package-lock.json` + sha) keeps cache-hit packages near-instant on rebuilds, so the win on build is small anyway. The real savings are on test, the most expensive step.

This is the canonical Nx 22.x CI pattern — same shape as nrwl/nx's own `ci.yml`.

## Test plan
- [ ] CI on this PR completes successfully and `nx affected -t test` reports zero affected projects for a CI-only change (workflow YAML is not in any project's input set).
- [ ] Open a follow-up trivial PR touching a single package (e.g. `packages/agents`) and confirm only that package's tests run.
- [ ] Confirm `pkg-pr-new publish` still publishes all packages (build path unchanged).

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1535" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -7,17 +7,9 @@ on:
       - "design/**"
       - "*.md"
       - ".changeset/**"
-  push:
-    branches:
-      - main
-    paths-ignore:
-      - "docs/**"
-      - "design/**"
-      - "*.md"
-      - ".changeset/**"
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
   cancel-in-progress: true
 
 env:
@@ -68,13 +60,6 @@ jobs:
         if: steps.playwright-cache.outputs.cache-hit != 'true'
         run: npx playwright install --with-deps chromium
 
-      - name: Test (affected — pull_request)
-        if: github.event_name == 'pull_request'
-        run: CI=true npx nx affected -t test
+      - run: CI=true npx nx affected -t test
 
-      - name: Test (run-many — push to main)
-        if: github.event_name == 'push'
-        run: CI=true npx nx run-many -t test
-
-      - if: github.event_name == 'pull_request'
-        run: npx pkg-pr-new publish --peerDeps ./packages/*
+      - run: npx pkg-pr-new publish --peerDeps ./packages/*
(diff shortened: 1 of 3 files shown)


cjol:
One (perhaps overcautious) pattern I've used before is a partial test run like this for PRs where the CI loop is annoying, but then a full test run on merge to main so the build is at least visibly broken if something unexpected sneaks through.

mattzcarey:
Good call, just pushed [67f3044b](https://github.com/cloudflare/agents/pull/1535/commits/67f3044b) which wires this up. On `pull_request` we run `nx affected -t test`; on `push` to `main` we run the full `nx run-many -t test` as the safety net. Also updated AGENTS.md to match.

threepointone (approved):

.github/workflows/pullrequest.yml
```diff
@@ -7,9 +7,17 @@ on:
       - "design/**"
       - "*.md"
       - ".changeset/**"
+  push:
```
cjol: I had in mind that the `main` checks would be added to the existing `release` action so that it's a clearer division of responsibility. One file for pull requests, one file for main. But this works too.

mattzcarey:
Agreed, that's cleaner. Pushed [3015d7c7](https://github.com/cloudflare/agents/pull/1535/commits/3015d7c7) which:

- reverts the `push:` trigger / event-conditional gating from `pullrequest.yml` — back to PR-only
- grows `release.yml` to run `npm run check` + Playwright setup + `nx run-many -t test` before the changesets/publish step (bumped its `timeout-minutes` from 5 to 20)
- updates `AGENTS.md` to describe both workflows

Bonus: publishing is now gated on the full test suite passing, which it wasn't before.

PR #1547, opened 2026-05-18
refactor(codemode): dispatch all tool calls positionally

mattzcarey:
## Summary

Simplify codemode tool dispatch by treating every sandbox tool call as positional.

Previously providers could opt into `positionalArgs`, which made the sandbox proxy capture `(...args)` and made the dispatcher spread those args into the host function. Providers without the flag used a separate single-object path.

This PR removes that split. The sandbox proxy now always captures an argument list and the dispatcher always calls:

```ts
fn(...args)
```

Object-style calls still work naturally because they are just one positional argument:

```ts
await codemode.foo({ a, b }) // fn({ a, b })
await shell.writeFile("path", "contents") // fn("path", "contents")
```

## Changes

- Remove `positionalArgs` from `ResolvedProvider` and `ToolProvider`
- Remove `ToolDispatcher.positionalArgs`
- Use one dispatch path in both Dynamic Worker and iframe runtimes
- Update shell/browser/codemode call sites that used `positionalArgs: true`
- Rename the TanStack codemode test fixture from `.backup.ts` to the real test file and merge the existing small TanStack test into it

## Verification

```sh
npm test --workspace @cloudflare/codemode -- --run
npm run build
npm run check
```

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1547" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


packages/codemode/src/executor.ts
```diff
@@ -220,13 +205,8 @@ export class ToolDispatcher extends RpcTarget {
       return stringifyForCodemode({ error: `Tool "${name}" not found` });
     }
     try {
-      if (this.#positionalArgs) {
-        const args = argsJson ? parseForCodemode(argsJson) : [];
-        const result = await fn(...(Array.isArray(args) ? args : [args]));
-        return stringifyForCodemode({ result });
-      }
-      const args = argsJson ? parseForCodemode(argsJson) : {};
-      const result = await fn(args);
```
cjol, line 229: This is very slightly behaviour-changing in edge cases, I think. If args was undefined, we previously sent an empty object, but will now send undefined. Does that matter?

mattzcarey: I think I'm happy with this change. feel more correct if everything is positional.

mattzcarey: the ai sdk libs would have normalised this anyway. It would only be weird for people doing custom implementation in which case this was broken before so I think I'm actually fine with this being a patch.

threepointone (approved):

PR #1556, opened 2026-05-19
refactor(mcp): WorkerTransport extends SDK WebStandardStreamableHTTPServerTransport

mattzcarey:
Replace the 941-line in-house `WorkerTransport` with a thin subclass of the official MCP SDK's `WebStandardStreamableHTTPServerTransport`. The wrapper layers Workers-specific concerns on top of the SDK transport without forking it:

- **CORS** — preflight handling and response-header injection (`corsOptions`).
- **Persistent transport state** across DO hibernation via the existing `MCPStorageApi` adapter. `sessionId`, `initialized`, `initializeParams` are snapshotted after each request and replayed on cold start so client capabilities are restored without a fresh initialize round-trip.
- **SSE keepalive** — preserves the issue #1583 fix (cfc75bc9). The wrapper wraps SSE responses in a `TransformStream` that injects the shared `KEEPALIVE_FRAME` (`: keepalive\n\n`) at `KEEPALIVE_INTERVAL_MS` (25s) from `sse-keepalive.ts`. Keepalive is unconditional on POST response streams (no recovery path during a mid-flight tool call) and disabled on the standalone GET stream when an `eventStore` is configured (clients recover idle drops via `Last-Event-ID` instead).

Everything else — session validation, SSE streaming, protocol-version negotiation, event-store resumability, send/close lifecycle — is delegated to the SDK transport. **Net: ~500 fewer lines of code we have to maintain.**

## Public API unchanged

`WorkerTransport`, `WorkerTransportOptions`, `MCPStorageApi`, and `TransportState` all keep the same exported shape. Downstream callers (`createMcpHandler`, `McpAgent`, examples, third-party users) work without changes.

## Test changes are minimal

- Storage tests work unchanged — `MCPStorageApi` is preserved.
- Whitebox routing tests now reference SDK private field names (`_streamMapping`, `_requestToStreamMapping`) and the SDK's `{controller, encoder}` stream shape instead of the old `{writer, encoder, cleanup}`.
- One closed-stream test updated to assert silent drop (SDK behaviour) instead of throw (old behaviour) — a deliberate resilience choice the SDK makes for close-during-send races.
- `post-keepalive.test.ts` (the #1583 regression suite) passes unchanged.

## Verification

- ✅ 462 / 462 MCP tests pass
- ✅ Build clean
- ✅ Two files touched: `worker-transport.ts` and its test file

### DIFF
--- a/.changeset/worker-transport-extends-sdk.md
+++ b/.changeset/worker-transport-extends-sdk.md
@@ -1,0 +1,15 @@
+---
+"agents": minor
+---
+
+Refactor `WorkerTransport` to extend the official MCP SDK's `WebStandardStreamableHTTPServerTransport` instead of being a hand-rolled implementation.
+
+The wrapper is now a thin subclass that layers Workers-specific concerns on top of the SDK transport:
+
+- **CORS** — preflight handling and response-header injection (`corsOptions`).
+- **Persistent transport state** across DO hibernation via the existing `MCPStorageApi` adapter. `sessionId`, `initialized`, and `initializeParams` are snapshotted after each request and replayed on cold start so client capabilities are restored without a fresh initialize round-trip.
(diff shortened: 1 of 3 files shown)


mattzcarey:
Superseded by #1701 with the branch renamed to `feat/worker-transport-extends-streamable-http`.

PR #1558, opened 2026-05-19
fix: route RPC MCP responses by request id

### LINKED ISSUE #1540: RPC MCP transport can lose a response when handleMcpMessage calls overlap
When using MCP over the Agents SDK RPC transport, a completed tool call can still time out at the caller with:

```txt
Request timeout: No response received within 120000ms
```

Observed with an `McpAgent` proxying an upstream MCP server. The upstream tool returned successfully and the proxy logged the tool call as successful in ~3s, but the Agent-side MCP client later timed out waiting for the JSON-RPC response.

From the installed SDK code, `RPCServerTransport` appears to keep one `_pendingResponse` / `_responseResolver` pair for the transport. If two `handleMcpMessage()` calls overlap on the same `McpAgent` instance, the later call can replace the earlier waiter before its response is delivered.

Questions:

1. Is `handleMcpMessage()` expected to be called serially per RPC MCP server instance?
2. If not, should `RPCServerTransport` track pending responses by JSON-RPC request id?
3. If yes, could the docs mention that RPC MCP server implementations should serialize calls at the DO/session boundary?

Small repro shape:

```ts
class ProxyMcp extends McpAgent<Env> {
  server = new McpServer({ name: 'proxy', version: '0.1.0' })

  init() {
    this.server.registerTool('slow', {}, async () => {
      await new Promise((resolve) => setTimeout(resolve, 50))
      return { content: [{ type: 'text', text: 'ok' }] }
    })
  }
}

// Same Agent connects to ProxyMcp over RPC and starts overlapping tool calls.
```

Workaround on our side is to serialize `handleMcpMessage()` per `McpAgent` instance before delegating to `super.handleMcpMessage()`.

mattzcarey:
## Summary

Fixes #1540 by making `RPCServerTransport` correlate in-flight RPC MCP responses by JSON-RPC request id instead of using one shared pending response/resolver slot.

This preserves concurrent `tools/call` support for the same RPC-backed `McpAgent` instance: overlapping calls can now complete out of order and still resolve to the correct caller.

The transport now tracks:

- pending request waiters keyed by JSON-RPC request id for normal result/error responses
- related server-to-client requests/notifications via `TransportSendOptions.relatedRequestId`
- continuation waiters for elicitation responses where the original tool-call waiter already returned the server elicitation request
- close/timeout cleanup for all pending waiters

This replaces the earlier serialization approach. Serializing was safer than the old single-slot transport, but it would have reduced parallel tool-call capability below what MCP/JSON-RPC and the TypeScript SDK support.

## Testing

- `cd packages/agents && npx vitest run src/tests/mcp/transports/rpc.test.ts src/tests/mcp/elicitation.test.ts`
- `npm run check`
- Deployed `examples/mcp-elicitation` to `https://mcp-elicitation-demo.agents-b8a.workers.dev` and verified the full initialize → tools/call → elicitation/create → elicitation response → final tool result flow with curl.

### DIFF
--- a/packages/agents/src/mcp/index.ts
+++ b/packages/agents/src/mcp/index.ts
@@ -32,7 +32,6 @@ export abstract class McpAgent<
   State = unknown,
   Props extends Record<string, unknown> = Record<string, unknown>
 > extends Agent<Env, State, Props> {
-  private _rpcMessageQueue: Promise<void> = Promise.resolve();
   private _transport?: Transport;
   private _pendingElicitations = new Map<
     string,
@@ -420,13 +419,7 @@ export abstract class McpAgent<
       }
     }
 
-    const transport = this._transport;
-    const run = this._rpcMessageQueue.then(() => transport.handle(message));
-    this._rpcMessageQueue = run.then(
-      () => undefined,
-      () => undefined
-    );
-    return await run;
+    return await this._transport.handle(message);
   }
 
   /** Return a handler for the given path for this MCP.
--- a/packages/agents/src/mcp/index.ts
+++ b/packages/agents/src/mcp/index.ts
@@ -32,6 +32,7 @@ export abstract class McpAgent<
   State = unknown,
   Props extends Record<string, unknown> = Record<string, unknown>
 > extends Agent<Env, State, Props> {
+  private _rpcMessageQueue: Promise<void> = Promise.resolve();
   private _transport?: Transport;
   private _pendingElicitations = new Map<
     string,
@@ -419,7 +420,14 @@ export abstract class McpAgent<
       }
     }
 
-    return await this._transport.handle(message);
+    const run = this._rpcMessageQueue.then(() =>
+      this._transport!.handle(message)
+    );
+    this._rpcMessageQueue = run.then(
+      () => undefined,
+      () => undefined
+    );
+    return await run;
   }
 
   /** Return a handler for the given path for this MCP.
--- a/packages/agents/src/mcp/index.ts
+++ b/packages/agents/src/mcp/index.ts
@@ -405,6 +405,10 @@ export abstract class McpAgent<
 
     const transport = this._transport;
 
+    // RPC waits are intentionally in-memory. While a request/continuation is
+    // pending, keep the object alive so hibernation does not drop the resolver
+    // maps or the suspended tool stack. Making this sleep/resume across
+    // hibernation would require a durable continuation model instead.
     return await this.keepAliveWhile(async () => {
       // Intercept elicitation responses before they reach the transport.
       // Mirrors what onSSEMcpMessage() and StreamableHTTPServerTransport's
--- a/packages/agents/src/mcp/index.ts
+++ b/packages/agents/src/mcp/index.ts
@@ -420,9 +420,8 @@ export abstract class McpAgent<
       }
     }
 
-    const run = this._rpcMessageQueue.then(() =>
-      this._transport!.handle(message)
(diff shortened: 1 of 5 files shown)


packages/agents/src/mcp/index.ts
```diff
@@ -419,7 +420,13 @@ export abstract class McpAgent<
       }
     }
 
-    return await this._transport.handle(message);
+    const transport = this._transport;
+    const run = this._rpcMessageQueue.then(() => transport.handle(message));
```
threepointone: do you want to wrap this with a .keepAliveWhile()

threepointone (approved):
tentatively stamping

threepointone:
changeset needed ofc

terrizoaguimor:
Hi @mattzcarey — I missed this PR when I commented on #1540 earlier today. Started a fix at the transport level (locking `RPCServerTransport.handle` and `_awaitPendingResponse` directly); your `_rpcMessageQueue` at the agent level lands the same outcome with less surface — cleaner choice.

While reproducing the bug I wrote two regression tests that the single overlapping-handle test here doesn't cover. Both live in `packages/agents/src/tests/mcp/transports/rpc.test.ts`:

1. **`handle()` arriving while `_awaitPendingResponse()` is pending** — the elicitation continuation case. `_awaitPendingResponse` (mcp/index.ts:418) sits outside `_rpcMessageQueue`, so it's worth proving that path stays safe under the implicit serialization from the outer queue.
2. **`close()` with multiple pending waiters** — confirms closing the transport mid-queue doesn't leak resolvers or orphan timers.

Happy to push them on top of this branch, or open a follow-up PR after merge — whichever. Also curious if there's anything blocking the draft; @threepointone approved on 2026-05-19. If it's just bandwidth, I can help finalize.

terrizoaguimor:
@mattzcarey thanks for the refactor. Just pulled `2d746b4b` locally — `_pendingContinuations` plus the two tests you added (`should not steal an id-routed response while _awaitPendingResponse is pending` and `should reject pending waiters on close`) cover the two scenarios I flagged. 30/30 in `rpc.test.ts` pass on Node 24.

One thing I think is still open from @threepointone's earlier review: the `keepAliveWhile` wrap. The refactor moved the wait from a queue at the agent level into `_pendingRequests`/`_pendingContinuations` at the transport level, but `handleMcpMessage` still does two awaits without `keepAliveWhile`:

```ts
return await this._transport._awaitPendingResponse();
...
return await this._transport.handle(message);
```

The default RPC timeout is 60s, and `_pendingRequests`/`_pendingContinuations` live in DO memory. If the DO hibernates mid-await, the maps go away and any in-flight responses get dropped on rehydration. The same file already documents the pattern at `mcp/index.ts:327`:

> Keep the DO alive while we wait for the user's elicitation response. An unresolved Promise alone isn't enough to prevent hibernation.

Wrapping both awaits with `keepAliveWhile` would close that hole.

mattzcarey:
Added the keepAliveWhile wrap around the RPC MCP handler waits and pushed a follow-up comment explaining why it is intentional. I also deployed a temporary curl test with two variants (normal keepAlive vs keepAlive no-op). The one-shot active-request case succeeded with both, but the realistic two-step flow (start elicitation, let curl exit, wait ~150s, answer later) only succeeded with keepAlive; the no-keepAlive variant timed out because the in-memory elicitation resolver/continuation was gone. So this PR keeps the current in-memory model and avoids a larger durable-continuation/DB design change.

Issue #1553, opened 2026-05-19
Orphaned "processing" status: durable state desyncs from turn liveness after DO eviction


rwdaigle:
**Describe the bug**

A chat agent's durable `status` can become permanently pinned to `"processing"` with no turn actually running and no path back to `"ready"`. This is an orphaned-lock: `status` is acquired durably in `beforeTurn` (persisted via `Agent.setState()` → SQLite) but released only by in-process code (`onChatResponse` / `onChatError`). The "is a turn live" signal that pairs with it (`chatRecovery`'s fiber + the app's own in-memory gate) is *not* durable and resets to its default on the next DO instantiation. If the Durable Object is evicted/recycled after `beforeTurn` but before the turn reaches a recoverable-stream checkpoint (or before `onChatResponse`/`onChatError` run), the two facts desync forever:

- `status:"processing"` survives (durable).
- The in-memory liveness gate resets to "no turn" (ephemeral).
- `onChatResponse`/`onChatError` will never run (the turn is gone), so nothing clears `status`.
- `onChatRecovery` only fires if Think has a resumable-stream checkpoint *and* the agent is re-driven within `RECOVERY_FRESHNESS_MS`. A turn that died before that first checkpoint has nothing to recover, and once it's past the freshness window the stale branch only flips `status→ready` while **silently dropping the user's unanswered message** — a second latent bug behind the first.

Net effect for the operator: an indefinite spinner, the last message never answered, and (because `messageConcurrency = "queue"`) every subsequent message frozen behind a turn that ended without ending.

**To Reproduce**

1. Run a chat agent that extends `Think` with `messageConcurrency = "queue"` and `chatRecovery = true`, where `beforeTurn` sets a durable `state.status = "processing"` and `onChatResponse`/`onChatError` clear it (the standard pattern).
2. Submit a turn whose work includes a long async span with no stream output yet — e.g. a cross-DO RPC or a tool call that takes tens of seconds (in our case a queued message drained right after a publish, then a full site compile/`check_site` RPC).
3. While that turn is suspended on the async span (after `beforeTurn` fired, before the first recoverable-stream checkpoint and before `onChatResponse`), have the Durable Object evicted/recycled — i.e. exceed the eviction/duration boundary or otherwise lose the isolate.
4. Let the DO go idle and be re-instantiated, and do not re-drive it within `RECOVERY_FRESHNESS_MS`.
[...]

PR #1559, opened 2026-05-19
feat: Stash chat metadata for recovery before inference starts

cjol:
## Summary
- Add early chat-turn recovery snapshots so interrupted turns can be recovered before any stream metadata or chunks exist.
- Add `ChatRecoveryOptions.retry` so recovery can retry the latest unanswered user message instead of continuing a partial assistant response.
- Apply the recovery behavior consistently across `@cloudflare/think` and `@cloudflare/ai-chat`.

## Details
This addresses the pre-stream recovery gap from #1553. Previously, a turn interrupted after turn setup but before the first recoverable stream checkpoint could leave no useful chat recovery context. Apps that set durable “processing” state around a turn had no reliable SDK hook to reconcile that state.
The new shared recovery snapshot is written when a chat fiber starts. On recovery, the framework can now call `onChatRecovery` even when `streamId` and `partialText` are empty. Apps can then either return `{ retry: true }` to re-run the latest unanswered user message, or return `{ continue: false, persist: false }` and surface the interruption themselves.


threepointone:
Thanks again for putting this together. The core idea here was exactly right: chat recovery needed a snapshot before inference starts, otherwise pre-stream interruptions have no durable context to recover from.

I pushed the implementation a bit further before merging this back:

- Moved the chat recovery stash wrapping into the base fiber internals instead of overriding `stash()` in both `AIChatAgent` and `Think`. That keeps the framework metadata scoped to the active fiber via AsyncLocalStorage, which avoids leaking recovery state across concurrent turns.
- Removed `retry` as a public `ChatRecoveryOptions` option. After thinking through the semantics, it felt too easy to misuse once partial assistant output exists. Instead, the framework now automatically retries only the narrow safe case: recovery happened before any stream chunks were written, and the latest persisted message is still the unanswered user message from the interrupted turn.
- Made recovered retry and continuation jobs self-contained by carrying the recovered `lastBody` and `lastClientTools` in the scheduled payload. That way alarm execution does not depend on whatever mutable request context happens to be restored later.
- Tightened the recovery snapshot unwrapping so malformed or future-version framework envelopes fall back to normal user recovery data rather than being trusted blindly.
- Added regression tests across base fibers, shared chat recovery helpers, `@cloudflare/ai-chat`, and `@cloudflare/think` for the pre-stream path, concurrent stash isolation, and recovered request context.

So the PR still solves the original issue, but with the recovery policy kept internal and a bit more guarded around concurrency and partial-output edge cases. Really appreciate you finding and tackling this gap.

PR #1567, opened 2026-05-20
fix: fire chat responses for errors returned before parts are returned

### LINKED ISSUE #1527: Stream errors before any parts silently drop the response hook
**Describe the bug**

When `streamText` produces an in-band `{ type: "error" }` UI message chunk **before any text or tool parts have accumulated**, `_streamResult` on the WebSocket primary path silently drops the turn: it broadcasts a transient error frame, lets the iterator close normally, and falls through without firing either `onChatResponse` or `onChatError`. The assistant message is persisted with `parts: []`, and any per-turn state that subclasses mirror from those hooks (status flags, in-flight gates, pending-message drains) is permanently left out of sync.

Reading `node_modules/@cloudflare/think/dist/think.js`:

- `:2162-2172` — the error UI chunk is handled in-band: Think broadcasts `MSG_CHAT_RESPONSE { error: true, done: false }` and `continue`s the loop.
- `:2221` — persistence **and** the response hook are both gated on `accumulator.parts.length > 0`. When the error chunk arrives before any `text-delta`/`start` parts, this guard fails and neither `_persistAssistantMessage` nor `_fireResponseHook` runs.
- `:2194` `catch` — only entered on iterator throws. Provider errors don't reach it because the AI SDK promotes them to in-band chunks rather than throws (`ai/dist/index.js:7601-7605`, `:8152-8158`).
- `onChatError` is only invoked from the RPC / sub-agent paths (`:2125`, `:2127-2135`) — not from the WebSocket primary path — so overriding it on the agent cannot rescue this case either.

Every `Think<Env>` subclass whose lifecycle relies on `onChatResponse` / `onChatError` running on every turn has a state-corruption bug whenever the model errors before producing content.

**To Reproduce**

1. Set up any `Think<Env>` agent with `chatRecovery: true` and a tool whose `execute` returns a large string field (e.g. ~1 MB of base64), with a `toModelOutput` that ships the field as image content to a strict provider (Anthropic).
2. Turn 1: ask the model to call the tool. The model produces a normal response; `onChatResponse({status:"completed"})` fires; state is clean.
3. Turn 2: ask any follow-up. Before assembling messages for `streamText`, `truncateOlderMessages` clips the previous turn's large string field to ~100 chars + a `... [truncated N chars]` suffix. The provider receives a malformed image and returns an API error before any chunks stream.
4. On the wire: a single `MSG_CHAT_RESPONSE { error: true, done: false }` followed by `done: true / error: false`. The assistant message in the session has `parts: []`.
5. On the agent: any state the subclass set in `beforeTurn` is never cleared — neither `onChatResponse` nor `onChatError` is called. Subsequent turns queue behind a perceived "in-flight" turn or display "processing" forever.

A simpler scripted repro: replace step 1's tool with anything that lets you inject an `AI_APICallError` at the very start of turn 2, before any UI parts arrive. Same outcome.

**Expected behavior**

`_streamResult` should converge on a single completion event regardless of whether the stream produced any parts. When the iterator finishes and an in-band error chunk was observed, Think should fire `_fireResponseHook` with `status: "error"` (and the error payload) so subclasses get a reliable single callback for cleanup — even if `parts.length === 0`.

Today, persistence and the response hook are coupled by the same guard at `:2221`. The fix is to decouple them: skip persistence when there are no parts, but still fire the hook so subclasses can clear state, drain queues, and surface the error to clients.

**Screenshots**

N/A — this is a server-side state issue. The user-visible symptom is a chat input stuck in "sending", an agent status flag stuck at "processing", and an empty assistant message row in the thread.

**Version:**

- `@cloudflare/think` 0.6.0
- `agents` 0.12.4
- `ai` 6.0.182
- `@ai-sdk/anthropic` 3.0.76
- Cloudflare Workers runtime, `nodejs_compat`

**Additional context**

- Client-side, `useAgentChat` *does* pass `error: Error | undefined` and `status === "error"` through from `@ai-sdk/react`'s `useChat` (`@cloudflare/ai-chat/dist/react.js:1270-1271`). Apps that wire those fields can show a transient banner — but that's purely client state. The **server-side session is left inconsistent**, which is what the proposed fix addresses.
- `chatRecovery: true` does not rescue this: the fiber completed without throwing, so `_handleInternalFiberRecovery` doesn't kick in on the next DO start.
- Workaround: override `onChunk`, watch for `{ type: "error" }` chunks, and run cleanup defensively. It works, but every Think subclass has to reinvent it. A fix inside `_streamResult` would let subclasses delete the override.
- Suggested minimal patch sketch (around `:2221`):
  ```ts
  const status = streamError ? "error" : aborted ? "aborted" : "completed";
  if (this._turnQueue.generation === clearGen) {
    if (accumulator.parts.length > 0) {
      await this._persistAssistantMessage(/* … */);
    }
    await this._fireResponseHook({ status, error: streamError, message: assistantMessage });
  }
```

cjol:
Improves error handling in `@cloudflare/think by ensuring that the `onChatResponse` hook is fired when an in-band stream error occurs before any response parts are emitted. Also tests.

Fixes #1527 

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1567" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/fix-think-inband-stream-error-hooks.md
+++ b/.changeset/fix-think-inband-stream-error-hooks.md
@@ -1,5 +1,7 @@
 ---
+"agents": patch
+"@cloudflare/ai-chat": patch
 "@cloudflare/think": patch
 ---
 
-Fire `onChatResponse` for in-band stream errors that arrive before any response parts.
+Return error statuses for in-band stream errors across programmatic chat turns.
--- a/.changeset/fix-think-inband-stream-error-hooks.md
+++ b/.changeset/fix-think-inband-stream-error-hooks.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/think": patch
+---
+
+Fire `onChatResponse` for in-band stream errors that arrive before any response parts.
--- a/docs/chat-agents.md
+++ b/docs/chat-agents.md
@@ -334,9 +334,11 @@ messages against the latest persisted transcript when the turn actually starts.
 This avoids stale baselines when multiple `saveMessages()` calls queue up
 behind active work.
 
-`saveMessages()` returns `{ requestId, status }`. The `status` is `"completed"`
-when the turn ran, `"skipped"` when it was invalidated (chat cleared mid-flight),
-or `"aborted"` when an external `AbortSignal` cancelled it before completion.
+`saveMessages()` returns `{ requestId, status, error? }`. The `status` is
+`"completed"` when the turn ran, `"error"` when the stream reported an error,
+`"skipped"` when it was invalidated (chat cleared mid-flight), or `"aborted"`
+when an external `AbortSignal` cancelled it before completion. When `status` is
+`"error"`, `error` contains the stream error message when available.
 
 Pass `options.signal` to cancel the turn from outside. Useful for forwarding
 an upstream `AbortSignal` (e.g. an AI SDK tool `execute`'s `abortSignal` on a
@@ -361,7 +363,7 @@ retained, streaming tools.
 
 ### `onChatResponse`
 
-Called after a chat turn completes and the assistant message has been persisted. Override this to react when the agent finishes responding — broadcast state, process queued work, track analytics, or trigger follow-up messages.
+Called after a chat turn reaches a terminal state. Override this to react when the agent finishes responding — broadcast state, process queued work, track analytics, or trigger follow-up messages. Successful turns and errors that produced partial assistant chunks persist the assistant message before the hook runs; errors that occur before any assistant chunks still fire the hook with an empty assistant message and `status: "error"`.
 
 ```typescript
 import { AIChatAgent, type ChatResponseResult } from "@cloudflare/ai-chat";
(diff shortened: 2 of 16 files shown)


threepointone:
Thanks again for putting this together. The original fix identified the right bug: in-band stream errors were being observed by the chat protocol, but the server-side lifecycle and programmatic surfaces could still treat the turn as successful or fail to report the error consistently.

I expanded the patch so the behavior is now consistent across the chat entry points:

- `SaveMessagesResult.status` now includes `"error"`, with an optional `error` message. This means programmatic callers of `saveMessages()` and continuation-style helpers no longer see a misleading `"completed"` result when the stream reports an error.
- Think now treats in-band stream errors as terminal in `_streamResult()` and RPC `chat()` streaming. Once an error chunk is observed, we stop consuming additional chunks, fire `onChatResponse` with `status: "error"`, and surface the same error through RPC `onError`.
- Partial assistant output before the error is still persisted. If the error happens before any assistant parts are produced, `onChatResponse` still fires with an empty assistant message and `status: "error"`.
- Durable submissions now store `status: "error"` and preserve the original stream error text, including recovered submission paths that previously could fall back to generic recovery messages.
- `@cloudflare/ai-chat` now mirrors the same behavior for SSE error chunks and thrown stream failures: programmatic turns return `{ status: "error", error }`, agent-tool child runs fail with the real stream error, terminal SSE errors cancel the reader, and error turns no longer emit success observability events.
- Updated docs and type tests for the expanded `SaveMessagesResult` contract.
- Added focused tests covering Think WebSocket/programmatic/RPC/submission paths and AI Chat programmatic, hook, observability, and agent-tool behavior.

The main user-visible behavior is: if a provider/model reports an error in the stream, the turn now ends as an error everywhere instead of looking completed in some APIs. Any partial assistant text that arrived before the error is still saved, and callers can read the error message from the hook/result.

One TypeScript compatibility note: exhaustive `SaveMessagesResult.status` switches will now need to handle `"error"`. I think that is the right tradeoff here because it reflects the runtime state accurately rather than forcing callers to infer failures from side channels.

PR #1581, opened 2026-05-22
feat(codemode): connector model + durable runtime, snippets, and vite plugin

mattzcarey:
## feat(codemode): connector model + durable runtime, snippets, and vite plugin

Redesigns codemode into class-based **connectors** running on a durable **runtime**. The model gets one tool — `codemode({ code })` — and discovery, approvals, replay, and state all live behind it.

The whole developer experience:

```ts
const runtime = createCodemodeRuntime({
  ctx: this.ctx,
  executor: new DynamicWorkerExecutor({ loader: this.env.LOADER }),
  connectors: [github, repoApi]
});

tools: {
  codemode: runtime.tool();
}
```

The surface is deliberately minimal and split by audience:

- **The developer configures** — a connector class per service, plus the runtime handle as the control plane: `tool()`, `pending()`, `approve()`, `reject()`, `rollback()`, `executions()`, `saveSnippet()`/`snippets()`/`deleteSnippet()`. The handle is the **only** public API — the low-level proxy-tool functions are internal.
- **The model uses** — four sandbox methods (`codemode.search`/`describe`/`step`/`run`) plus one typed global per connector.

Two layers underneath, cleanly split:

- **Executor** — the simple, stateless code sandbox. Runs a block of code once, dispatches tool calls back. `DynamicWorkerExecutor` (Workers), `IframeSandboxExecutor` (browser).
- **Runtime** — public runtime handle plus `CodemodeRuntime` DurableObject facet. The handle owns the executor/connectors for this run; the facet owns durable log, approvals, executions, and snippets.

### Connectors

`WorkerEntrypoint` subclasses that bridge external services. Constructors carry dependencies; a connector is three things — `name()`, optional `instructions()`, and `tools()`. Everything about a tool (docs, schema, approval, execution, revert) lives in one record entry; derived connectors generate the record and you decorate it via the `tool(name, t)` hook:

```ts
// github.codemode.ts
export class GithubConnector extends McpConnector<Env> {
  name() {
    return "github";
  }
  protected instructions() {
    return "Use for GitHub operations.";
  }
  protected createConnection() {
    return this.conn;
  }
  // decorate derived MCP tools: approval + rollback in one place
  protected tool(name, t) {
    if (name === "create_issue") {
      return {
        ...t,
        requiresApproval: true,
        revert: (_args, result) => this.closeIssue(result.number)
      };
    }
    return t;
  }
}
```

**Base classes:** `CodemodeConnector` (author `tools()` directly; AI SDK toolsets are shape-compatible), `McpConnector`, `OpenApiConnector`
**Authoring surface:** `name()`, `instructions()?`, `tools()`, `tool(name, t)?`
**RPC wire surface** (derived, Gatekeeper-compatible): `describe()`, `executeTool()`, `revertAction()`, `getTypeScriptTypes()`

### Runtime — durable execution via abort-and-replay

`CodemodeRuntime` is a DO facet that owns a per-execution **tool-call log** (the replay spine):

- Every connector call **and** every `codemode.step(name, fn)` is recorded in the log.
- Reads / steps execute and their result is recorded.
- An approval-required action is recorded as `pending` and the run **aborts**.
- On resume, the same code re-runs: calls already in the log are served from it (noop — reads/steps return recorded results), and the newly-approved action executes for real. The run proceeds to the next pause or to completion.

```
run 1:  list_prs() ─exec→ [...]          [applied]
        create_issue() ─PAUSE─           [pending]   ✗ aborts
user approves
run 2:  list_prs() ─replay→ [...]        (from log, no re-exec)
        create_issue() ─exec→ { number } (approved, runs for real)   ✓ completes
```

Lifecycle helpers (driven by the agent, e.g. from `@callable()` methods):

```ts
const runtime = createCodemodeRuntime({ ctx, connectors, executor });
const pending = await runtime.pending(); // list actions awaiting approval
await runtime.approve(); // approve + continue
await runtime.approve({ executionId }); // resume a specific execution
await runtime.reject({ seq }); // reject — ends execution
await runtime.rollback(); // revert applied actions in reverse
const runs = await runtime.executions(); // audit trail, newest first
await runtime.saveSnippet("name", { executionId }); // promote a run to a snippet
await runtime.snippets(); // list; runtime.deleteSnippet(name) removes
```

### codemode.step — the explicit side-effect boundary

A closure can't cross the RPC boundary, so `step` is **sandbox-side JS** (a provider `prelude`) that runs the closure locally and records only its result via two host primitives. Anything that isn't a connector call — raw `fetch`, random, time — goes in a step so it runs once and replays thereafter:

```ts
const id = await codemode.step("gen-id", () => crypto.randomUUID());
const data = await codemode.step("fetch", async () =>
  (await fetch(url)).json()
);
```

Steps share the same sequence/cursor as tool calls, so they interleave and replay in order. Replay divergence (the Nth recorded op differs across runs) is a hard error.

### Model-facing tool

One tool, one input: `codemode({ code })`. Inside the sandbox:

```ts
codemode.search("pull request")     // ranked search across methods + snippets
codemode.describe("github.foo")     // TypeScript type docs
codemode.step(name, fn)             // record-once side-effect boundary
codemode.run("name", input)         // run a snippet the developer saved
github.list_pull_requests(...)      // connector methods as globals
```

### Architecture

Connectors are passed as **env bindings** to the sandbox; calls go via Workers RPC directly (no ToolDispatcher serialization), through the runtime for a replay/execute/pause decision:

```
sandbox: github.list_pull_requests(args)
  → env.__connectors.github.callTool("list_pull_requests", args)
  → runtime.decide(...) → execute → GithubConnector.executeTool() → runtime.recordResult(...)
```

The platform SDK (`codemode` namespace) uses the provider/dispatcher path since it's host-created, not a WorkerEntrypoint.

```
Agent DO
  ├─ runtime handle    (executor, connectors, lifecycle helpers)
  └─ facet: codemode   (CodemodeRuntime — durable log, approvals, state)

runtime.tool()       (spawns the facet, builds platform SDK, runs code in the Executor)
```

### Snippets

Reusable code patterns. Connectors provide raw capability; snippets are recipes that worked. The model writes and runs scripts; the developer promotes the good ones; the model reuses them:

```ts
// host side — e.g. a "save this script" button
const runs = await runtime.executions();
await runtime.saveSnippet("repo-overview", {
  executionId: runs[0].id,
  description: "Fetch repo metadata, open PRs, and latest releases."
});

// sandbox side — the model
const overview = await codemode.run("repo-overview", { owner, repo });
```

Snippets live on the runtime facet, surface in `codemode.search`/`describe`, and the facet's identity is derived from the connector set — so a snippet can never reference a connector that isn't present. There is no separate skill-source interface; external snippet sources (e.g. Think) can come later.

### Vite plugin

`@cloudflare/codemode/vite` discovers `*.codemode.ts` files and auto-exports connector classes (and `CodemodeRuntime`) from the worker entry:

```ts
// vite.config.ts
import codemode from "@cloudflare/codemode/vite";
export default { plugins: [codemode()] };

// server.ts
import { GithubConnector } from "./github.codemode" with { type: "connectors" };
```

### Other changes

- The runtime handle is the one public API: `createProxyTool`/`resumeCodemode`/`rollbackCodemode` etc. are no longer exported
- Executor-style ranked search (normalize, tokenize, score, coverage threshold, cap 50)
- `ToolAnnotations` type (`requiresApproval`) — annotate only what needs a human; unannotated methods run immediately. The approval UI describes the action from the method's own descriptor plus its args
- `DynamicWorkerExecutor.bindings` slot for passing ServiceStubs as env
- `ConnectorBinding` on `Executor.execute()` options; `ResolvedProvider.prelude` for sandbox-side fns
- `OpenApiConnector` exposes two overridable primitives — `spec()` (the OpenAPI doc, returned into the sandbox, no prompt tokens) and `request()` (an authenticated request). The model reads the spec and writes its own lookup; no pre-baked per-operation methods. (`request`, not `fetch` — `fetch` is reserved by `WorkerEntrypoint`.)
- Example: `examples/codemode-connectors` (`github.codemode.ts`, `repoapi.codemode.ts`)
- Docs: `docs/codemode/{index,connectors,runtime,snippets,approvals,vite-plugin}.md`

### Design notes

- **Abort-and-replay, not generators.** True mid-stack suspension would require `yield*` everywhere (a real error surface for LLM-written code). We keep plain async/await and re-run between pauses — cheap, since prior calls serve from the log. `step` gives explicit-effect correctness, the one idea worth borrowing from generator-based durable-execution engines.

### Not in this PR (future)

- External snippet sources (e.g. Think exposing session memory/skills as snippets)
- Approval-queue UI wired to agent `@callable()` + `useAgent().call`
- React hooks (`useCodemodePending` etc.) once the approval UI shape is proven
- capnweb-validate `@validateRpc()` on the connector RPC surface


docs/codemode/approvals.md
```diff
@@ -0,0 +1,177 @@
...
+Later:
+  → Agent queries codemode.pending()
+  → Agent shows to user: "GitHub wants to create an issue. Approve?"
+  → User approves
+  → Agent calls session.applyAction(actionId)
+  → Session calls connector.executeTool("create_issue", args) for real
+  → Real result replaces provisional
+```
+
+## Simulation
```
cjol: I really like the optimistic simulation model. Is that novel here or from gatekeeper/executor? I keep thinking about what it means for multi-step dependent actions though (which are kind of the bread and butter of code mode). Later calls might be hard to faithfully simulate (or grant permission for) if their inputs depend on values that don't exist until an earlier mutation actually runs. 


I wonder if we could preserve and surface a symbolic dependency graph? For example "Grant permission for `create_issue` with title=`action_1.generated_title`", which can still be resolved/rewritten when `action_1` is executed but contains a bit more information in the meantime

mattzcarey: yep definitely need to think more about this

threepointone:
## This session's work: two phases

Two commits landed this session, on top of the existing connector + durable runtime work. Phase 1 hardens the runtime's correctness under hibernation/concurrency and cleans up the API; phase 2 adds the two primitives a stateful connector (e.g. reusable browser sessions, see #1656) needs to ride on the durable runtime, plus a result-shaping hook.

---

### Phase 1 — `a14c42bf` `fix(codemode): harden durable runtime — stateless, explicit executionId, resilient rollback`

**Runtime architecture**
- **Stateless across calls.** The `CodemodeRuntime` facet keeps no in-memory cursor or annotations; every interaction is addressed by `(executionId, seq)` with `seq` allocated host-side. *Why:* a run must survive the Durable Object being evicted between any two tool calls, and multiple runs must be able to interleave in one DO without clobbering each other.
- **Removed the global `CURRENT_KEY` "current execution" pointer** and its helpers. `approve`/`reject`/`rollback`/`saveSnippet` now require an explicit `executionId`, and every tool outcome carries `executionId`. *Why:* a single shared "current run" pointer is racy the moment two runs are in flight; making the target explicit removes a whole class of concurrency bugs.

**Replay correctness**
- **`executing` log state.** A non-approval call/step is logged `executing` by `decide()` and only flips to `applied` once `recordResult()` stores its value. *Why:* a crash between deciding and recording would otherwise replay `undefined`; now it re-executes.
- **Replay divergence detection.** Connector/method/args are compared (args via a stable stringify) against the recorded run; a mismatch is recorded as a failed execution and **returned** as `{ status: "error" }` rather than thrown. *Why:* catch nondeterminism that wasn't wrapped in `codemode.step()` without breaking the agent loop with a cross-RPC exception.
- **Terminal-state guard in `decide()`.** Once a run is paused/terminal, further calls/steps get a pause decision and record nothing. *Why:* model code that catches the pause sentinel and keeps going must not be able to apply extra side effects.

**Approvals & rollback**
- **`rollback()` reverts all applied reversible actions** (any tool with a `revert`, not just approval-gated ones), in reverse order, attempting every revert even if one throws (failures aggregated afterward), and marks the run `rolled_back`. *Why:* `requiresApproval` (pause-before-do) and `revert` (undo-after-do) are orthogonal; a failing compensation must not strand later ones.
- **`pending()` aggregates across all paused runs** when no `executionId` is given. *Why:* an approvals UI must see every paused run, not just the last one touched.
- Documented that `reject()` ends a run but does not auto-undo applied actions.

**Retention**
- `begin()` auto-prunes terminal executions to `maxExecutions` (default 50); added `deleteExecution()` / `pruneExecutions()`. Running/paused runs are never pruned.

**Connectors / DX**
- `OpenApiConnector` derives one typed tool **per operation** host-side (zero prompt tokens), with deeper `$ref` resolution (`allOf`/`oneOf`/`anyOf`), `WeakMap` memoization by spec identity, and collision warnings. `request()` stays as an escape hatch.
- Connector bindings are passed as `RpcTarget` `evaluate()` arguments (fixing a `DataCloneError`); pause is signalled via a control marker instead of throwing across the sandbox→host RPC boundary.
- Executor switched to `loader.load()` for one-off dynamic workers (`loader.get(random-id)` gave no caching benefit).
- `CodemodeConnector` ctx widened to `DurableObjectState | ExecutionContext` so connectors inside a DO don't cast `this.ctx`.

**Tests / cleanup**
- New `src/runtime-tests/` e2e suite (real DO host + facet + `DynamicWorkerExecutor`): read-only, pause/approve, replay, reject, rollback, divergence, step replay-safety, concurrent runs, retention, snippets, pause-swallow guard, pending aggregation.
- Deleted orphaned `src/mcp-provider.ts` and stray `.pr-body.md` / `EXECUTOR_TODO.md`.

---

### Phase 2 — `4fe9d937` `feat(codemode): per-execution connector lifecycle + result shaping`

This phase adds exactly what a connector needs to own a stateful resource for the lifetime of one run (the thing #1656's browser sessions require), so that work can build on the connector + durable runtime instead of reinventing session storage. Everything is additive.

**Per-execution resource lifecycle**
- **`execute(args, ctx)` / `revert(args, result, ctx)` receive a `ToolExecuteContext` with the run's `executionId`** — stable across pause/resume. *Why:* a connector can lazily acquire (or reconnect to) a resource keyed by that id, so the resource is addressable even after a pause that resumes in a fresh Worker invocation.
- **`CodemodeConnector.disposeExecution(executionId, status)`** — optional override (default no-op), called when a run reaches a terminal state so the connector can tear the resource down. It fires on each terminal transition (`completed`/`error`/`rejected`/`rolled_back`) and **never on pause**. *Why:* a paused run may resume, so a run-scoped resource must outlive a pause; the terminal transition is the single safe teardown point. Documented to be idempotent (a completed-then-rolled-back run disposes twice), to read state from durable storage rather than instance memory (the connector may be reconstructed / the opening pass may have hibernated), and to never throw (rejections are swallowed so teardown can't fail a finished run).
- **Stale `reject()` no longer disposes.** `runtime.reject()` now returns whether it actually terminated the run, and teardown is gated on that. *Why:* a duplicate/stale reject (seq no longer pending, e.g. from a second browser tab) is a no-op — the run is still live and resumable, so its resources must be left intact.

**`rejected` is now a first-class `ExecutionStatus`**
- `reject()` marks the run `rejected` instead of masquerading as `error`. *Why:* a user rejection isn't a failure; the audit trail (`executions()`) should distinguish them, and this makes `ExecutionEndStatus` exactly the terminal subset of `ExecutionStatus`.

**Result shaping**
- **`createCodemodeRuntime({ ..., transformResult })`** reshapes the *model-facing* result of a completed run (initial run and resume), after the raw value is recorded. *Why:* lets a caller cap response size without losing audit fidelity — the execution keeps the full result, the model sees the shaped one. A throwing transform falls back to the raw result rather than failing a completed run.
- **Exported `truncateResult` / `truncateResponse`** (token-aware, `{ maxChars?, maxTokens? }`): small structured results pass through unchanged; oversized ones serialize to a bounded, marked string. These are the default building blocks for `transformResult`.
- Note: in the codemode model intermediate results stay in the sandbox, so this is a final-result safety net rather than the per-call truncation the old AI-SDK-tool model needed.

**Tests / docs**
- e2e: `executionId` threading, dispose on complete/reject/rollback, no dispose while paused, no dispose on a stale reject, `transformResult` on run + resume.
- unit: `truncateResponse` / `truncateResult`.
- Documented the lifecycle contract, result shaping, the `rejected` status, and a **determinism note**: issue connector calls sequentially — `Promise.all([a(), b()])` can reorder across a pause/resume and trip divergence detection.

---

Full suite green throughout: **282 unit + 20 e2e + 33 browser**, and `pnpm run check` (sherif, exports, oxfmt, oxlint, typecheck across 103 projects) passes.

PR #1584, opened 2026-05-24
Add first-class Agent Skills (`agents/skills`) with Think integration

threepointone:
## Summary

This PR adds first-class **Agent Skills** support, built as a framework-agnostic engine in **`agents/skills`** and wired into Think.

A skill is an [Agent Skills](https://agentskills.io/) directory (`SKILL.md` + colocated resources/scripts). The engine lets an agent advertise a compact skill catalog in the system prompt, activate full `SKILL.md` instructions on demand, read colocated resources, and optionally run skill scripts in a bounded, Worker-compatible runtime — instead of injecting everything into every prompt.

Because the engine lives in `agents/skills` (alongside the `agents/chat` shared layer), it is not tied to Think:

- `@cloudflare/think` re-exports it as `skills` and wires `getSkills()` into the turn automatically.
- `@cloudflare/ai-chat` — or any AI SDK caller — can build a `SkillRegistry` directly and merge `registry.tools()` + `registry.systemPrompt()` into its own `streamText` call.

```ts
import { Think, skills } from "@cloudflare/think";
import bundledSkills from "agents:skills"; // -> ./skills next to this file

export class MyAgent extends Think<Env> {
  getSkills() {
    return [
      bundledSkills,
      skills.r2(this.env.SKILLS_BUCKET, { prefix: "skills/" })
    ];
  }

  // optional, experimental
  getSkillScriptRunner() {
    return skills.runner({
      loader: this.env.LOADER,
      workspaceInstance: this.workspace
    });
  }
}
```

### `agents/skills` engine

- New `agents/skills` export: `SkillSource`, `SkillRegistry`, `fromManifest(...)`, `r2(...)`, the `SKILL.md` frontmatter parser, and the script `runner(...)`. The runner depends only on a local structural `SkillWorkspace` interface, so the engine does not couple to Think or `@cloudflare/shell`.
- Model-facing tools produced by the registry: `activate_skill`, `read_skill_resource`, and `run_skill_script` (only when a runner is configured).
- Resilient loading: duplicate skill names (first source wins) and sources that fail to list are skipped with diagnostics instead of throwing, so a bad source can never break `onStart` or a turn. The Vite plugin also warns on duplicate bundled names at build time.

### `agents:skills` import (Agents Vite plugin)

- Bundled skills are imported with an explicit `agents:skills` specifier (replacing the earlier `with { type: "skills" }` import attribute and its per-project type shim). The path is optional and defaults to a `./skills` directory next to the importer; `agents:skills/<dir>` targets a sibling directory.
- The plugin parses `SKILL.md` frontmatter, bundles allowed resource roots with binary-safe encoding/MIME/size metadata, watches files during dev, and warns when bundled assets exceed size thresholds (recommending `skills.r2()`).
- Ambient types ship from `agents` (`skills-module.d.ts`, referenced from the built `dist/index.d.ts`), so importing `agents` (directly or via `@cloudflare/think`) types the specifier with no per-project shim.

### R2-backed source

- `skills.r2(...)` reads standard `skills/<name>/SKILL.md` layouts, lazily fetches resources, supports metadata or content fingerprinting and refresh intervals, and handles binary resources safely.

### Experimental script runner (`skills.runner`)

- Flagged `@experimental` with a one-time warning on first use.
- **JS/TS scripts are function-style**: `export default async function run(input, ctx)` where `ctx = { skill, files, workspace, tools, output }`. Bundled text resources are exposed via `ctx.files`; scratch artifacts via `ctx.output.writeFile(name, content)`. (No `node:fs`/`path` shim.)
- Capabilities and permission enforcement live behind a **single `SkillScriptHostBridge`**, constructed fresh per `run()` so `/output` artifacts never leak between concurrent runs. JS providers, Python RPC, and Bash commands all delegate to it.
- Python and Bash scripts use the path-based `/skill` / `/input.json` / `/context.json` / `/output` contract. TypeScript is compiled via `@cloudflare/worker-bundler`.
- Default envelope: no network, no tools, read-only workspace access when `workspaceInstance` is provided, 30s timeout. Network, tools, and `workspace: "read-write"` are opt-in.

### `@cloudflare/worker-bundler`

- Adds a `virtualModules` option to `createWorker` for aliasing exact import specifiers to inline module source during bundling.

### Default Think workspace `bash` tool

- Enabled by default (opt out with `workspaceBash = false`), runs in a sandboxed `just-bash` virtual filesystem with network disabled, syncs created/updated/deleted/empty-directory changes back to the workspace, and protects skipped/oversized files.

### Example (`examples/agent-skills`)

- Full-stack demo: bundled skills via `agents:skills`, on-demand `activate_skill`, resource reads, and a function-style TypeScript `run_skill_script`.
- Skills: `release-notes` (TS script + reference), `test-plan` (procedure), `debug-plan` (procedure + reference), `pirate-voice`.
- The UI renders skill tool activity inline and lights up activated skills in the sidebar.

## Safety and compatibility model

Skills are useful without being ambiently powerful. Instructions load on demand; scripts get explicit capabilities through runner options; workspace access defaults to read-only and writes require `workspace: "read-write"`; network and tool access are opt-in. Resource paths are validated and binary-safe.

## Documentation

- `docs/think/index.md`, `packages/think/README.md`, `design/skills.md`, `packages/agents/AGENTS.md`, `packages/worker-bundler/README.md`, `examples/agent-skills/README.md`.

## Testing

- `SKILL.md` parsing/validation, manifest sources, duplicate/bad-source handling (no-throw), activation and resource reads, R2 indexing/refresh/binary/unsafe-path rejection, function-style JS/TS execution with `ctx`, bridge-backed `/output` and per-run isolation, Python and Bash execution, workspace read/write permissions, the Vite plugin specifier (default + named) and asset-size warning, `worker-bundler` virtual modules, and the default workspace Bash tool.
- Verified with `npm run check`, the full `@cloudflare/think` and `agents` test suites, and an end-to-end `vite build` of the example.

## Package impact

Changesets for minor releases of:

- `agents` (new `agents/skills` engine + Vite `agents:skills` transform)
- `@cloudflare/think` (consumes the engine, adds `getSkills()` wiring + workspace Bash)
- `@cloudflare/worker-bundler` (`virtualModules` option)

## Follow-ups

- Git-backed skill sources
- R2 write/delete helpers for skill-management UIs
- Continued hardening of script compatibility as real-world skill libraries exercise the runtime

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1584" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


abhagsain:
I hope skills can be used by the sub-agents? 
Main orchestrator can spawn different agents that can use different skills?

threepointone:
Yeah just define subagent think classes with the skills you need

zmzlois:
would be great for this to land, subscribe.

threepointone:
image

PR #1585, opened 2026-05-24
Add declarative scheduled tasks and handlers to Think

threepointone:
## Summary

This PR adds declarative scheduled tasks to `@cloudflare/think`: a `getScheduledTasks()` API that lets Think agents declare recurring prompt submissions or deterministic scheduled handlers in code, with durable scheduling, startup reconciliation, and stable idempotency context built in.

The goal is to make proactive agent work feel native to Think. Instead of manually calling `schedule()` in `onStart()` and wiring callback methods by hand, a Think subclass can now declare durable recurring work:

```ts
getScheduledTasks(): ThinkScheduledTasks {
  return {
    hourlyQueueDigest: {
      schedule: "every 1 hour",
      prompt:
        "Write a concise hourly reminder that durable background task queues should be checked for stuck or failed work."
    },
    dailyWorkflow: {
      schedule: "every day at 09:00",
      timezone: "UTC",
      retry: { maxAttempts: 3 },
      handler: async ({ idempotencyKey, scheduledFor, timezone }) => {
        await this.env.REPORT_WORKFLOW.create({
          id: idempotencyKey,
          params: { scheduledFor, timezone }
        });
      }
    }
  };
}
```

## What This Adds

- Adds `getScheduledTasks()` to `Think` for code-declared recurring prompts and handlers.
- Adds `getDefaultTimezone()` for app/user-level wall-clock timezone context.
- Adds exported scheduled-task types and `defineScheduledTasks()` for optional literal inference without method return annotations.
- Adds mutually exclusive `prompt` and `handler` task actions.
- Gives handlers stable occurrence context: `taskId`, `scheduledFor`, `scheduledForDate`, `occurrenceKey`, `idempotencyKey`, `schedule`, `scheduleKind`, `timezone`, and `metadata`.
- Adds a small deterministic recurring DSL:
  - `every <n> minutes`
  - `every <n> hours`
  - `every day at HH:mm`
  - `every weekday at HH:mm`
  - `every week on <day>[,<day>...] at HH:mm`
- Adds runtime validation for task IDs, DSL shape, timezone sources, pluralization, invalid/duplicate days, and interval/timezone mismatches.
- Adds timezone-aware wall-clock scheduling using `Intl.DateTimeFormat`, including explicit DST gap/overlap semantics.
- Reconciles declared tasks into owner-scoped one-shot Agent schedules and tracks them in `cf_think_scheduled_tasks`.
- Preserves schedule rows for action-only changes, replaces rows when timing semantics change, removes rows when tasks are removed, and repairs missing/pending schedule rows.
- Runs prompt occurrences through durable `submitMessages()` using stable owner/task/occurrence idempotency keys.
- Runs handler occurrences directly for app-owned deterministic work such as starting Workflows or writing ledgers.
- Supports sub-agents through the existing root scheduler delegation and owner-scoped schedule APIs.

## Why This Matters

Scheduled work is a core agent use case: daily summaries, recurring check-ins, background reports, routine reminders, health digests, and deterministic app jobs like Workflow starts. Before this PR, developers needed to reason about scheduler rows, callback names, idempotent `schedule()` calls in `onStart()`, payload stability, and duplicate execution safety themselves.

This API lifts that into Think's mental model:

- `getTools()` defines what the agent can do.
- `configureSession()` defines what the agent remembers.
- `getScheduledTasks()` defines what the agent should proactively do over time.

Prompt tasks are normal durable Think turns, so existing hooks, storage, submission inspection, recovery behavior, and idempotency semantics continue to apply. Handler tasks are for deterministic app-owned work and are delivered at least once, so callers should use the provided `idempotencyKey` or `occurrenceKey` for external side effects.

## Deploy/Reconciliation Semantics

Declared schedules are reconciled on startup after Think session/protocol setup and before recovered submissions drain. If `getScheduledTasks()` reads product-owned data that can change while the Durable Object is live, app code can call `internal_reconcileScheduledTasks()` after updating that data.

The reconciler:

- Validates the current declaration set.
- Normalizes each schedule and resolved timezone.
- Computes separate schedule/task hashes.
- Creates schedule rows for new tasks.
- Leaves the existing schedule row in place for prompt/handler/metadata-only changes.
- Cancels/replaces schedule rows when schedule, timezone, or retry semantics change.
- Cancels/deletes owned rows when tasks are removed from code.
- Repairs pending rows where the Think ledger exists but `schedule_id` has not been written yet.
- Recreates rows whose underlying Agent schedule is missing.
- Ignores stale callback payloads after task changes.
- Never scans or cancels unrelated user-created schedules.

## Retry, Failure, And Delivery Semantics

The task `retry` option applies to the scheduled action itself: prompt generation/submission for prompt tasks, or handler execution for handler tasks. After the action succeeds or exhausts its retries, Think still schedules the next occurrence so one bad run does not stall the recurrence chain.

Delivery remains at least once. If the Durable Object or scheduler retries after a partial failure, the same occurrence may be delivered again. Prompt tasks use stable `submitMessages()` idempotency keys; handler tasks receive stable keys and should use them when creating Workflows or performing external side effects.

## Timezone And Timing Behavior

Wall-clock schedules require one of:

- inline DSL timezone, e.g. `every day at 08:00 in Europe/London`
- task-local `timezone`
- `getDefaultTimezone()`

Interval schedules intentionally do not accept timezone because they are duration based.

The runtime uses one-shot `Date` schedules for declared tasks instead of cron rows, which gives Think control over timezone behavior, late alarms, and interval drift.

Defined behavior:

- Nonexistent local time during DST spring-forward runs at the next valid local time after the gap.
- Repeated local time during DST fall-back runs once, at the first occurrence.
- Late alarms run the intended occurrence once and schedule the next future occurrence.
- Missed occurrences are not backfilled.
- Different intended occurrence timestamps may queue separately if previous work is still active.

## Example And Docs

This PR updates `examples/think-submissions` to show scheduled prompt tasks alongside manual durable submissions. The example keeps an hourly cadence to avoid surprising model usage, and the README explains how to shorten the interval locally for a quick demo.

The Think README and docs now cover:

- `getScheduledTasks()`
- prompt vs handler tasks
- handler context and idempotency expectations
- `getDefaultTimezone()`
- the supported DSL
- runtime validation
- no-backfill behavior
- optional `defineScheduledTasks()` usage
- dynamic reconciliation with `internal_reconcileScheduledTasks()`
- scheduled handler usage for Workflow starts

## Test Coverage

Adds coverage for:

- Runtime DSL validation and timezone validation.
- Type-level literal schedule checks.
- Prompt vs handler type exclusivity.
- Startup reconciliation.
- Action-only changes preserving schedule rows.
- Schedule/timezone changes replacing schedule rows.
- Removed tasks cancelling owned rows.
- Unrelated user-created schedules being preserved.
- Default timezone changes affecting wall-clock schedules.
- Late interval/no-backfill behavior.
- DST spring-forward and fall-back behavior.
- Idempotent scheduled execution through `submitMessages()`.
- Prompt and handler failures still re-arming future occurrences.
- Action-level retries for scheduled handlers.
- Pending `schedule_id` repair without duplicate task rows.
- Sub-agent scheduled task ownership and handler support.

## Test Plan

- `npx vitest run packages/think/src/tests/scheduled-tasks.test.ts --config packages/think/src/tests/vitest.config.ts`
- `npm run check`

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1585" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/tiny-ravens-schedule.md
+++ b/.changeset/tiny-ravens-schedule.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/think": minor
+---
+
+Add declarative scheduled tasks for Think agents with a typed recurring DSL, timezone-aware reconciliation, and durable idempotent submissions.
--- a/docs/think/index.md
+++ b/docs/think/index.md
@@ -141,6 +141,7 @@ driving the work and what the caller needs back.
 | A browser user sends chat messages                             | `useAgentChat` over the WebSocket chat protocol |
 | Server code can wait for the model response                    | `saveMessages()`                                |
 | Server code needs fast durable acceptance and later status     | `submitMessages()`                              |
+| Code should create recurring prompt-driven turns               | `getScheduledTasks()`                           |
 | Parent code needs direct streaming RPC to a specific child     | `subAgent(...).chat()`                          |
 | A parent agent delegates work to a retained child agent        | `agentTool()` or `runAgentTool()`               |
 | Surround a turn with idempotent app-owned side effects         | `startFiber()`                                  |
@@ -178,6 +179,8 @@ with retries per step, long waits, external events, or approvals.
 | `getModel()`            | throws                           | Return the `LanguageModel` to use                                               |
 | `getSystemPrompt()`     | `"You are a helpful assistant."` | System prompt (fallback when no context blocks)                                 |
 | `getTools()`            | `{}`                             | AI SDK `ToolSet` for the agentic loop                                           |
+| `getScheduledTasks()`   | `{}`                             | Code-declared recurring prompts                                                 |
+| `getDefaultTimezone()`  | `undefined`                      | Default timezone for wall-clock scheduled tasks                                 |
 | `maxSteps`              | `10`                             | Max tool-call rounds per turn                                                   |
 | `sendReasoning`         | `true`                           | Send reasoning chunks to chat clients                                           |
 | `configureSession()`    | identity                         | Add context blocks, compaction, search, skills — see [Sessions](../sessions.md) |
@@ -227,6 +230,49 @@ export class MyAgent extends Think<Env> {
   }
 }
 ```
+
+## Scheduled Tasks
+
+Use `getScheduledTasks()` when code should create recurring Think turns. Think
+reconciles the declarations on startup, stores a durable one-shot schedule for
+the next occurrence, and submits each occurrence through `submitMessages()` with
+a stable idempotency key.
+
+```typescript
+import { Think, defineScheduledTasks } from "@cloudflare/think";
+
+export class DigestAgent extends Think<Env> {
+  getDefaultTimezone() {
+    return "Europe/London";
+  }
+
+  getScheduledTasks() {
+    return defineScheduledTasks({
+      weeklyCommitReport: {
+        schedule: "every week on monday at 09:00",
+        prompt:
+          "Compile all my GitHub commits for the last week and send a concise summary."
+      },
+      workout: {
+        schedule: "every day at 08:00 in Europe/London",
+        prompt: "Start my workout."
+      }
+    });
+  }
+}
+```
+
+The DSL supports `every <n> minutes`, `every <n> hours`,
+`every day at HH:mm`, `every weekday at HH:mm`, and
+`every week on monday,wednesday at HH:mm`. Wall-clock schedules require either
+an inline timezone, a task `timezone`, or `getDefaultTimezone()`. If an alarm is
+late, Think runs the intended occurrence once and schedules the next future
+occurrence; it does not backfill missed runs.
+
+Use `getScheduledTasks()` for recurring prompt-driven turns. If your product
+lets users edit schedules at runtime, or a scheduled occurrence should start a
+deterministic Workflow rather than submit a prompt, keep that reconciliation in
+application code for now.
 
 ## Session Integration
 
--- a/docs/think/index.md
+++ b/docs/think/index.md
@@ -141,7 +141,7 @@ driving the work and what the caller needs back.
 | A browser user sends chat messages                             | `useAgentChat` over the WebSocket chat protocol |
 | Server code can wait for the model response                    | `saveMessages()`                                |
 | Server code needs fast durable acceptance and later status     | `submitMessages()`                              |
-| Code should create recurring prompt-driven turns               | `getScheduledTasks()`                           |
+| Code should create recurring prompt-driven turns or handlers   | `getScheduledTasks()`                           |
 | Parent code needs direct streaming RPC to a specific child     | `subAgent(...).chat()`                          |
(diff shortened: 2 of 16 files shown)


Konan69:
Nice, this is timely. I hand rolled something pretty similar recently, an "Automations/Workflows" feature and have been wondering how much of it should eventually move onto first class SDK/Think patterns as I learn the SDK
 better.

   The shape I have right now is roughly:

   - user-created schedules 
   - cron expression + timezone on the product row
   - `cron-parser` with `tz` to compute the next wall-clock occurrence
   - one-shot `this.schedule(nextDate, ...)`
   - re-arm after each fire
   - concurrency policy before dispatch (`skip` / `replace`)
   - then create a run ledger row and kick off a Cloudflare Workflow for the actual durable work

  Curious if this kind of product shape is something you imagine Think supporting first class
 eventually: product-managed schedules with a deterministic handler, not only scheduled prompts. In my case the scheduled occurrence is less "ask the model to do something" and more "deterministically start this Workflow/run that runs on think primitives".

   The two places where my current shape does not quite map are:

   ### Product-owned reconciliation

   Right now this looks startup-reconciled via a private method. In my case schedules are edited
 through the app, so after a DB write I need the live DO to notice the new desired state.

   The shape I keep wanting is something like:

   ```ts
   await this.reconcileScheduledTasks()
   ```

   where `getScheduledTasks()` can read from app state / DB and reconcile from there.

   ### Scheduled workflow/job, not only prompt

   Some scheduled work cant become a simple a model prompt. In my case the scheduled occurrence needs
 to deterministically create a run row and start a Workflow.

   The shape I keep imagine is something like:
``` ts
 getScheduledTasks() {
     return {
       dailyCustomerDigest: {
         schedule: "every day at 09:00",
         timezone: "America/New_York",
         handler: async ({ scheduledFor }) => {
           await this.automations.startScheduledRun({
             automationId: "customer-digest",
             triggerId: "daily-schedule",
             scheduledFor,
             // internally creates the run ledger row and starts/schedules the Workflow
             // e.g. await this.env.RUN_WORKFLOW.create({ id, params: { runId } })
           })
         },
       },
     }
   }
 ```

 or maybe an override hook:

 ```ts
   async onScheduledTask(task, occurrence) {
     await this.automations.startScheduledRun({
       taskId: task.id,
       scheduledFor: occurrence.scheduledFor,
       // internally creates the run ledger row and starts/schedules the Workflow
       // e.g. await this.env.RUN_WORKFLOW.create({ id, params: { runId } })
     })
   }
 ```
   The rest can stay app-owned pretty cleanly: concurrency policy, run history, idempotency keys,
 workflow payloads, etc. I mostly want to understand whether "user/product-managed scheduled
 workflows" fits where this abstraction is headed. thanks! :)

threepointone:
@Konan69 implemented a handler: option

PR #1596, opened 2026-05-27
feat(mcp): support caller-supplied stable ids in addMcpServer

### LINKED ISSUE #1564: mcp: support stable server ids in addMcpServer for connector-style integrations
We’re building a connector system where MCP servers map to first-class integrations like GitHub, Slack, Gmail, etc. Each integration has permissions, audit logs, capability metadata, and UI state keyed by a stable connector id.

`addMcpServer("github", binding, { props })` gets close, but the actual MCP server id is generated internally. For RPC transport, `serverName` becomes the RPC DO name (`rpc:github`), while the connection id is generated with `nanoid(8)` on first connect.

That generated id then appears in:

- `listServers()`
- `listTools()`
- `getAITools()`
- AI SDK tool namespacing, including Think’s automatic MCP tool merge

For connector systems, stable ids are useful so tool keys can stay tied to the integration:

```txt
tool_github_create_pull_request
tool_slack_send_message
```

instead of opaque connection ids.

Suggestion: add an optional caller-provided id:

```ts
await this.addMcpServer("GitHub", env.MCP_SESSION, {
  id: "github",
  props: { ... },
})
```

or:

```ts
await this.addMcpServer("GitHub", env.MCP_SESSION, {
  serverId: "github",
  props: { ... },
})
```

If omitted, current generated-id behavior stays unchanged. If provided, SDK uses that id for storage, restore, `listServers()`, `listTools()`, and `getAITools()` namespacing.

This would let connector-style apps use the high-level `addMcpServer()` lifecycle without maintaining a random-id → connector-id mapping layer or dropping down to lower-level `MCPClientManager` APIs.

mattzcarey:
Closes #1564.

## What

`addMcpServer` now accepts an optional stable `id` so connector-style apps can register MCP servers under predictable keys (`github`, `slack`, `gmail`) instead of opaque `nanoid(8)` connection ids.

```ts
await this.addMcpServer("GitHub", env.MCP_SESSION, {
  id: "github",
  props: { token: "..." }
});

// AI SDK tool keys become `tool_github_create_pull_request`
// instead of `tool_<random8chars>_create_pull_request`.
```

The supplied id flows through every place the SDK surfaces it: storage primary key, restore-after-hibernation, `listServers()` / `listTools()` / `getAITools()` tool-name namespacing, the `mcpConnections` map, and the OAuth `authProvider.serverId`.

Available on both overloads:

```ts
// HTTP
addMcpServer(name: string, url: string, options?: AddMcpServerOptions)
// RPC (Durable Object binding)
addMcpServer(name: string, binding: DurableObjectNamespace, options?: AddRpcMcpServerOptions)
```

If `id` is omitted, behaviour is unchanged (generated `nanoid(8)`).

## Normalization

Stable ids must be safe to use as a SQLite primary key **and** as part of an AI SDK tool name (which must match `/^[A-Za-z0-9_]+$/` after the SDK strips hyphens). Rather than make the caller think about both, the SDK normalizes the supplied id through a new exported `normalizeServerId` helper:

```ts
normalizeServerId("my-supplied-id");  // "my-supplied-id" (unchanged)
normalizeServerId("GitHub MCP!");     // "github-mcp"
normalizeServerId("42-things");       // "id-42-things"  (must start with a letter)
normalizeServerId("");                // "id"            (must be non-empty)
```

Rules: lowercase → replace runs of disallowed chars with `-` → collapse `-` → trim leading/trailing `-`/`_` → ensure leading letter → truncate to 64 chars. The function is idempotent and exported from `agents` and `agents/mcp` for callers that want to pre-compute the id (e.g. for `removeMcpServer`).

## Fully additive — no user code breaks

Adopting stable ids on an already-deployed agent must not require an explicit migration step. A user who upgrades:

```diff
- await this.addMcpServer("GitHub", env.MCP_SESSION);
+ await this.addMcpServer("GitHub", env.MCP_SESSION, { id: "github" });
```

…should not have to call `removeMcpServer` first, re-authenticate with OAuth, or lose tool-call state. So when the same `(name, url)` is already registered under a different id (typically an auto-generated nanoid from a previous call), `addMcpServer` calls a new `MCPClientManager.migrateServerId(oldId, newId, clientName)` which atomically renames the id across every place it's used as a key:

- the `cf_agents_mcp_servers` SQL row
- the in-memory `mcpConnections` map key
- the connection-disposables map key
- `authProvider.serverId` on the live connection
- OAuth-related Durable Object storage keys under `/${clientName}/${oldId}/...` (PKCE verifier, client info, tokens, state) — listed under the old prefix, re-put under the new prefix, then deleted from the old prefix in one batched `put` + one batched `delete`

The contract "the id you get back is the id you asked for" holds from the very first call, with no stale storage rows and no broken hibernation restore.

## When does it still throw?

Only on a genuinely ambiguous collision: the supplied stable id already belongs to a *different* `(name, url)` server. This is unrecoverable without the caller telling us which server they meant.

```
MCP server id "github" is already in use by server "GitHub Enterprise" (https://ghe.example.com/mcp). Stable ids must be unique per (name, url).
```

## Tests

- `packages/agents/src/tests/mcp/normalize-server-id.test.ts` — 12 unit tests covering happy path, normalization rules, idempotency, AI-SDK-tool-name safety after hyphen-strip, and truncation.
- `packages/agents/src/tests/mcp/add-rpc-mcp-server.test.ts` — new RPC integration tests:
  - stable id round-trips through storage
  - `"GitHub MCP!"` normalizes to `"github-mcp"`
  - JIT-migration: round-trips a real tool call through the migrated id and asserts no stale storage row remains, connection count unchanged
  - re-supplying the same stable id for the same `(name, url)` dedups (no regression on existing dedup)
  - collision against a different server throws
- `packages/agents/src/tests/mcp/http-dedup.test.ts` — new HTTP integration tests:
  - supplied id is used for the new server row
  - JIT-migration moves OAuth-style storage keys from `/${clientName}/${oldId}/...` to `/${clientName}/${newId}/...`
  - collision throws

All 433 MCP tests pass; `npm run check` clean.

## Out of scope

The lower-level `MCPClientManager.connect` / `registerServer` APIs already accept arbitrary ids; this PR only changes the high-level `addMcpServer` lifecycle.

### DIFF
--- a/.changeset/stable-mcp-server-ids.md
+++ b/.changeset/stable-mcp-server-ids.md
@@ -10,7 +10,7 @@ The supplied id is normalized via the exported `normalizeServerId` helper so tha
 
 **Fully additive — no user code breaks.** If you add `{ id: "github" }` to an existing `addMcpServer` call for a server that's already registered under an auto-generated nanoid, the SDK transparently migrates the existing storage row, in-memory connection, and OAuth-related DO storage keys to the new stable id. No `removeMcpServer` step required, no stale rows, no broken hibernation restore.
 
-`addMcpServer` only throws on a genuinely ambiguous collision: the same stable id already belongs to a *different* `(name, url)` server.
+`addMcpServer` only throws on a genuinely ambiguous collision: the same stable id already belongs to a _different_ `(name, url)` server.
 
 ```ts
 await this.addMcpServer("GitHub", env.MCP_SESSION, {
--- a/.changeset/stable-mcp-server-ids.md
+++ b/.changeset/stable-mcp-server-ids.md
@@ -1,0 +1,19 @@
+---
+"agents": minor
+---
+
+Support stable, caller-supplied server ids in `addMcpServer` for connector-style integrations.
+
+Both the HTTP and RPC overloads of `addMcpServer` now accept an optional `id` field on their options object. When provided, this id replaces the generated `nanoid(8)` as the server's id in storage, restore, `listServers()`, `listTools()`, `getAITools()` (so tool keys become e.g. `tool_github_create_pull_request` instead of opaque connection ids), and OAuth state.
+
+The supplied id is normalized via the exported `normalizeServerId` helper so that values like `"GitHub MCP!"` become `"github-mcp"` — guaranteeing the id is safe to embed in AI SDK tool names and storage keys. If a caller-supplied id already maps to a server with a different name or url, `addMcpServer` now throws instead of silently overwriting the existing row.
+
+```ts
+await this.addMcpServer("GitHub", env.MCP_SESSION, {
+  id: "github",
+  props: { token: "..." }
+});
+// tools surface as `tool_github_<name>`
+```
+
+Closes #1564.
--- a/.changeset/stable-mcp-server-ids.md
+++ b/.changeset/stable-mcp-server-ids.md
@@ -1,5 +1,5 @@
 ---
-"agents": minor
+"agents": patch
 ---
 
 Support stable, caller-supplied server ids in `addMcpServer` for connector-style integrations.
--- a/.changeset/stable-mcp-server-ids.md
+++ b/.changeset/stable-mcp-server-ids.md
@@ -8,10 +8,9 @@ Both the HTTP and RPC overloads of `addMcpServer` now accept an optional `id` fi
 
 The supplied id is normalized via the exported `normalizeServerId` helper so that values like `"GitHub MCP!"` become `"github-mcp"` — guaranteeing the id is safe to embed in AI SDK tool names and storage keys.
 
-`addMcpServer` now throws explicitly when a stable id would conflict with existing storage:
+**Fully additive — no user code breaks.** If you add `{ id: "github" }` to an existing `addMcpServer` call for a server that's already registered under an auto-generated nanoid, the SDK transparently migrates the existing storage row, in-memory connection, and OAuth-related DO storage keys to the new stable id. No `removeMcpServer` step required, no stale rows, no broken hibernation restore.
 
-- the supplied id already belongs to a different `(name, url)` server, or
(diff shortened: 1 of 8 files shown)


threepointone:
1. packages/agents/src/index.ts:R9462-9479

**Dedup path returns existing server ID, ignoring caller-supplied requestedId**

When an existing server is found by name+url dedup AND has an active connection (`packages/agents/src/index.ts:9462-9479`), the code returns `existingServer.id` without checking whether it matches the caller's `requestedId`. This means a caller who provides `{ id: "github" }` could receive back a different auto-generated nanoid if the server was previously connected without a stable id. The returned id silently differs from the requested one.

This only manifests during a transition: (1) server first connected without a stable id → gets nanoid, (2) same server still has an active connection, (3) user upgrades code to pass a stable id. After hibernation and reconnection the stable id takes effect normally. The pre-existing dedup logic always returned the existing server, so this isn't a regression — but the new stable-id contract makes the mismatch more visible to callers who now expect a specific id back.

2. packages/agents/src/index.ts:R9490-9492

**Stale storage row when transitioning from auto-generated to stable IDs post-hibernation**

In the RPC path (`packages/agents/src/index.ts:9492`), `reconnectId = requestedId ?? existingServer?.id`. If both are defined but different (e.g., existingServer has nanoid `abc123` from a prior call without stable id, now `requestedId` is `"github"`), `requestedId` wins. A new connection is created with id `"github"` and saved via `INSERT OR REPLACE`. However, the old row with id `abc123` remains in storage because the replace is keyed on the new id.

On subsequent restores via `_restoreRpcMcpServers` (`packages/agents/src/index.ts:9111-9151`), both rows are iterated, potentially creating duplicate connections to the same underlying MCP server. In practice, the second restore's `addMcpServer` call would likely dedup against the first restore's active connection, but the stale row persists indefinitely. This is a narrow edge case (only during a one-time migration from auto-generated to stable ids after hibernation) with low practical impact.

threepointone (approved):

PR #1602, opened 2026-05-28
fix(mcp): SSE resumability for McpAgent + correct keepalive for stateless WorkerTransport (#1583)

### LINKED ISSUE #1583: SSE keepalive missing / broken in McpSession — streams drop every ~5 min on CF edge
**Describe the bug**

MCP sessions served via `McpSession.serve('/mcp')` drop every ~5 minutes on the Cloudflare edge due to three defects in the keepalive implementation across two code paths in `dist/mcp/index.js`.

**To Reproduce**

1. Serve an MCP server via `McpSession.serve('/mcp')` on a Cloudflare Durable Object
2. Open an SSE connection and leave it idle (no tool calls)
3. After ~5 minutes the stream closes
4. Workers logs show `waitUntil() tasks did not complete` warnings at each drop boundary

**Expected behavior**

SSE streams remain open indefinitely. The CF edge sees periodic keepalive bytes and does not invoke its idle-stream watchdog.

**Screenshots**

N/A

**Version**

agents@0.12.4

**Additional context**

Three defects combine to cause this:

1. **`createStreamingHttpHandler` has no keepalive at all.** `McpSession.serve()` routes through this handler. Its GET path opens the SSE channel and never writes another byte, so the CF edge 5-minute idle watchdog closes it.

2. **`WorkerTransport`'s `setInterval` is not wrapped in `ctx.waitUntil`.** The interval body runs bare. The Workers runtime cancels background work ~30s after the fetch handler returns, so the keepalive fires a few times then silently dies.

3. **Wrong SSE frame format.** The interval writes `event: ping\ndata: \n\n` (a named event) instead of the SSE comment form `: ping\n\n`. Named events trigger `onmessage` / event listeners on the client — unintended for a keepalive frame.

The fix for all three sites is the same pattern — wrap the tick in a `ctx.waitUntil`-guarded IIFE, use `: ping\n\n`, and clear the interval on stream close/error. A 25s interval gives comfortable headroom below both the 30s runtime cancellation window and the 5-minute edge watchdog.

We are currently carrying this as a `pnpm patch` against the compiled output:

```diff
diff --git a/dist/mcp/index.js b/dist/mcp/index.js
index 06a45a44f72216d24e12ca8d6cafa44456e30d18..81ef9096f58069a5fe7ce52e1501b1628d044c2c 100644
--- a/dist/mcp/index.js
+++ b/dist/mcp/index.js
@@ -271,6 +271,15 @@ const createStreamingHttpHandler = (basePath, namespace, options = {}) => {
                                        return new Response("Failed to establish WS to DO", { status: 500 });
                                }
                                ws.accept();
+                               const keepAlive = setInterval(() => {
+                                       ctx.waitUntil((async () => {
+                                               try {
+                                                       await writer.write(encoder.encode(": ping\n\n"));
+                                               } catch {
+                                                       clearInterval(keepAlive);
+                                               }
+                                       })());
+                               }, 25_000);
                                ws.addEventListener("message", (event) => {
                                        try {
                                                async function onMessage(ev) {
@@ -285,9 +294,11 @@ const createStreamingHttpHandler = (basePath, namespace, options = {}) => {
                                        }
                                });
                                ws.addEventListener("error", () => {
+                                       clearInterval(keepAlive);
                                        writer.close().catch(() => {});
                                });
                                ws.addEventListener("close", () => {
+                                       clearInterval(keepAlive);
                                        writer.close().catch(() => {});
                                });
                                return new Response(readable, {
@@ -911,12 +922,8 @@ var WorkerTransport = class {
                });
                if (this.sessionId !== void 0) headers.set("mcp-session-id", this.sessionId);
                const keepAlive = setInterval(() => {
-                       try {
-                               writer.write(encoder.encode("event: ping\ndata: \n\n"));
-                       } catch {
-                               clearInterval(keepAlive);
-                       }
-               }, 3e4);
+                       writer.write(encoder.encode(": ping\n\n")).catch(() => clearInterval(keepAlive));
+               }, 25_000);
                this.streamMapping.set(streamId, {
                        writer,
                        encoder,
@@ -1097,12 +1104,8 @@ var WorkerTransport = class {
                });
                if (this.sessionId !== void 0) headers.set("mcp-session-id", this.sessionId);
                const keepAlive = setInterval(() => {
-                       try {
-                               writer.write(encoder.encode("event: ping\ndata: \n\n"));
-                       } catch {
-                               clearInterval(keepAlive);
-                       }
-               }, 3e4);
+                       writer.write(encoder.encode(": ping\n\n")).catch(() => clearInterval(keepAlive));
+               }, 25_000);
                this.streamMapping.set(streamId, {
                        writer,
                        encoder,
```

---

## Addendum — POST branch of `createStreamingHttpHandler` was also missing keepalive

The patch above covered three sites:

- `createStreamingHttpHandler` GET branch
- `WorkerTransport` (two `setInterval` sites)

In production we then observed the issue recurring for clients that use the Streamable HTTP **request-response** mode (Claude Code, web chat agents — anything calling `POST /mcp` for tool execution). The POST branch of `createStreamingHttpHandler` (around line 175 in the unpatched 0.12.4 source) is structurally identical to the GET branch but had no keepalive at all.

Symptom: every `POST /mcp` for a tool call > ~5 minutes (e.g. a slow tool call against a reasoning model) hits Cloudflare's edge idle timeout and drops the SSE stream. The client sees the connection die and reinitialises the MCP session, which surfaces to users as "auth keeps dying" since reinit re-runs the OAuth flow on some clients.

The fix is structurally identical: a `setInterval` writing `: ping\n\n` every 25s, cleared in 5 sites — `ws.addEventListener('message')` when `message.close === true`, `ws.addEventListener('error')`, `ws.addEventListener('close')`, the notifications-only early-return, and the `setInterval` itself.

Note: the POST branch's `setInterval` omits the `ctx.waitUntil` wrapper used in the GET branch. The POST handler returns a streaming `Response` whose body lifecycle keeps the Worker alive through the entire stream, so the 30s background-task cancellation window doesn't apply (the runtime considers the response body to be the foreground work).

Additional hunks for the same `pnpm patch`:

```diff
@@ -175,6 +175,9 @@ const createStreamingHttpHandler = (basePath, namespace, options = {}) => {
                                        return new Response(body, { status: 500 });
                                }
                                ws.accept();
+                               const keepAlive = setInterval(() => {
+                                       writer.write(encoder.encode(": ping\n\n")).catch(() => clearInterval(keepAlive));
+                               }, 25_000);
                                ws.addEventListener("message", (event) => {
                                        async function onMessage(event) {
                                                try {
@@ -183,6 +186,7 @@ const createStreamingHttpHandler = (basePath, namespace, options = {}) => {
                                                        if (message.type !== "cf_mcp_agent_event") return;
                                                        await writer.write(encoder.encode(message.event));
                                                        if (message.close) {
+                                                               clearInterval(keepAlive);
                                                                ws?.close();
                                                                await writer.close().catch(() => {});
                                                        }
@@ -194,17 +198,20 @@ const createStreamingHttpHandler = (basePath, namespace, options = {}) => {
                                });
                                ws.addEventListener("error", (error) => {
                                        async function onError(_error) {
+                                               clearInterval(keepAlive);
                                                await writer.close().catch(() => {});
                                        }
                                        onError(error).catch(console.error);
                                });
                                ws.addEventListener("close", () => {
                                        async function onClose() {
+                                               clearInterval(keepAlive);
                                                await writer.close().catch(() => {});
                                        }
                                        onClose().catch(console.error);
                                });
                                if (messages.every((msg) => isJSONRPCNotification(msg) || isJSONRPCResultResponse(msg))) {
+                                       clearInterval(keepAlive);
                                        ws.close();
                                        return new Response(null, {
                                                headers: corsHeaders(request, options.corsOptions),
```

mattzcarey:
Closes #1583.

## Summary

Two-commit fix for SSE keepalive + resumability across both MCP transports.

The policy is now split by stream direction, consistently across `McpAgent` (`utils.ts`) and `createMcpHandler` (`worker-transport.ts`):

| Stream | Keepalive? | Recovery on idle drop |
|---|---|---|
| **GET** standalone listen stream | No | Client reconnects with `Last-Event-ID` against a configured `EventStore` |
| **POST** tool response stream | Yes — `: keepalive\n\n` every 25s | None (POST is scoped to a request id and can't be resumed) |

| Commit | What it does |
|---|---|
| `8b0c30e2` | Default-enable SSE resumability for `McpAgent`: new `DurableObjectEventStore` (storage-backed, hibernation-safe), `McpAgent.getEventStore()` hook, wired into `StreamableHTTPServerTransport` (the `DurableObjectEventStore` itself is `@internal`, callers override `getEventStore()` to swap or disable). Also fixes a pre-existing transport bug where resumed GET streams were never re-tagged as the standalone SSE connection. |
| `97f238da` | Replace the broken `event: ping` keepalive in both transports with a spec-compliant `: keepalive\n\n` comment frame, armed only on POST response streams. GET streams now always rely on resumability. Shared helper in `sse-keepalive.ts`. |

### Real bugs (user-visible)

- **`McpAgent.serve()` standalone GET SSE drops at ~270s with no recovery.** Reproduced on `agents@0.13.2` against a deployed worker — the stream dies at the Cloudflare edge idle watchdog and there is no resumability wired in. Fixed by defaulting `EventStore` on for `McpAgent`. (Commit 1)
- **`McpAgent`'s `handleGetRequest` returned early after `replayEvents()` without tagging the reconnected connection as the standalone SSE stream.** Resumed streams received the backlog and then sat idle — server-initiated notifications had no connection to land on. Pre-existing bug, not in the original report. Fixed by tagging *before* replay. (Commit 1)
- **`McpAgent`'s POST tool-response SSE had no keepalive.** A long-running tool that didn't emit progress would let the response stream sit silent past the ~5min edge watchdog and the client would never get the result. POST streams are scoped to a single request id and can't be replayed, so resumability isn't an option — the only fix is to keepalive them. Fixed in commit 2.
- **Keepalive frame format `event: ping\ndata: \n\n`** is a named SSE event. Per the WHATWG SSE algorithm a `data:` line with empty value still appends a line feed to the data buffer, so the buffer is non-empty at dispatch time and an event with `type="ping"` and empty data fires on the client. `onmessage` doesn't fire (wrong type), but any `addEventListener("ping", …)` listener does. The MCP TS SDK client skips these (it discards events with empty data) but other SSE clients see them. Fixed by switching to the comment frame `: keepalive\n\n`. (Commit 2)

### Correctness (spec alignment, no behaviour change measured)

- **Interval period 30s on the 30s cancellation boundary.** No safety margin. Moved to 25s, which gives headroom below both the post-handler cancellation window and the ~5min edge watchdog. The WHATWG SSE spec recommends "every 15 seconds or so"; 25s is plenty. (Commit 2)
- **Stateless `createMcpHandler` keepalive cancellation.** The reporter argued the Workers runtime cancels the bare `setInterval` ~30s after the fetch handler returns. **I could not reproduce this** — a 330s `slow` tool call survived end-to-end against an unfixed `agents@0.13.2` baseline. The open SSE response itself counts as foreground I/O and keeps the worker (and therefore the interval) alive. The frame format and interval period fixes are still correct on their own merits.

### Cleanup

- **Shared `sse-keepalive.ts` helper.** Both transports were duplicating the same keepalive logic. Now exported once.

- **JSDoc on `WorkerTransportOptions.eventStore`** now explains the GET-vs-POST split and the BYO recipe.

## What's deliberately out of scope

- **Default `eventStore` for stateless `createMcpHandler`.** Stateless deployments can't ship a sensible default — there's no per-session storage handle available without the caller bringing one. Callers who want resumability supply their own store via `WorkerTransportOptions.eventStore`. Stateless callers get the corrected POST keepalive by default.
- **`ctx.waitUntil` threading.** Considered briefly to anchor the keepalive against post-handler background-work cancellation, but I couldn't reproduce that scenario empirically and the threading muddied the public API. The streaming SSE response itself keeps the worker alive in practice.
- **Anything for the next-draft MCP spec.** This whole code path is being replaced when we target the draft (which removes the GET stream and `Last-Event-ID` entirely). All of these additions delete cleanly together.

## Architecture

This sticks to serverless principles:

- GET streams hibernate-friendly: the DO is allowed to idle between events, the edge is allowed to close the stream, the client reconnects with `Last-Event-ID` and the `EventStore` (backed by `this.ctx.storage`) replays missed messages. No artificial keepalive holding the DO awake.
- POST streams are short-lived by nature (single request id, terminates with the response). For the long-running edge cases, the comment-frame keepalive keeps the response alive without firing spurious client-side events.

## Compatibility

All changes are **additive** — patch-level, no breaking changes:

- `McpAgent.getEventStore()` is overridable (`return undefined` to opt out of the default store).
- No public API signatures changed.

## Verification

- 430/430 MCP tests pass.
- Deployed repro workers (now torn down) confirmed:
  - Baseline `McpAgent`: GET drops at 271s with `INTERNAL_ERROR`, zero bytes received.
  - Fixed `McpAgent`: GET still drops at 271s (DO is free to hibernate), and a follow-up GET with `Last-Event-ID` returns 200 OK. Tool-call POST frames now carry `id: <streamId>:<seqHex>`.
  - Baseline `createMcpHandler` 330s tool call: completed, but emitted 10 spurious `event: ping` frames.
  - Fixed `createMcpHandler` 330s tool call: completed, emitted ~12 `: keepalive` comment frames the SSE parser discards.

## New tests

- `mcp/event-store.test.ts` — 10 unit tests for `DurableObjectEventStore` against a mock `DurableObjectStorage` (id format, per-stream isolation, eviction, hibernation rehydration, no timers).
- `mcp/resumability.test.ts` — 5 e2e tests through the worker entry exercising the `McpAgent` resumability path.
- `mcp/transports/post-keepalive.test.ts` rewritten — asserts GET never arms a keepalive (both with and without `eventStore`) and POST always does, in both modes.

### DIFF
--- a/packages/agents/src/mcp/index.ts
+++ b/packages/agents/src/mcp/index.ts
@@ -573,8 +573,6 @@ export {
 
 export { getMcpAuthContext, type McpAuthContext } from "./auth-context";
 
-export { DurableObjectEventStore } from "./event-store";
-
 export {
   WorkerTransport,
   type WorkerTransportOptions,
--- a/packages/agents/src/mcp/index.ts
+++ b/packages/agents/src/mcp/index.ts
@@ -194,27 +194,28 @@ export abstract class McpAgent<
   }
 
   /**
-   * Maximum age (in milliseconds) of an event in the SSE event store. Events
-   * older than this are dropped by the periodic sweep scheduled in
-   * {@link onStart}. Default 1 hour — plenty of room for a client to
-   * reconnect with `Last-Event-ID`, short enough to bound storage growth
-   * from abandoned POST streams whose clients never returned.
+   * Maximum age (in milliseconds) of an event in the SSE event store.
+   * Events older than this are dropped by the periodic sweep scheduled in
+   * {@link onStart}. Default 24 hours — generous enough for clients to
+   * reconnect with `Last-Event-ID` after long pauses, while still bounding
+   * storage growth from abandoned POST streams whose clients never
+   * returned.
    *
    * Override (in conjunction with {@link getEventStore}) to customise.
    * Return `Infinity` to disable the sweep.
    */
   protected getEventStoreMaxAgeMs(): number {
-    return 60 * 60 * 1000; // 1 hour
+    return 24 * 60 * 60 * 1000; // 24 hours
   }
 
   /**
    * Cron expression for the recurring sweep that prunes expired SSE events
-   * from the default {@link DurableObjectEventStore}. Default: every 5
-   * minutes. Override to change the cadence; return `undefined` to disable
-   * scheduling entirely.
+   * from the default {@link DurableObjectEventStore}. Default: every hour,
+   * at minute 0. Override to change the cadence; return `undefined` to
+   * disable scheduling entirely.
    */
   protected getEventStoreSweepCron(): string | undefined {
-    return "*/5 * * * *";
+    return "0 * * * *";
   }
 
   /**
--- a/packages/agents/src/mcp/index.ts
+++ b/packages/agents/src/mcp/index.ts
@@ -25,6 +25,8 @@ import {
   MCP_MESSAGE_HEADER
 } from "./utils";
 import { McpSSETransport, StreamableHTTPServerTransport } from "./transport";
+import type { EventStore } from "@modelcontextprotocol/sdk/server/streamableHttp.js";
+import { DurableObjectEventStore } from "./event-store";
 import { RPCServerTransport, type RPCServerTransportOptions } from "./rpc";
 
 export abstract class McpAgent<
@@ -124,6 +126,21 @@ export abstract class McpAgent<
     return {};
   }
 
+  /**
+   * Returns the {@link EventStore} used for SSE resumability on the streamable
+   * HTTP transport. Defaults to a {@link DurableObjectEventStore} backed by
+   * this agent's storage, which lets clients reconnect with `Last-Event-ID`
+   * after the Cloudflare edge closes an idle SSE stream (~5 minute watchdog)
+   * instead of relying on a server-side keepalive that would block DO
+   * hibernation.
+   *
+   * Override to disable resumability (`return undefined`) or to plug in a
+   * different store.
+   */
+  protected getEventStore(): EventStore | undefined {
+    return new DurableObjectEventStore(this.ctx.storage);
+  }
+
   /** Returns a new transport matching the type of the Agent. */
   private initTransport() {
     switch (this.getTransportType()) {
@@ -131,7 +148,9 @@ export abstract class McpAgent<
         return new McpSSETransport();
       }
       case "streamable-http": {
-        const transport = new StreamableHTTPServerTransport({});
+        const transport = new StreamableHTTPServerTransport({
+          eventStore: this.getEventStore()
+        });
         transport.messageInterceptor = (message) => {
           return Promise.resolve(this._handleElicitationResponse(message));
         };
@@ -553,6 +572,8 @@ export {
 } from "./handler";
 
 export { getMcpAuthContext, type McpAuthContext } from "./auth-context";
+
+export { DurableObjectEventStore } from "./event-store";
 
 export {
   WorkerTransport,
--- a/packages/agents/src/mcp/index.ts
+++ b/packages/agents/src/mcp/index.ts
@@ -66,6 +66,58 @@ export abstract class McpAgent<
     return this.ctx.storage.get<JSONRPCMessage>("initializeRequest");
   }
 
+  /**
+   * Storage key prefix for the `streamId -> requestIds` mapping used to
+   * support POST stream resumption across WebSocket reconnects. See
+   * {@link StreamableHTTPServerTransport.handleGetRequest}.
+   *
+   * @internal
+   */
+  private static readonly STREAM_REQS_KEY_PREFIX = "__mcp_stream_reqs__:";
+
+  /**
+   * Persist the `requestIds` belonging to a POST tool-call stream so a
+   * future GET reconnect (carrying `Last-Event-ID`) can restore them
+   * onto a fresh WebSocket connection. Internal — used by the streamable
+   * HTTP transport.
+   *
+   * @internal
+   */
+  async setStreamRequestIds(
+    streamId: string,
+    requestIds: RequestId[]
+  ): Promise<void> {
+    await this.ctx.storage.put<RequestId[]>(
+      `${McpAgent.STREAM_REQS_KEY_PREFIX}${streamId}`,
+      requestIds
+    );
+  }
+
+  /**
+   * Read the persisted `requestIds` for a POST stream, or `undefined`
+   * if the stream has already completed (or never existed). Internal.
+   *
+   * @internal
+   */
+  async getStreamRequestIds(
+    streamId: string
+  ): Promise<RequestId[] | undefined> {
+    return this.ctx.storage.get<RequestId[]>(
+      `${McpAgent.STREAM_REQS_KEY_PREFIX}${streamId}`
+    );
+  }
+
+  /**
+   * Drop the persisted `requestIds` for a POST stream. Internal.
+   *
+   * @internal
+   */
+  async deleteStreamRequestIds(streamId: string): Promise<void> {
+    await this.ctx.storage.delete(
+      `${McpAgent.STREAM_REQS_KEY_PREFIX}${streamId}`
+    );
+  }
+
   /** Read the transport type for this agent.
    * This relies on the naming scheme being `sse:${sessionId}`,
    * `streamable-http:${sessionId}`, or `rpc:${sessionId}`.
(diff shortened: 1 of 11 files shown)


packages/agents/src/mcp/index.ts
```diff
@@ -124,14 +126,31 @@ export abstract class McpAgent<
...
   /** Returns a new transport matching the type of the Agent. */
   private initTransport() {
     switch (this.getTransportType()) {
       case "sse": {
         return new McpSSETransport();
       }
       case "streamable-http": {
-        const transport = new StreamableHTTPServerTransport({});
+        const transport = new StreamableHTTPServerTransport({
+          eventStore: this.getEventStore()
```
whoiskatrin, line 245: wondering if this could keep adding stuff to storage over time

the cap is per stream, but it looks like every POST gets a new connection id. so most tool calls would add a couple events to a new stream, never hit the cap, and i can't see anything removing them after the response is done

now that this is on by default for McpAgent, feels like a busy session could just keep building up old POST response events in DO storage

should finished POST streams be cleared, or should there be some cap for the whole session?

packages/agents/src/mcp/worker-transport.ts
```diff
@@ -348,23 +363,18 @@ export class WorkerTransport implements Transport {
...
-    this.streamMapping.set(streamId, {
-      writer,
-      encoder,
-      cleanup: () => {
-        clearInterval(keepAlive);
-        this.streamMapping.delete(streamId);
-        writer.close().catch(() => {});
-      }
-    });
+    // No keepalive on the standalone GET stream. Idle drops are recovered
```
whoiskatrin: could this break WorkerTransport setups that don't pass an eventStore?

before this, the GET stream stayed open because of the keepalive. now it can get dropped when idle, but without an eventStore there aren't any ids or replay to recover from that

i think the mcp elicitation example might hit this too since it has a persistent WorkerTransport with storage but no eventStore

feels like GET should keep the new comment keepalive when there isn't an eventStore, and only rely on reconnecting once one is set

packages/agents/src/mcp/worker-transport.ts
```diff
@@ -76,8 +77,22 @@ export interface WorkerTransportOptions {
...
+   * Event store for SSE resumability.
+   *
+   * When set, the transport assigns a globally-unique `id:` to each SSE
+   * event and replays missed events when a client reconnects with the
+   * `Last-Event-ID` header. This is the supported recovery path for the
+   * standalone GET stream, which is allowed to drop at the Cloudflare
+   * edge idle watchdog (~5 min) and never carries a server-side
+   * keepalive.
+   *
+   * POST response streams are scoped to a single request id and can't be
```
whoiskatrin: not sure about saying POST streams can't be resumed here

i think with streamable HTTP, if the POST SSE connection drops, the client can reconnect with GET and Last Event ID and get events from that original stream replayed. this code also seems to do that already since it stores the POST events

would be clearer to say the keepalive stops an in progress tool call from getting dropped, instead of saying POST can't be resumed

packages/agents/src/mcp/index.ts
mattzcarey: POST streams shouldnt add to storage. Should just be GETs. POST needs the keep alive for this reason. Will double check

packages/agents/src/mcp/worker-transport.ts
mattzcarey: the keepalive on the GET is just burning cash tho. it should only be for notifications as is being removed real soon. Elicitations go down the POST for the tool call that caused them. If they dont thats a bug that users need to fix.

packages/agents/src/mcp/worker-transport.ts
mattzcarey: will check this. I'm not sure you can resume POST messaged down the reconnected GET cause now they would be orphaned with respect to the tool call they were for.

packages/agents/src/mcp/index.ts
mattzcarey: Adding a ttl or some type of cleanup. good shout.

packages/agents/src/mcp/worker-transport.ts
mattzcarey: > feels like GET should keep the new comment keepalive when there isn't an eventStore, and only rely on reconnecting once one is set

will add this this tho. Nice to keep it a patch change.

whoiskatrin (approved):
Overall looks pretty solid, just a few comments above

threepointone (approved):
feel free to land after CI passes

PR #1605, opened 2026-05-28
fix(voice): support AI SDK fullStream tool-call responses

### LINKED ISSUE #1591: fix: websocket client buffers text between tool calls instead of streaming
## Problem

The websocket client doesn't receive text segments that appear between tool calls until the entire stream ends. This causes the voice to read all buffered text at once at the end, instead of reading text as it arrives, going silent during tool calls, and then continuing.

## Expected behavior

Text should be delivered to the client immediately as it is generated, even when interleaved with tool calls:

1. Agent sends intro text → client receives and voice reads it right away  
2. Agent makes a tool call → voice goes silent  
3. Tool returns, agent sends follow-up text → client receives and voice reads it  
4. Agent makes another tool call → voice goes silent  
5. Agent sends conclusion → client receives and voice reads it

## Actual behavior

All text segments are buffered on the server side and only flushed to the client when the stream completes. The voice then reads the entire message in one go.

cjol:
This PR updates the voice pipeline to handle AI SDK `fullStream` responses correctly when model output is split by tool calls.

## Problem
  - AI SDK `textStream` only yields text deltas, so it drops the structure that shows when text was separated by tool calls.
  - For responses shaped like `text1 + tool + text2`, `textStream` can concatenate non-adjacent text without spacing, producing transcripts like `text1text2`.
  - The voice TTS pipeline also only flushed pending text when it found a sentence boundary or reached the end of the stream, so pre-tool speech could remain buffered until post-tool text started.

### Bug demo

This is the example voice app with a 5-sec delay added to tool calls.

https://github.com/user-attachments/assets/ed916954-2cdf-42f2-bb86-9c7479530ffe

Note that the text transcript is missing a space between the two parts of the message, and the voice output doesn't begin until after the final text is emitted.

## Solution
  - Accept AI SDK `fullStream` parts from `onTurn()` and extract `text-delta` events for the existing transcript and TTS flow.
  - Preserve non-text `fullStream` events as internal stream boundaries so tool calls become flush points for pending TTS text.
  - Insert a boundary space when separate `fullStream` text sections are split by tool events and neither side already provides whitespace.
  - Keep deprecated `textStream` support for existing users, but warn that `fullStream` should be returned instead.

### Fix Demo

https://github.com/user-attachments/assets/b46b1cff-81c4-44da-b7a8-434892bda3fb

## User-facing changes:
  - Voice docs and examples now return `result.fullStream` from AI SDK `streamText()` calls.
  - `onTurn()` documentation now describes returning a string, AI SDK `fullStream`, or other supported text-producing streams.
  - Returning a text stream from `onTurn` logs a warning recommending use of `fullStream` instead.
  - A patch changeset is included for `@cloudflare/voice`.

## Test coverage:
  - Adds mocked AI SDK tool-call streams for both `fullStream` and legacy `textStream` behavior.
  - Covers correct spacing for `fullStream` text separated by tool calls.
  - Covers the known legacy `textStream` concatenation behavior so compatibility remains intentional.
  - Covers boundary-event ordering and verifies pre-tool audio is sent before a delayed tool result completes.

Fixes #1591 

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1605" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


mattzcarey (approved):

mattzcarey:
I got this clanker review for ya @cjol I didnt have time to test this myself:

 - isTextBoundary() currently treats almost every non-text event
   as a speech boundary. That includes sources, files, reasoning,
   raw metadata, aborts, and errors—not only tool transitions.
   This can:
     - flush TTS unexpectedly;
     - insert spaces into otherwise adjacent text;
     - silently treat AI SDK error events as successful
       completion.
 - Tool boundary behavior needs spacing tests for:
     - existing trailing/leading whitespace;
     - repeated tool events;
     - multiple tool calls;
     - no text before/after a tool;
     - non-tool metadata between text deltas.
 - The added test agents explicitly disable hibernation.
   Production voice agents default to hibernation, although
   withVoice acquires a keepAlive() lease for an active call. A
   deployed cold-wake/default-hibernation check is worthwhile.
 - design/voice.md still documents the old onTurn() return
   contract and stale hibernation implementation details.

PR #1607, opened 2026-05-28
fix(mcp): replay every event after POST stream interruption (#1583 follow-up)

mattzcarey:
Follow-up to #1602. Fixes the correctness bugs on the streamable-HTTP resumability path and replaces the cleanup strategy with immediate clear-on-close — no background sweep, no alarms, quiescent when idle.

### Final tool response is now replayable

The previous code stored the final response in the event store and then immediately called `clearStream(streamId)` on `shouldClose`, deleting every event for that stream — including the one just written. A client that lost the connection mid-flight could reconnect with `Last-Event-ID` and find nothing to replay.

Fixed by ordering: write the SSE frame to the wire **first**, then clear. Every event up to and including the final response stays replayable while the in-flight stream is open. Trade-off: if the WS message is enqueued but the client TCP dies before the bytes arrive, that one final message is lost — accepted in exchange for not running any background cleanup.

### Request-bound `send()` writes are unconditional

The standalone branch already stored events regardless of whether a live connection was attached. The request-bound branch only stored when a live connection existed. Long-running tool calls on flaky networks routinely lose the POST connection mid-stream, and those messages were silently disappearing.

`McpAgent.getStreamForRequestId()` does a reverse lookup against `__mcp_stream_reqs__:*`, returning both the `streamId` and the stream's `requestIds` in one pass, so the transport can resolve a stream even after the originating WS has died. The event still gets stored; only the wire write is skipped when there's no live connection.

### Resumed GET registers under the source streamId, and supersedes stale connections

When a client reconnects with `Last-Event-ID`:

- **Active POST**: persisted `requestIds` are restored so future tool messages route to the new WS.
- **Standalone listen stream**: connection takes over that role under `STANDALONE_STREAM_ID`.
- **Completed POST or unknown**: one-shot replay channel.

In every resumable case (and when a fresh standalone GET opens) any prior connection bound to the same streamId is closed, so there's at most one live connection per stream. This kills the routing race where `send()` could pick a stale POST bridge over the resumed GET, and keeps us within the MCP rule of one stream per message.

### Spec: one stream per message

MCP 2025-06-18 "Multiple Connections": the server **MUST** send each JSON-RPC message on only one connected stream and **MUST NOT** broadcast across streams. Server-initiated notifications go to the single standalone GET; POST responses go to their own stream. Combined with the supersede-on-resume behaviour above, there is never more than one standalone GET to choose from.

## Cleanup architecture

Immediate, not background. Each POST stream's events are cleared the moment the close frame is written — no alarms, no metadata index, no sweep. Idle DOs do **zero** periodic work.

Storage cost is bounded by the in-flight POST streams plus the standalone GET stream. Defensive bounds:

- `clearStream` lists + deletes in chunks of 128 (the DO multi-key delete cap), so a large stream never loads the whole event log into memory.
- `replayEventsAfter` caps each call at 1000 events (`start: <lastEventId>\x00` for strict-after, exclusive replay).
- `getStreamForRequestId` caps its scan at 1000 keys and warns if hit (signals leaked `__mcp_stream_reqs__` entries).
- `storeEvent` rejects `streamId`s containing `:` so custom embedder ids can't cause prefix-scan collisions.

Standalone GET events are not cleared automatically; they accumulate for the lifetime of the session's Durable Object (bounded by session length).

## Exports

`DurableObjectEventStore` and the `ClearableEventStore` interface are exported from `agents/mcp` so callers embedding `WorkerTransport` inside an Agent or Durable Object can opt into resumability with `new DurableObjectEventStore(this.ctx.storage)`.

## Verified against the MCP spec + SDK reference

- One stream per message; no broadcast (spec "Multiple Connections"). ✅
- Event ids globally unique across streams in a session. ✅
- Replay is per-stream and exclusive of `Last-Event-ID`; never cross-stream. ✅
- Request-bound storage outlives WS drops via persistent reverse lookup (mirrors SDK `_requestToStreamMapping`). ✅
- Resumed connection maps under source `streamId` (matches SDK `replayEvents`). ✅

## Testing

- 462/462 MCP tests pass. Added coverage for: mid-flight POST disconnect + GET resume, standalone single-send + no-live store-only, supersede-on-resume (POST + standalone), 128-key chunked delete, 1000-event replay cap, global event-id uniqueness, streamId `:` rejection. Two behaviours mutation-checked (supersede close, id uniqueness).

## Compatibility

Patch-level, additive:

- No public API changes on existing types.
- New: `DurableObjectEventStore` + `ClearableEventStore` exported.
- `McpAgent.getStreamForRequestId()` is `@internal`.
- Behaviour changes only widen the set of events that survive disconnects and remove idle DO billing.


threepointone:
clanker review, some of these look legit: 

## High Severity

### Stale POST WebSockets can win routing after GET resume

`StreamableHTTPServerTransport.handleGetRequest()` restores `requestIds` onto a
new GET connection when resuming an active POST stream, but it does not clear or
close the original POST connection. If the HTTP SSE client disconnected but the
Durable Object WebSocket has not been torn down, both connections can carry the
same `requestIds`.

`send()` picks the first live connection whose state contains the request id:

```ts
const liveConnection = Array.from(
  agent.getConnections<TransportConnState>()
).find((conn) => conn.state?.requestIds?.includes(requestId as RequestId));
```

If the stale POST bridge is iterated first, subsequent tool progress or final
results are written to the stale connection instead of the resumed GET stream.
Events are still stored for replay, but live delivery after resume becomes
nondeterministic.

Suggested fix: when a GET resumes a stream, supersede prior connections for that
`streamId`/`requestIds` by clearing their `requestIds` state or closing them.
Also consider propagating HTTP response abort/cancel back to the DO WebSocket in
the streamable HTTP handler.

### Bulk deletes can exceed Durable Object storage limits

`DurableObjectEventStore.clearStream()` and `sweep()` gather all event keys and
call `storage.delete(keys)` once. Durable Object storage bulk delete is limited
to 128 keys per call.

A stream with more than 128 events, or a cleanup sweep covering many streams,
can fail in production even though the mock storage accepts any array size.

Suggested fix: delete in chunks of at most 128 keys, and add tests with storage
mocks that enforce this limit.

### Sweep can delete a stream that receives a concurrent write

`sweep()` scans stream metadata, marks expired streams, then awaits further
storage operations before deleting those streams. A `storeEvent()` can occur
during those awaits, update `lastWriteAt`, and write a new event. The sweep still
deletes the stream because it was expired in the earlier scan.

Suggested fix: before deleting a stream, re-read its metadata or use another
guard to confirm `lastWriteAt` is still older than the cutoff.

## Medium Severity

### Replay and cleanup are not paginated

`replayEventsAfter()`, `clearStream()`, and `sweep()` each perform single
`storage.list()` calls. This creates two risks:

- Long missed histories may be truncated if list results are capped, or may load
  too much into memory if unbounded.
- Cleanup may miss metadata or event rows beyond the first page, and deleting
  metadata after only partially deleting events can leave orphaned event keys
  that future sweeps will never see.

Suggested fix: use explicit limits and pagination loops for replay, per-stream
deletion, and metadata sweeps.

### Prefix scanning can collide if stream ids contain colons

Event keys are stored as:

```text
__mcp_event__:<streamId>:<seqHex>
```

Operations build prefixes with `__mcp_event__:${streamId}:`. If stream ids such
as `a` and `a:b` both exist, scanning or clearing stream `a` also matches keys
for `a:b`.

Default stream ids appear safe, but exported `DurableObjectEventStore` can be
used by embedders with custom stream ids.

Suggested fix: encode `streamId` in storage keys, or document/enforce that
stream ids must not contain `:`.

### Completed-POST resume test is mislabeled as standalone resume

`resumability.test.ts` has a test named "resumed GET stream is registered as the
standalone SSE stream", but it resumes from a POST tool-call event id. After the
POST completes, `deleteStreamRequestIds()` has already run, so this is a
completed-POST replay channel, not a standalone stream. The test only asserts
HTTP 200 and does not verify standalone registration or future live delivery.

Suggested fix: add a true standalone test: open standalone GET, receive a
standalone event id, disconnect, reconnect with `Last-Event-ID`, trigger a
server-initiated notification, and assert it arrives live.

### No test covers true mid-flight POST disconnect and live continuation

The POST replay test reads the final result, cancels the response, then resumes
from the prior id. This validates replay of a completed stream, not the main
failure mode where the POST SSE stream is interrupted while the tool call is
still running and the resumed GET should receive later messages live.

Suggested fix: add an end-to-end test with a deferred tool: start POST, read an
initial event id, cancel/abort the POST response before releasing the tool,
resume with GET + `Last-Event-ID`, release the tool, and assert the final result
arrives on the resumed GET stream.

## Lower Severity

### `getStreamIdForEventId()` accepts synthetic event ids

`getStreamIdForEventId()` derives the stream id by splitting on the last colon
and does not verify that the event exists. A client can send a fake
`Last-Event-ID`, causing the transport to register a ghost replay connection and
replay nothing.

Suggested fix: verify the event key exists before treating the id as resumable,
or fall through to a fresh standalone stream when no matching event is found.

### `_requestResponseMap` is in-memory for batched POSTs

For batched requests, `shouldClose` depends on `_requestResponseMap`. If the DO
hibernates between responses, that in-memory map can be lost, leaving the stream
mapping uncleared and the close frame unsent. The MCP spec discourages batched
requests, so this is lower risk, but the transport still accepts them.

### Multiple standalone GET streams have last-writer-wins behavior

The standalone send path iterates all `_standaloneSse` connections and keeps the
last one. Earlier standalone streams silently stop receiving live
server-initiated messages, even though the spec allows multiple concurrent SSE
streams. Stored events still make replay possible.

### Comments are stale or misleading

- `transport.ts` says cleanup is "scheduled hourly from `McpAgent.onStart`",
  but the new cleanup is armed by `storeEvent()` and scheduled around
  `maxAgeMs`.
- `index.ts` mentions `maxEventsPerStream`, but the PR removed that cap.
- `replayEventsAfter()` comments about list defaults should be verified against
  current Durable Object storage semantics.

## Coverage Gaps To Add

- Bulk delete over 128 keys.
- Replay with more than one list page of events.
- Sweep with more than one list page of metadata.
- Sweep of a stream with more than one list page of events.
- Sweep racing with a fresh `storeEvent()`.
- Prefix-collision stream ids such as `a` and `a:b`.
- Malformed or synthetic `Last-Event-ID`.
- Active POST disconnect, GET resume, and live final-result delivery.
- Standalone GET disconnect, GET resume, and live notification delivery.

mattzcarey:
actually gonna drop the store sweep and just clean up on successful stream. only DOs with un finished streams will have events left and I think thats fine. this is way too much complexity since I am gonna rip it all out in a month anyway.

mattzcarey:
/bonk do a last review. does it all look good?

mattzcarey:
/bonk go fo it. rip me to shreds

threepointone (approved):

PR #1608, opened 2026-05-28
Fix auto-continuation stream resume race

### LINKED ISSUE #1586: `addToolOutput` / client-tool auto-continuations don't attach to the response stream — delivered only as an end-of-turn replay
## Affected versions

- `@cloudflare/think` **0.7.2** (latest)
- `@cloudflare/ai-chat` **0.7.1** (latest, client `useAgentChat`)
- `agents` **0.13.2** (`agents/chat` `ContinuationState`)
- `ai` 6.0.191, `@ai-sdk/react` 3.0.193

## TL;DR

When a client-side tool result is submitted with `addToolOutput(...)` (autoContinue enabled), the server runs the continuation turn but the client **never observes its stream**. The assistant message (text + tool-input/-output chunks) is withheld and only appears once the turn finishes, via a single `cf_agent_chat_messages` replay — i.e. "nothing, then everything at once." Live `agent-tool-event` frames (sub-agent runs) DO stream throughout; only the assistant message stream is missing.

A normal `sendMessage` turn streams incrementally and correctly. The regression is specific to the **server-pushed auto-continuation** path after a client tool result.

## Repro

1. Server agent extends Think (`AIChatAgent`), with a client-side (no-execute) tool, e.g. a "choose/picker" tool the user resolves in the UI.
2. Client uses `useAgentChat` (default `autoContinueAfterToolResult: true`, `resume: true`).
3. User resolves the client tool → client calls `addToolOutput({...})`.
4. The model continues (multi-step: more tool calls + text).

**Expected:** the continuation streams to the client incrementally (as `sendMessage` turns do).
**Actual:** the client sees nothing until the turn completes, then the whole continuation appears at once.

## Evidence (live WebSocket capture, client side)

Instrumented `WebSocket.send`/`onmessage` + a DOM MutationObserver, drove a real picker→submit→continuation (a ~43s turn):

- `send cf_agent_tool_result` @ dt 6913ms
- `send cf_agent_stream_resume_request` @ dt 6913ms (immediately after — the client DOES ask to observe)
- **No** `cf_agent_stream_resuming` ever received.
- **No** `cf_agent_use_chat_response` body chunks for the whole turn — only the terminal `{done:true}` @ dt ~43.5s, immediately followed by `cf_agent_chat_messages` (full replay).
- DOM text jumps from N→N+large in one shot at turn end.
- `agent-tool-event` frames (sub-agent/worker runs) stream live throughout (dozens), proving the connection is alive and that channel works.

(Before the partial fix below, the same probe showed `cf_agent_stream_resume_none` arriving early instead — see "Gap 1".)

## Root cause — one confirmed code bug + a remaining notify-on-start failure

Gap 1 is confirmed at the code level (and its fix verified to change behavior). Gap 2 is confirmed as an observed symptom (no `stream_resuming` after the Gap-1 fix) but **not yet localized to a specific line** — the candidates are listed below.

### Gap 1 — `@cloudflare/think` `_scheduleAutoContinuation` doesn't register `pending` synchronously

`think.js` `_scheduleAutoContinuation(connection)`: in the no-pending case it only sets a 50ms coalesce timer and defers creating `this._continuation.pending` / `awaitingConnections` to `_fireAutoContinuation` (which runs after the timer). Meanwhile the client sends `cf_agent_tool_result` and *immediately* `resumeStream()` → `cf_agent_stream_resume_request`. That request hits `_handleStreamResumeRequest` during the window where there's no active stream **and** no `pending` → it replies `cf_agent_stream_resume_none`, and the client stops trying to observe.

**Suggested fix** (establish `pending` + `awaitingConnections` synchronously; keep the 50ms timer only to defer `_fireAutoContinuation`, which already no-ops pending creation when it already exists):

```diff
@@ _scheduleAutoContinuation(connection) — no-pending branch @@
 			this._continuation.awaitingConnections.set(connection.id, connection);
 			return;
 		}
+		// Establish pending + awaiting synchronously so an immediate
+		// cf_agent_stream_resume_request registers as "awaiting" instead of
+		// getting cf_agent_stream_resume_none during the 50ms coalesce window.
+		this._continuation.pending = {
+			connection,
+			connectionId: connection.id,
+			requestId: crypto.randomUUID(),
+			clientTools: this._lastClientTools,
+			body: void 0,
+			errorPrefix: "[Think] Auto-continuation failed:",
+			prerequisite: null,
+			pastCoalesce: false
+		};
+		this._continuation.awaitingConnections.set(connection.id, connection);
 		if (this._continuationTimer) clearTimeout(this._continuationTimer);
 		this._continuationTimer = setTimeout(() => {
 			this._continuationTimer = null;
 			this._fireAutoContinuation(connection);
 		}, 50);
```

We verified this **eliminates `cf_agent_stream_resume_none`** (the request now registers as awaiting). But it is **not sufficient** on its own — see Gap 2. (Worse, alone it converts the fast `resume_none` into a silent ~5s tool-continuation timeout, so it should land together with Gap 2.)

### Gap 2 — the awaiting connection is never notified when the stream starts

With Gap 1 fixed, `_streamResult` is supposed to notify awaiting connections when the continuation stream begins:

```js
if (this._continuation.pending?.requestId === requestId) {
  this._continuation.activatePending();
  this._continuation.flushAwaitingConnections((c) => this._notifyStreamResuming(c));
}
```

But in the repro, **no `cf_agent_stream_resuming` is ever sent**, so the registered-awaiting client is never told the stream started. `ContinuationState.activatePending()` clears only `pending` (not `awaitingConnections`), and `flushAwaitingConnections()` iterates+clears that map — so the agents-side data structure looks compatible. The remaining candidates (need maintainer insight / server-side tracing):

1. `this._continuation.pending?.requestId === requestId` is **false** at `_streamResult` time (requestId not preserved from `_fireAutoContinuation` through the turn queue / inference loop into `_streamResult`).
2. `awaitingConnections` is **empty** by then — e.g. the WebSocket connection churns during the (multi-second) turn, `onClose` deletes the awaiting entry, and the new connection isn't re-registered.
3. `_notifyStreamResuming` → `sendIfOpen` silently no-ops because the stored connection object is stale/closed (even though a *new* connection is alive and receiving `agent-tool-event`s).

A cheap definitive confirm is temporary logging at four points: `_handleStreamResumeRequest` (pending id/connection id, awaiting size, branch taken), `_streamResult` just before the pending check (requestId, pending requestId, awaiting size), `_notifyStreamResuming` (connection id, send result), and `onClose` (whether it cleared pending/awaiting).

## Workaround we shipped

We surface live `agent-tool-event` (sub-agent run) progress in the UI during the continuation so the long server tool isn't a silent gap, and we keep live state authoritative over the (stale, status-keyed) server-trace snapshot while a turn is in flight. This doesn't recover the assistant text/tool chunks — those still appear at turn end — but it covers the dominant wait. The real fix is Gap 1 (confirmed) plus resolving the remaining notify-on-start failure (Gap 2).

## Why it matters

Any agent using client-side tools (HITL pickers, approvals, browser-executed tools) with `autoContinueAfterToolResult` loses incremental streaming for the entire post-tool turn — which is often the longest turn (it does the real work). `sendMessage` turns are unaffected, which is why this is easy to miss.

cjol:
This PR fixes a race in Think auto-continuation streaming after client-side tool output.

When a client calls `addToolOutput(...)` with `autoContinue: true`, the client immediately asks to resume/attach to the continuation stream. Think was only creating `ContinuationState.pending` after the 50ms coalescing timer fired, so the immediate `cf_agent_stream_resume_request` could arrive while there was no active stream and no pending continuation. The server would reply `cf_agent_stream_resume_none`, causing the client’s deferred stream to close and leaving the user with no live assistant chunks until end-of-turn replay.

The fix keeps the existing 50ms coalescing delay for starting the continuation turn, but registers the continuation as pending synchronously:

- `pending` now exists before the coalescing timer fires, so an immediate resume request is held in `awaitingConnections` instead of getting `cf_agent_stream_resume_none`.
- The timer fires the continuation from the latest pending connection, avoiding a stale captured connection if the pending continuation is updated during the coalescing window.
- `_fireAutoContinuation` now consumes the already-registered pending state instead of creating it late.
- `ContinuationState` is now generic over the concrete connection type, so Think and AIChatAgent can store full `Connection` objects without casts while keeping the shared chat primitive decoupled from the Agents connection type.
- 
The regression test sends `cf_agent_tool_result` with `autoContinue: true` and immediately follows it with `cf_agent_stream_resume_request`. It failed before this patch, and passes now. It asserts that the server responds with `cf_agent_stream_resuming`, not `cf_agent_stream_resume_none`, which captures the broken timing from #1586.

Fixes #1586 

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1608" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/early-ravens-continue.md
+++ b/.changeset/early-ravens-continue.md
@@ -1,0 +1,6 @@
+---
+"@cloudflare/think": patch
+"agents": patch
+---
+
+Fix Think auto-continuation stream resumes so immediate client-tool resume requests attach to the pending continuation instead of receiving `cf_agent_stream_resume_none`.
--- a/.changeset/early-ravens-continue.md
+++ b/.changeset/early-ravens-continue.md
@@ -1,6 +1,7 @@
 ---
 "@cloudflare/think": patch
+"@cloudflare/ai-chat": patch
 "agents": patch
 ---
 
-Fix Think auto-continuation stream resumes so immediate client-tool resume requests attach to the pending continuation instead of receiving `cf_agent_stream_resume_none`.
+Fix auto-continuation stream resumes so immediate client-tool resume requests attach to the pending continuation instead of receiving `cf_agent_stream_resume_none`.
--- a/packages/agents/src/chat/__tests__/continuation-state.test.ts
+++ b/packages/agents/src/chat/__tests__/continuation-state.test.ts
@@ -113,6 +113,49 @@ describe("ContinuationState", () => {
     expect(state.awaitingConnections.size).toBe(0);
   });
 
+  // ── releaseConnection ─────────────────────────────────────────────
+
+  it("releaseConnection clears ownership without removing pending", () => {
+    const state = new ContinuationState();
+    const conn = makeConnection("c1");
+    state.pending = {
+      connection: conn,
+      connectionId: "c1",
+      requestId: "r1",
+      errorPrefix: null,
+      prerequisite: null,
+      pastCoalesce: false
+    };
+    state.awaitingConnections.set("c1", conn);
+
+    state.releaseConnection("c1");
+
+    expect(state.pending).not.toBeNull();
+    expect(state.pending!.connection).toBe(conn);
+    expect(state.pending!.connectionId).toBeNull();
+    expect(state.awaitingConnections.size).toBe(0);
+  });
+
+  it("releaseConnection clears deferred and active ownership", () => {
+    const state = new ContinuationState();
+    const conn = makeConnection("c1");
+    state.deferred = {
+      connection: conn,
+      connectionId: "c1",
+      errorPrefix: "[test]",
+      prerequisite: null
+    };
+    state.activeRequestId = "r1";
+    state.activeConnectionId = "c1";
+
+    state.releaseConnection("c1");
+
+    expect(state.deferred).not.toBeNull();
+    expect(state.deferred!.connectionId).toBeNull();
+    expect(state.activeRequestId).toBe("r1");
+    expect(state.activeConnectionId).toBeNull();
+  });
+
   // ── sendResumeNone ────────────────────────────────────────────────
 
   it("sendResumeNone sends to all awaiting connections then clears", () => {
--- a/packages/agents/src/chat/continuation-state.ts
+++ b/packages/agents/src/chat/continuation-state.ts
@@ -46,8 +46,10 @@ export interface ContinuationConnection {
   send(message: string): void;
 }
 
-export interface ContinuationPending {
-  connection: ContinuationConnection;
+export interface ContinuationPending<
+  TConnection extends ContinuationConnection = ContinuationConnection
+> {
+  connection: TConnection;
   connectionId: string;
   requestId: string;
   clientTools?: ClientToolSchema[];
@@ -57,8 +59,10 @@ export interface ContinuationPending {
   pastCoalesce: boolean;
 }
 
-export interface ContinuationDeferred {
-  connection: ContinuationConnection;
+export interface ContinuationDeferred<
+  TConnection extends ContinuationConnection = ContinuationConnection
+> {
+  connection: TConnection;
   connectionId: string;
   clientTools?: ClientToolSchema[];
   body?: Record<string, unknown>;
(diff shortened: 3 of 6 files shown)


threepointone (approved):
lgtm, added a small edge case fix

PR #1618, opened 2026-05-29
Fix chat continuation failing on models that reject assistant-prefill

threepointone:
## Summary

When a chat turn is interrupted mid-stream (e.g. by a deploy) and recovery
continues it, `continueLastTurn` replays a transcript whose **final message is
the partial assistant message** — an "assistant prefill". Modern chat models
reject a request that ends in an assistant message, so the continuation throws
and the turn is left interrupted.

- **Anthropic Claude 4.6+** (Opus 4.6/4.7, **Sonnet 4.6**) returns a **400**:
  `This model does not support assistant message prefill. The conversation must
  end with a user message.` (Anthropic's own migration guidance: *"move the
  continuation text into a user message"*.)
- **Kimi 2.6 (Workers AI)** tolerates it (responds normally) — so this is
  model-specific, and bites the modern Anthropic models the customer runs.

## Fix

`ensureValidContinueCheckpoint(messages)` appends an ephemeral user "continue"
checkpoint whenever a model request would otherwise end in an assistant message.
It is applied to `finalMessages` in `_runInferenceLoop` (after `beforeTurn`, so
subclass message overrides are protected too), shapes **only the model request**,
and is never persisted to the transcript.

Auto-continuation after a tool call is unaffected: that transcript ends in a
tool-role (user-side) message, not an assistant message.

`@cloudflare/ai-chat` is not affected — there the user's `onChatMessage` builds
the model request; Think is the one that assembles and calls the model in
`_runInferenceLoop`.

## Reproduction

The `examples/deploy-churn` harness gains `/probe/trailing-{user,assistant}`
routes (Workers AI default; `?provider=anthropic&model=claude-sonnet-4-6`):
- Kimi 2.6: trailing-assistant → `ok` (no repro).
- Anthropic Sonnet 4.6: trailing-assistant → `AI_APICallError` (the 400 above).

Findings are logged in `examples/deploy-churn/INVESTIGATION.md`.

## Test

New deterministic regression test in `think-session.test.ts`
("continuation does not replay a trailing assistant message (assistant
prefill)") uses a mock model that rejects a trailing-assistant prompt. It
**fails without the fix** (`expected 'assistant' not to be 'assistant'`) and
passes with it.

## Test plan

- [x] `npm run test -w @cloudflare/think` — 432 pass (incl. new test)
- [x] Confirmed the new test fails without the fix and passes with it
- [x] lint + typecheck
- [x] Live repro via `examples/deploy-churn` against Kimi 2.6 and Anthropic Sonnet 4.6
- [ ] CI green

## Follow-up (not in this PR)

Interrupted turns can still look "frozen" on the client after a WS reconnect
storm — but that is a separate **hydration** gap (terminal/recovery status is
broadcast transiently and not replayed on reconnect), tracked separately in
`INVESTIGATION.md`. Companion to the recovery fixes in #1615 and #1617.


threepointone:
/bonk can you run the formatter on INVESTIGATION.md and commit it and push it?

threepointone:
/bonk just run `npm run format` in the root, you'll see it'll format that file. thn commit and push it.

PR #1621, opened 2026-05-30
Persist settled tool results durably so recovery never re-runs completed tool calls

threepointone:
## Summary

Under deploy churn, recovery could re-run **already-completed tool calls** (the customer observed byte-identical duplicate, non-idempotent INSERTs — a whole message's tool calls rolling back, not just the interrupted step).

Root cause: resumable-stream chunks are buffered in memory and batch-flushed to SQLite every ~10 chunks (`CHUNK_BUFFER_SIZE`). The main **WebSocket** `_streamResult` (used by chat turns and `continueLastTurn` recoveries) did **not** force a flush on settled tool results, so an isolate eviction before the next batch flush lost them. On recovery, `_persistOrphanedStream` rebuilds the partial assistant message from SQLite — without those settled tool calls — so the continuation re-anchors without them and the model re-runs the completed tool call.

The **sub-agent RPC** streaming path (`_streamResultToRpcCallback`) already flushed recoverable content on settle (the tool-call-level durability referenced by #1615/#1617). This PR brings the WebSocket path to parity.

## Fix

A shared `_storeChunkDurably` helper (used by both the WebSocket and RPC stream loops) stores each chunk and flushes the buffer **immediately** on a settled tool result (`tool-output-available` / `tool-output-error`). Frequent recoverable content (text / reasoning / tool-input streaming) stays throttled (first chunk + every ~10) to avoid write amplification. `_shouldFlushRpcStreamChunk` is generalized to `_shouldFlushRecoverableChunk`.

Net effect: a settled tool result is durable the moment it settles, so `_persistOrphanedStream` always reconstructs every completed tool call, and **recovery loses at most the single in-flight step** — even when multiple evictions hit one turn.

## Also: two terminal-status hydration gaps (#1619 follow-up)

Review of #1619 surfaced two remaining "frozen turn" cases where a terminal outcome was broadcast (or silent) but never **persisted**, so a client disconnected at that moment stayed frozen on reconnect. Both are folded in here:

- **Pre-stream errors in `_handleChatRequest`.** The outer catch broadcasts `MSG_CHAT_RESPONSE { error, done }` for failures before the stream starts (e.g. message reconciliation/persist), but never called `_recordTerminalChatStatus`. #1619 only hooked the stream-level `_fireResponseHook` and recovery exhaustion. It now records terminal `"error"` before broadcasting, so `_buildIdleConnectMessages` replays it on connect.
- **`continue_disabled` recovery skip.** When `onChatRecovery` returns `{ continue: false }`, the turn was only marked interrupted via `_markRecoveredSubmissionInterrupted` (which updates the programmatic submission row — it does **not** touch the WS transcript), so a WS chat client got no durable terminal signal at all. Unlike a benign `conversation_changed` skip (a newer turn already owns the UI), disabling recovery abandons the turn with no superseding turn, so it now records terminal status and broadcasts an error like exhaustion does. `conversation_changed` / `not_recoverable` skips remain silent.

## Test

`think-session.test.ts`:
- "flushes a settled tool result to durable storage immediately": stream two text deltas (throttled → one stays buffered), then a settled tool result; assert the tool result forces a flush so it is durable in SQLite **before** the stream completes. Verified to fail without the tool-output special-case (`expected 1 to be greater than 1`).
- "replays a pre-stream error to a reconnecting client (hydration)": a `beforeTurn` throw (standing in for a reconciliation failure) drives `_handleChatRequest`'s outer catch; assert the idle-connect replay surfaces the terminal error.
- "{ continue: false } surfaces a terminal error on reconnect": an interrupted turn whose `onChatRecovery` returns `{ continue: false }`; assert the idle-connect replay surfaces the terminal error.

## Test plan

- [x] `npm run test -w @cloudflare/think` — `think-session.test.ts` 130 pass (incl. 3 new regression tests)
- [x] Confirmed the tool-result test fails without the fix and passes with it
- [x] lint + typecheck (`tsc --noEmit -p packages/think/tsconfig.json`)
- [ ] `npm run test:e2e -w @cloudflare/think` — confirmed green by the author locally
- [ ] CI green

## Related

Completes the deploy-churn recovery hardening series: #1615, #1617, #1618, #1619.

### DIFF
--- a/.changeset/persist-settled-tool-results.md
+++ b/.changeset/persist-settled-tool-results.md
@@ -1,0 +1,15 @@
+---
+"@cloudflare/think": patch
+---
+
+Settled tool results are now flushed to durable storage immediately during a
+chat turn, so recovery never re-runs an already-completed (often non-idempotent)
+tool call. Stream chunks are batched in memory and flushed to SQLite every ~10
+chunks; the WebSocket chat path did not force a flush on settled tool results,
+so an isolate eviction (deploy) before the next batch flush lost them. Recovery
+then rebuilt the partial assistant message without those tool calls and the
+model re-ran them (e.g. duplicate INSERTs). The sub-agent RPC streaming path
+already flushed recoverable content; this brings the WebSocket path to parity
+via a shared `_storeChunkDurably` helper that flushes immediately on
+`tool-output-available` / `tool-output-error`. Net effect: recovery loses at
+most the single in-flight step, even when multiple evictions hit one turn.
--- a/.changeset/persist-settled-tool-results.md
+++ b/.changeset/persist-settled-tool-results.md
@@ -13,3 +13,11 @@ already flushed recoverable content; this brings the WebSocket path to parity
 via a shared `_storeChunkDurably` helper that flushes immediately on
 `tool-output-available` / `tool-output-error`. Net effect: recovery loses at
 most the single in-flight step, even when multiple evictions hit one turn.
+
+Also closes two remaining "frozen turn" hydration gaps from the terminal-status
+work: a turn that fails before the stream starts (e.g. a message reconciliation
+error in `_handleChatRequest`) now records its terminal status, and a recovery
+skip caused by `onChatRecovery` returning `{ continue: false }` now surfaces a
+terminal error too. Both were previously broadcast (or silent) but not persisted,
+so a client disconnected at that moment stayed frozen on reconnect. Benign skips
+such as `conversation_changed` (a newer turn already owns the UI) remain silent.
--- a/packages/agents/src/e2e-tests/fiber-eviction.test.ts
+++ b/packages/agents/src/e2e-tests/fiber-eviction.test.ts
@@ -14,9 +14,16 @@
  */
 import { describe, it, expect, afterEach, beforeEach } from "vitest";
 import { spawn, execSync, type ChildProcess } from "node:child_process";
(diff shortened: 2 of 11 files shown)


threepointone:
Folded in an e2e harness stability fix (`881f11f4`).

**Symptom:** `nx run @cloudflare/think:test:e2e` intermittently exited non-zero with `Timeout terminating forks worker for test files .../assistant-e2e.test.ts` / `Worker exited unexpectedly`, even though all 9 e2e tests passed.

**Diagnosis:** not a logic failure and not introduced by this PR's code — the e2e harness (untouched since #1611/#1601) leaves open handles that keep the vitest forks-pool worker alive past the teardown window under load. Reproduced ~1/4 full-suite runs on this branch; passed in isolation and on a pre-#1619 commit. The error always named `assistant-e2e` because its `killProcess` left an uncleared `setTimeout(resolve, 3000)` (the sibling `chat-recovery` file already cleared it), and it keeps one `wrangler dev` alive across all tests so undici keep-alive sockets linger until the final `afterAll`.

**Fix (harness only, no published-package change):**
- Clear the fallback timer in `assistant-e2e`'s `killProcess` once the child exits.
- Raise the e2e `teardownTimeout` to 60s.
- Raise `waitForReady` boot budget to 60s in both files (absorbs slow boots → `Wrangler did not start in time`).

Verified green on repeated full-suite runs after the change.

Issue #1593, opened 2026-05-26
Auto-compaction trigger fires, but compaction doesn't run due to internal token counting


rwdaigle:
**Describe the bug**

Follow-up to https://github.com/cloudflare/agents/issues/1538

`Session.compactAfter(threshold, { tokenCounter })` correctly fires the registered `onCompaction` callback when our custom counter reports a value above the threshold, but the bundled `createCompactFunction` helper returns `null` *without ever invoking the `summarize` callback* on conversations whose bulk is tool-call inputs and tool-result outputs (not text). The result is that the auto-compaction trigger fires every turn, no summarization runs, no compaction overlay is stored, and the history keeps growing.

The root cause appears to be that `createCompactFunction`'s boundary logic (`findTailCutByTokens` → `estimateMessageTokens`) uses a chars/4 heuristic that effectively only counts `text` + `tool-*` parts' string content and severely under-counts agent-style histories. With our histories the estimator sees ~30–40K "tokens" total while the model reports 140K+, so the configured `tailTokenBudget` (default 20K, we tried 80K) already covers the entire history and `compressEnd <= compressStart`, returning `null`.

The new `tokenCounter` option fixed the trigger half of the problem — it gives the consumer a way to feed the SDK a real, model-reported token count for the *fire/no-fire* decision. But it does not extend into the boundary-cut logic inside `createCompactFunction`, so the *what to compress* decision is still bound to the legacy heuristic.

**To Reproduce**

1. Create a `Session` with:
   - `compactAfter(0, { tokenCounter })` where `tokenCounter` returns a large value (simulating model-reported `usage.inputTokens`) once you want compaction to fire.
   - `onCompaction(createCompactFunction({ summarize, tailTokenBudget }))`.
2. Append a tool-heavy conversation history (~20 messages where each assistant message is a multi-step turn with several `tool-*` parts whose `input` / `output` together total a few KB each — typical of agent loops doing file reads, DB queries, etc.). Keep `text` and `reasoning` parts small.
3. Append a new user message so the SDK's `_appendMessage` threshold check runs and `tokenCounter` returns over budget.
4. Observe:
   - `onCompaction` is invoked.
   - The function returned by `createCompactFunction` returns `null`.
   - `summarize` is never called.
   - No `cf_agent_session` status broadcast carries `compacted: { tokensBefore }`.
   - History on disk is unchanged; the next turn replays the full transcript.

**Expected behavior**

[...]

Issue #10, opened 2025-02-22
better error handling everywhere


threepointone:
lots of errors get swallowed atm

PR #1623, opened 2026-05-30
fix(think,ai-chat,agents): harden recovery, transcript integrity & compaction under deploy churn

threepointone:
## Summary

This PR hardens **Think** (and, where the machinery is shared, **AIChatAgent** and the **agents** Session memory / sub-agent layer) against the failure modes a customer hit while running long, tool-heavy agent turns through **repeated `wrangler deploy` churn**. A mid-turn deploy resets the Durable Object (`"reset because its code was updated"`), and under rapid redeploys a single turn can be interrupted several times in a row. The bugs below combined to make turns that *actually completed* look failed, make completed tool calls re-run, make interrupted tool calls vanish, make malformed tool inputs 400 forever, make auto-compaction silently no-op on tool-heavy histories, and make sub-agent interruptions look like terminal failures.

All fixes are covered by unit tests plus new e2e/real-deploy churn harnesses (`examples/deploy-churn`, Think e2e `tool-rollback` + `task-amplification`).

Changesets included:
- `recovery-submission-status-under-churn.md` — `@cloudflare/think` + `@cloudflare/ai-chat` (patch)
- `preserve-interrupted-tool-calls.md` — `@cloudflare/think` (patch)
- `compaction-tokencounter-drives-boundary.md` — `agents` (patch, #1593)
- `agent-tool-structured-failure.md` — `agents` (minor)

---

## 1. Recovery falsely erroring a completed turn under repeated mid-turn deploys

When several deploys interrupt one turn, recovery runs a **chain** of continuations. Three bugs combined to leave the durable submission in `error` even though every step completed:

- **Lost ownership** — the submission link (`recoveredRequestId`) was derived from each continuation's own *fresh* requestId, so chained continuations dropped it; the continuation that finally finished the turn could no longer mark the submission `completed`.
- **Stale-continuation clobber** — a superseded continuation tripped the `conversation_changed` guard because the leaf had advanced via recovery's *own* forward progress (a new assistant message), not a new user turn, and overwrote the still-running submission to `error`.
- **Premature `stable_timeout`** — a timeout while waiting for the isolate to settle (common with a deploy in flight) failed the turn terminally on the first attempt.

**Fix:** submission ownership is keyed off the stable **recovery root** and threaded through the entire continuation chain; a superseded continuation **skips benignly** (only a genuinely newer *user* turn marks the submission `skipped`, never `error`); and a stable-state timeout **reschedules within the `maxAttempts` budget** instead of abandoning. `@cloudflare/ai-chat` shares the recovery machinery but has no durable-submission layer, so it gets the `stable_timeout` reschedule fix only.

Commits: `ff8a4340`, `d59a7287`.

## 2. Interrupted tool calls disappeared and were silently re-run

Transcript repair previously **removed** a tool call with no recorded output (interrupted mid-execution by a deploy, or an `ask_user`-style tool answered by the user's next message) before the next turn. That made the call vanish from the broadcast transcript and let the model **re-run** it — duplicating non-idempotent side effects.

**Fix:** the orphan is now flipped to `state: "output-error"` with an explanatory message, so the user-visible record survives, the model sees the tool errored instead of re-running it blind, and the provider still receives a valid tool-result (no `AI_MissingToolResultsError`). As a **last-line backstop**, `convertToModelMessages` is now called with `ignoreIncompleteToolCalls: true`.

Commits: `aad63224`, `bfec18c2`, `5ca0e378`.

## 3. Malformed tool inputs 400 the provider forever

Repair parsed stringified-JSON tool `input`s, but a tool call with a **missing or `null`** `input` was left unrepaired — Anthropic rejects a `tool_use` block whose `input` is absent, so the turn 400s on every retry.

**Fix:** a new `_normalizeToolInput` helper parses stringified JSON **and** defaults a missing/`null` input to `{}`, applied on both the orphan-healing and settled-part paths.

Commit: `5c57ac52`.

## 4. Auto-compaction fired every turn but silently no-op'd on tool-heavy histories (#1593)

`Session.compactAfter(threshold, { tokenCounter })` fired the `onCompaction` callback, but the bundled `createCompactFunction` boundary walk still used the Workers-safe `chars/4` heuristic. On tool-heavy histories that under-counts ~4–5×, so the configured tail budget covered the entire history, `compressEnd <= compressStart`, and the function returned `null` — compaction fired every turn but never shortened anything.

**Fix:** the Session passes its counter to the compaction function via a new `CompactContext` argument; `createCompactFunction` uses it for the tail-budget walk when no explicit `CompactOptions.tokenCounter` was given. A single `tokenCounter` on `compactAfter()` now drives **both** "should we compact?" and "what should we compact?". A trigger that still returns `null` logs a one-time warning (re-armed on the next successful compaction).

Commits: `ce06564e`, `52e69855`.

## 5. Sub-agent interruptions looked like terminal failures

`agentTool()` collapsed every non-completed sub-agent run to an opaque `{ ok: false, error: string }`. A child reset/superseded by a deploy or parent recovery (`interrupted`) was indistinguishable from a genuine failure or an intentional cancellation, so the parent model would often parrot the interruption text back to the user as final.

**Fix:** failures now return a structured `AgentToolFailure { ok: false, status, error, retryable }`. `interrupted` → `retryable: true` (and surfaces the interruption reason); `aborted` and `error` → `retryable: false`. Backward compatible for consumers reading `ok`/`error`; `AgentToolFailure` is exported from `agents` so an orchestration harness (or a parent prompt convention) can re-dispatch an interrupted sub-agent instead of reporting it as final.

Commit: `1f62e9ff`.

## 6. Test harnesses (no shipped behavior)

- `examples/deploy-churn` — a real `wrangler deploy` churn harness (`scripts/deploy-rollback.ts`).
- `packages/think/src/e2e-tests/tool-rollback.test.ts` — settled-tool durability under recovery.
- `packages/think/src/e2e-tests/task-amplification.test.ts` — proves sub-agent (`task`) re-runs stay **bounded** under churn (no parent↔child budget amplification).
- Unit coverage: `think-session.test.ts`, `message-reconciliation.test.ts` (orphan-preserved + stringified **and missing** input), `agent-tools-failure.test.ts` (the new failure envelope), `durable-chat-recovery.test.ts`, `session.test.ts`.

Commits: `1fc27c2e`, `ce07fccc`.

---

## 7. Silent-hang watchdog for the streaming read loop

A model stream that parked without ever throwing (no chunk, no error, no `done`) left the chat read loop waiting forever — an infinite spinner with no terminal state. There was no detection for a silently hung turn (only the recovery-path `stable_timeout`, which guards recovery *scheduling*, not a live stream).

**Fix:** opt-in `chatStreamStallTimeoutMs` (default `0` = off). If no UI-message-stream chunk arrives within the window, the watchdog aborts the turn so the loop exits with a terminal stream error (routed through `onChatError` with `stage: "stream"`) and emits a new `chat:stream:stalled` observability event. Applies to both the WebSocket turn loop and the `chat()` / sub-agent callback loop. It measures *inter-chunk* inactivity (which includes server-side tool execution), so it must be set above the slowest expected model TTFT and tool latency — hence opt-in, not a surprising default.

Commit: `2d5f9f2d`. Changeset: `think-stream-stall-watchdog.md`.

## Test plan

- [x] `npm run check` (sherif + export checks + oxfmt + oxlint + typecheck) — all 91 projects typecheck, lint + format clean
- [x] `think` message-reconciliation (5/5), `think`/`ai-chat` agent-tools reconciliation (14/14, 19/19), `agents` agentTool envelope (4/4), `agents` session (74/74)
- [ ] `npx nx affected -t test` (full affected suite in CI)
- [ ] Think e2e: `tool-rollback`, `task-amplification`, `chat-recovery`
- [ ] Manual: `examples/deploy-churn` real-deploy churn — turn ends `completed`, no duplicate tool execution, no disappearing tool calls

## Out of scope (separate follow-ups)

These remain genuine gaps that no engine owns yet — intentionally not in this PR:

- **Prompt caching** + dynamic per-turn cached prefix (`withCachedPrompt` is opt-in and the cached prefix freezes until `refreshSystemPrompt()`; no Anthropic `cacheControl` breakpoints in core).
- ~~Silent mid-stream hang watchdog~~ — **addressed in this PR** (see section 7).


threepointone:
/bonk run `npm run format` and commit and push the changes.

threepointone:
/bonk do a deep review on this PR, look for edge cases, what do you think

threepointone:
## Update — review-driven fixes + a terminal-state sweep

Since the deep review, I've addressed both review findings and run a focused sweep for the bug class they exposed ("is every terminal tool/stream state enumerated as terminal?"). All of the below are on the branch; `npm run check` is green (91/91 typecheck, lint + format clean) and every fix has a regression test verified to fail before the fix.

### Review findings
- **Submission abandon-paths used the per-continuation requestId** (`1e36ea3a`). `_markRecoveredSubmissionInterrupted` was called with the per-continuation id in both terminal abandon paths — recovery **exhaustion** (`_exhaustChatRecovery`) *and* `{ continue: false }`. Under chained continuations the submission row still carries the recovery-root id, so the lookup missed it and left the submission stuck `running`. Now keyed off the recovery root (stored on the incident record). The reviewer flagged the `continue_disabled` path; the exhaustion path had the identical defect and is fixed too.
- **`output-error` not treated as settled in transcript repair** (`c3b71cdb`). `_repairToolTranscriptParts.hasOutput` omitted `output-error`, so an already-healed (or legitimately errored) tool call was re-flipped every turn — clobbering a real `errorText` with the generic "interrupted" message and emitting spurious `chat:transcript:repaired` events/writes/broadcasts for the life of the conversation. Now settled.

### Sweep fixes (same bug class, additional instances)
- **`output-denied` not recognized as a settled tool state in Think** (`5077de94`). The AI SDK converts `output-denied` (a user-denied approval) into a denial `tool-result`, but Think's repair, backstop detector, and immediate-flush check all omitted it — so repair turned denials into generic errors and the denial wasn't durably flushed. Centralized the terminal-state check into one shared `_toolPartHasSettledResult` predicate so repair and the backstop can't drift again, and added `tool-output-denied` to the immediate-flush set.
- **Stall-watchdog wrapper leaked the source stream on early exit** (`3d77ff4e`). The inactivity-watchdog generator (from the watchdog feature in this PR) didn't forward `.return()` to the source on an early consumer `break` (notably an in-band stream error, where the abort signal isn't set), leaving the underlying `ReadableStream` un-cancelled. Now cancels on early termination, except after a watchdog-triggered abort (which already tore it down).
- **Shared reconciler clobbered server-resolved errors/denials** (`9b522fd7`). `reconcileMessages` (run at persist by **both** Think and AIChatAgent) only protected `output-available` from a stale client part — so a client persisting a stale `input-available` for a tool the server had resolved to `output-error`/`output-denied` overwrote the resolved result. Now protects all three terminal states, overlaying the matching field (`output`/`errorText`/`approval`).

### Reviewed and confirmed complete (no change needed)
`_isTerminalSubmissionStatus`, `streamIsTerminal`, `_messageHasPendingInteraction`, `shouldMarkSkippedAfterGenerationChange`, and AIChatAgent's own tool-state handling (it's the HITL home and already enumerates `output-denied` everywhere it matters).

### Also addressed from the review
- Extracted the stable-timeout reschedule into a shared `_rescheduleRecoveryAfterStableTimeout` helper (mirrors `@cloudflare/ai-chat`) + added the missing `_chatRecoveryRetry` reschedule test (`90655e3e`).
- Docs aligned with the behavior changes, incl. the new `chat:stream:stalled` event and `chatStreamStallTimeoutMs` (`893643c7`, `4b1d7f73`).

### Deferred (tracked, intentionally out of scope)
- #1624 — first-class prompt caching / cacheable-but-dynamic system prefix.
- #1626 — route a stream-stall abort into bounded recovery instead of a terminal error.

threepointone:
/bonk do another deep review of this PR, look at the existing replies and updates, look for edge cases and production worthiness

threepointone:
## Deep review #2 — disposition

Triaged all 10 items against HEAD. Landed the clear wins in `be29d607`; confirmed the rest with reasons; filed one genuine follow-up. Details below so anyone can pick this up later.

### Landed (`be29d607`)
- **#1 Dead code** — removed the unreachable `removedIds`/`deleteMessages` branch in `_repairTranscriptForProvider` (repair preserves every message now, never deletes). Comment added.
- **#3 Stringified-array inputs** — `_normalizeToolInput` now parses a stringified `[...]` input, not just `{...}`. + test (`message-reconciliation.test.ts`).
- **#10 Reconciler overlay** — `mergeServerToolOutputs` overlay is now state-driven: only the field matching the server's terminal state is carried (`output` for output-available, `errorText` for output-error, `approval` for output-denied), so a stray `output` on an output-error part can't ride along. + test asserting `output` is dropped.
- **#6 ai-chat divergence** — added a comment in `_chatRecoveryContinue` explaining AIChatAgent intentionally does NOT split assistant-leaf vs user-leaf on `conversation_changed` (no durable-submission layer to protect), and to mirror Think if submissions are ever added.

### Confirmed — not changing (with reasons)
- **#2 Two-query SELECT race** — negligible: it's a `crypto.randomUUID()` collision risk. In the realistic path the SELECT re-finds the *same* submission (whose `request_id` the UPDATE set), and `_isTerminalSubmissionStatus` gates the emit. `RETURNING` is used in exactly one other place in the codebase; not worth diverging the submission code's style for a UUID-collision-only risk.
- **#5 `AgentToolRunInspection` excludes `"interrupted"`** — valid but **pre-existing** and unrelated to this PR's diffs. Should be a separate type-consistency issue, not folded into this PR. (Flagging here for tracking; happy to file if wanted.)
- **#9 `removedToolCalls` name** — misleading, but it's a field on the **already-released** `chat:transcript:repaired` event, so renaming is a breaking change, not free cleanup. Docs already clarify the semantics ("counts orphaned tool calls healed — preserved as errored, not deleted"). Left as-is deliberately.
- **#4 Watchdog per-chunk timer** — acknowledged micro-opt. `.refresh()` doesn't compose cleanly with the race-and-abandon pattern (loses the per-iteration reject handle); it's an opt-in watchdog and the runtime handles the allocation fine. Not worth the added complexity/risk.
- **#8 addToolOutput-vs-repair race** — acknowledged. The turn queue serializes turns; the residual is a direct `addToolOutput` WS mutation during one of repair's `await`s. Real but narrow, and a deterministic test would be flaky. Tracked as a known invariant.

### Deferred — filed as a follow-up
- **#7 Progress marker is count-based** — **confirmed real.** `_chatRecoveryProgressMarker()` counts current assistant messages and `progress` is stored as a `Math.max` high-water mark, so auto-compaction (which collapses messages into a summary overlay) drops the count → `madeProgress` reads false → the attempt budget isn't reset → **premature exhaustion** under simultaneous compaction + deploy churn. The wall-clock ceiling still bounds it (no infinite loop), so it's a premature-failure risk. Proper fix is a monotonic progress signal (persisted ever-incrementing assistant-persist counter, or a non-decreasing ordinal), applied to both Think and ai-chat. Full detail + repro in #1628.

### Verification
`npm run check` green (91/91 typecheck, lint + format clean). Reconciler unit (30), Think reconciliation (8), ai-chat reconciliation (41) suites pass.

PR #1629, opened 2026-05-31
Fix approved Think tool continuation results reaching clients

### LINKED ISSUE #1627: Bug: needsApproval tool approval-responded state never transitions to output-available (think@0.7.3 + agents@0.13.3)
### Environment
- `@cloudflare/think@0.7.3`
- `agents@0.13.3`
- `@cloudflare/ai-chat@0.7.2`

### Description

After a user approves a `needsApproval` tool, the tool executes successfully on the server, but the tool part state never transitions from `approval-responded` to `output-available` on the frontend. The UI permanently shows "Executing..." with a spinning indicator.

### Related Issues
- #1586 — Fixed `_scheduleAutoContinuation` 50ms coalesce race condition. Our issue persists after upgrading to the versions that include this fix.
- #1404 — Tool part state regression during client tool round-trip.

### Steps to Reproduce
1. Define a tool with `needsApproval: true`
2. User sends a message that triggers the tool
3. Tool approval card appears, user clicks "Approve"
4. Tool executes successfully on the server (verified via database state)
5. Frontend tool part stays at `approval-responded` state indefinitely
6. UI shows spinning "Executing..." forever
7. Only when user sends a NEXT message does `beforeTurn` clean up the stale state

### Expected Behavior
After approval, the continuation should:
1. Execute the tool
2. Update the tool part state to `output-available` with the tool output
3. Stream the result back to the frontend
4. UI transitions from "Executing..." to "Complete" with output

### Actual Behavior
- Server-side: Tool executes correctly, state changes persist
- Frontend: Tool part stays at `approval-responded`, never receives `output-available`
- The continuation stream appears to not reach the frontend

### WebSocket Evidence

After user approves the tool, the `cf_agent_chat_messages` broadcast shows the tool part stuck at:
```json
{
  "toolName": "updateTrigger",
  "state": "approval-responded",
  "approval": { "id": "...", "approved": true }
}
```

No subsequent `cf_agent_message_updated` or stream chunk updates the state to `output-available`.

### Workaround

In our `beforeTurn` override, we clean up stale `approval-responded` parts on the next non-continuation turn. This means the state only resolves when the user sends another message, not immediately after approval.

whoiskatrin:
## Summary

Fixes the `@cloudflare/think` server-side `needsApproval` continuation path so an approved tool result is delivered to the client and persisted instead of leaving the UI indefinitely at `approval-responded` / “Executing…”.

Fixes #1627.

## What was happening

For an approved server-executed tool, Think handles the approval in one message update and then runs the actual tool in an automatically scheduled continuation. There were several gaps in that handoff:

1. The durable approval update did not synchronously update Think's live message cache before the continuation read it.
2. Transcript repair treated `approval-responded`, `output-denied`, and `output-error` parts without an `output` field as abandoned tool calls and removed them from model context.
3. A server tool's eventual `tool-output-available` / `tool-output-error` chunk is emitted during the continuation, but the corresponding tool part belongs to the preceding assistant message. The continuation accumulator cannot persist that cross-message state transition by itself.
4. Think's WebSocket continuation response frames were missing the `continuation: true` marker used by `useAgentChat` when applying server-pushed continuation events to existing client state.

Together these could leave the server having performed the approved side effect while the browser remained stuck at `approval-responded`.

### `packages/think/src/think.ts`

- Keep valid non-output-bearing tool lifecycle states (`approval-responded`, `output-denied`, `output-error`) in transcript repair.
- Synchronously patch the in-memory transcript after an approval/tool update is persisted, before an immediate continuation consumes it.
- When terminal tool output arrives during a continuation for a tool part stored in a prior assistant message, apply it to that stored message and broadcast `cf_agent_message_updated`.
- Include `continuation: true` on continuation stream chunks, errors, and completion frames.

### Tests

Added an actual server `needsApproval` tool to the Think WebSocket test agent and regression tests covering:

- **Approval success:** the tool executes exactly once, broadcasts an `output-available` update, persists output, and marks completion as a continuation.
- **Rejection:** the tool is not executed and `output-denied` is retained through continuation processing.
- **Execution error after approval:** the tool is invoked once and its terminal `output-error` / error text is persisted.

### Release note

- Added a patch changeset for `@cloudflare/think`.

## Reviewer notes

The key code path to review is the transition across messages:

```text
approval-requested
  -> CF_AGENT_TOOL_APPROVAL
  -> approval-responded persisted + live cache patched
  -> continuation runs approved execute()
  -> tool-output-available/error targets original assistant message
  -> MESSAGE_UPDATED reaches frontend + terminal state persists
```

The cross-message update mirrors the situation already handled in `@cloudflare/ai-chat`: the continuation's newly accumulated assistant content is not the message that owns the original tool part.

### DIFF
--- a/.changeset/fix-think-approval-continuations.md
+++ b/.changeset/fix-think-approval-continuations.md
@@ -1,5 +1,6 @@
 ---
 "@cloudflare/think": patch
+"agents": patch
 ---
 
 Fix server-side `needsApproval` tool continuations remaining stuck after the
@@ -8,3 +9,13 @@ model transcript, updates its live transcript before an immediate continuation,
 and persists and broadcasts terminal tool output emitted for a prior assistant
 message. Continuation response frames are also labelled consistently so
 `useAgentChat` can apply streamed continuation updates to the active UI state.
+A pending `approval-responded` tool is no longer mis-reported by the
+incomplete-tool-call backstop, so approval continuations stop logging a false
+"repair gap" warning and emitting a spurious `chat:transcript:repaired` event.
+
+The cross-message tool result now flows through `StreamAccumulator`'s
+`cross-message-tool-update` action and a shared, replay-safe
+`crossMessageToolResultUpdate` builder (exported from `agents/chat`): it matches
+terminal states for first-write-wins idempotency against provider replays (e.g.
+the OpenAI Responses API, #1404), preserves a streamed `preliminary` flag, and
+lets `Think` skip redundant writes/broadcasts when a result is already settled.
--- a/.changeset/fix-think-approval-continuations.md
+++ b/.changeset/fix-think-approval-continuations.md
@@ -1,0 +1,10 @@
+---
+"@cloudflare/think": patch
+---
+
+Fix server-side `needsApproval` tool continuations remaining stuck after the
+user approves them. Think now keeps approved/denied/errored tool parts in the
+model transcript, updates its live transcript before an immediate continuation,
+and persists and broadcasts terminal tool output emitted for a prior assistant
+message. Continuation response frames are also labelled consistently so
+`useAgentChat` can apply streamed continuation updates to the active UI state.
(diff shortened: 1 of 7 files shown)


threepointone:
taking this over

Issue #1631, opened 2026-05-31
discussion: recovery/repair ergonomics so products don't need a full onChatRecovery override (persist-false footgun, ask_user preserve-as-text, onExhausted payload)


threepointone:
## Background

A production agent product (long tool-heavy turns, continuous deploys, DO eviction) carries a custom `onChatRecovery` override because the framework — even after #1623's bounded recovery — doesn't give them a few things they need for a *user-facing terminal-state policy*. Investigating their deploy-churn data-loss report surfaced three framework-level gaps. Each is independently small; together they're what forces the override (and the override is where their bug lived). Filing as one discussion so we can decide what to absorb.

All references are to `packages/agents/src/index.ts` / `packages/think/src/think.ts` / `packages/ai-chat/src/index.ts`.

---

## 1. `onChatRecovery` returning `persist: false` silently discards settled work (data-loss footgun)

**What happens.** In the recovery fiber, `onChatRecovery(ctx)` is called first, then the orphan is persisted only if `options.persist !== false`:

```ts
const options = (await this.onChatRecovery({ ... })) ?? {};
if (options.persist !== false && streamId && (streamStillActive || (streamIsTerminal && !streamAlreadyPersisted))) {
  await this._persistOrphanedStream(streamId);
}
```

So a subclass that returns `{ persist: false, continue: false }` on a *terminal* decision — a very natural-looking "give up on this turn" return — **throws away the partial assistant message**, including every settled (often non-idempotent) tool result from a 10–20-call turn. The model then re-runs all of it on the next message. The customer hit exactly this: a healthy long turn interrupted past their cap by deploy churn returned `{ persist: false }` and "went all the way back / lost the files."

**Why it's a footgun.** `persist: false` reads like "don't bother continuing," not "delete the settled work." The settled tool results are the most valuable, least-reproducible state in the turn.

**Suggestions (pick one):**
- Make the orphan persist independent of the *continue* decision — i.e. persist the settled partial whenever there is one, and treat `persist: false` as advisory only (or rename it to make the data-loss explicit).
- Or default `persist` to `true` and require an explicit `discardPartial: true` to drop settled work, with a one-time `console.warn` when a terminal return discards a partial that contained settled tool results.
- At minimum: document loudly on `ChatRecoveryOptions.persist` that `{ continue: false }` should keep `persist: true` to avoid losing settled tool calls.

---

## 2. `ask_user` / client-resolved orphans should be preservable as text, not flipped to an errored result

[...]

Issue #1628, opened 2026-05-31
bug: chat-recovery progress marker is count-based, so compaction can cause premature budget exhaustion


threepointone:
## Summary

Think's chat-recovery progress detection counts the **current** number of assistant messages. Auto-compaction reduces that count (it replaces several messages with a single summary overlay), so a recovery incident that fires after a compaction can read "no forward progress" even though the turn actually advanced — failing to reset the attempt budget and risking **premature exhaustion** under simultaneous compaction + deploy churn.

Surfaced during the deep review of #1623 (item #7). Deferred from that PR because the proper fix is a design change (a monotonic progress signal), not a quick patch.

## Where

`packages/think/src/think.ts`

`_chatRecoveryProgressMarker()` counts assistant messages:

```ts
private _chatRecoveryProgressMarker(): number {
  return this.messages.reduce(
    (count, message) => (message.role === "assistant" ? count + 1 : count),
    0
  );
}
```

`_beginChatRecoveryIncident()` uses it to decide whether to reset the attempt budget:

```ts
const prevProgress = existing?.progress ?? 0;
const currentProgress = this._chatRecoveryProgressMarker();
const madeProgress = existing != null && currentProgress > prevProgress;
// ...
const attempt = madeProgress && !windowExceeded ? 1 : (existing?.attempt ?? 0) + 1;
const exhausted = windowExceeded || attempt > config.maxAttempts;
// stored high-water mark:
progress: Math.max(prevProgress, currentProgress)
```

## The bug

The intent: a mid-turn deploy interrupts a turn that is *advancing* (each recovery persists a new assistant message via `_persistOrphanedStream`), so `currentProgress > prevProgress` → reset the attempt budget instead of burning it. This keeps a turn that's genuinely making progress from being abandoned by repeated deploys.

But `progress` is a **count** of current assistant messages, and it's stored as a **high-water mark** (`Math.max`). Auto-compaction (fires on `appendMessage` when `compactAfter` is configured) collapses multiple assistant messages into one summary overlay, so the count **drops**. After a compaction:

- `currentProgress` (post-compaction count) < `prevProgress` (pre-compaction high-water mark)
- → `madeProgress` is `false`
- → the attempt counter is **not** reset, even though the turn advanced

[...]

PR #1633, opened 2026-05-31
fix(think,ai-chat): make chat-recovery progress signal compaction-immune (#1628)

threepointone:
## Stack

This is **PR 1 of 4** addressing #1631 (recovery/repair ergonomics) and #1628. The series is stacked; review/merge bottom-up:

1. **➡️ #1628 — progress signal compaction-immune (this PR)** · patch
2. preserve settled work when a turn is given up · patch
3. `repairInterruptedToolPart` hook · minor
4. incident identity + `onExhausted` payload · minor

Each PR is independently reviewable; this one has no dependency on the others (it's the base of the stack). Targets `main`.

---

## Problem (#1628)

Bounded chat recovery (#1623) resets the per-incident retry budget when an interrupted turn is making **forward progress**, so that a turn repeatedly interrupted by deploy churn isn't falsely treated as a poison turn and sealed at `maxAttempts`. That progress signal was **recomputed from the live transcript** by counting assistant messages:

```ts
// packages/think/src/think.ts (and the verbatim mirror in packages/ai-chat/src/index.ts)
private _chatRecoveryProgressMarker(): number {
  return this.messages.reduce(
    (count, message) => (message.role === "assistant" ? count + 1 : count),
    0
  );
}
// ...in _beginChatRecoveryIncident:
const prevProgress = existing?.progress ?? 0;
const currentProgress = this._chatRecoveryProgressMarker();
const madeProgress = existing != null && currentProgress > prevProgress;
```

`this.messages` is the live, mutable transcript. **Compaction collapses older assistant messages into a summary, lowering the count.** So a turn that genuinely advanced between two recovery attempts can read as `currentProgress < prevProgress` → `madeProgress = false` → the budget is **not** reset → the turn exhausts at `maxAttempts` and is sealed, even though it was healthy. The stored `progress: Math.max(prevProgress, currentProgress)` only protects the persisted high-water; the *comparison* still uses the live (compacted) recompute, so it doesn't help.

This is the data-loss path a tool-heavy agent under continuous deploys actually hits.

## Fix

Track progress with a **durable, monotonic counter** that compaction can never lower, instead of recomputing from the transcript:

- New storage key `cf:chat-recovery:progress`.
- Bumped exactly once when `_persistOrphanedStream` materializes a **non-empty** partial assistant message — which is precisely the "forward progress" event the old assistant-message count was a proxy for.
- `_chatRecoveryProgressMarker()` now reads that counter; `_beginChatRecoveryIncident` awaits it.

Because it only ever increments, compaction is irrelevant. Semantics are otherwise preserved:

- **Healthy turn under churn:** each interruption persists a non-empty partial → counter grows → `madeProgress` true → budget resets. ✅
- **Poison turn (no chunks):** `_persistOrphanedStream` early-returns on empty chunks → counter flat → budget counts toward the cap → exhausts. ✅
- **Genuinely-stuck-but-producing turn:** still ultimately bounded by the unchanged 15-minute `CHAT_RECOVERY_MAX_WINDOW_MS` wall-clock ceiling.

### Timing (why the counter is read at incident-begin and bumped at persist)

Per interruption: `_beginChatRecoveryIncident` reads the counter and compares against the stored high-water → then (if not exhausted) `_persistOrphanedStream` bumps it. So attempt _N_'s bump is observed at attempt _N+1_'s comparison, exactly mirroring how the old assistant-message count grew after each persist.

## Back-compat / migration

A session interrupted mid-incident *before* this counter existed has no `cf:chat-recovery:progress` key → it reads `0`. For a **new** incident that's fine (no comparison on first detection). For an in-flight incident at upgrade time the comparison is conservative (treats it as no-progress for that one alarm), which is bounded and harmless. No persisted-data migration required.

## Bonus

`@cloudflare/ai-chat`'s `_persistOrphanedStream` **merges** a continuation partial into the existing assistant message in place, so its old marker barely grew across continuations (only the first persist added a message) — it was *already* under-counting progress. The monotonic counter fixes that too and brings ai-chat in line with think.

## Files

- `packages/think/src/think.ts` — counter constant, `_chatRecoveryProgressMarker` (now async, reads storage), new `_bumpChatRecoveryProgress`, bump in `_persistOrphanedStream`, `await` in `_beginChatRecoveryIncident`, updated `ChatRecoveryIncident.progress` doc.
- `packages/ai-chat/src/index.ts` — same changes; `_persistOrphanedStream` made `async` (callers updated to `await`).
- Tests + harness helpers (`bumpRecoveryProgressForTest`, `dropAssistantMessagesForTest`).
- `.changeset/chat-recovery-progress-monotonic.md` (patch · think + ai-chat).

## Tests

New regression test in **both** packages — *"detects forward progress even after compaction collapses the transcript (#1628)"*: open an incident, make progress (bump the counter), then drop every assistant message from the live cache (simulating compaction), and assert the next alarm still resets the budget. The old message-count marker fails this; the durable counter passes. Existing progress-reset and wall-clock-ceiling tests were re-pointed at the counter and still pass.

## Verification

- `tsc --noEmit` clean for `@cloudflare/think` and `@cloudflare/ai-chat`.
- `oxlint` clean; `oxfmt` applied.
- Full suites green: think **460 passed** (15 files), ai-chat **481 passed** (42 files).

## Test plan

- [ ] CI green (`npm run build && npm run check && nx affected -t test`)
- [ ] Confirm the compaction regression test fails on `main` and passes here
- [ ] Sanity-check the back-compat note for in-flight incidents at deploy time

### DIFF
--- a/.changeset/chat-recovery-progress-monotonic.md
+++ b/.changeset/chat-recovery-progress-monotonic.md
@@ -1,0 +1,16 @@
+---
+"@cloudflare/think": patch
+"@cloudflare/ai-chat": patch
+---
+
+Fix chat recovery prematurely exhausting its retry budget under compaction
+(#1628). The deploy-churn forward-progress signal — which resets the recovery
+budget when an interrupted turn is actually advancing — was recomputed from the
+live transcript by counting assistant messages. Compaction collapses older
+assistant messages into a summary, lowering that count, so a turn that had
+genuinely advanced could read as "no progress" between recovery attempts and
+exhaust at `maxAttempts`, sealing a healthy turn. Progress is now tracked by a
+durable, monotonic counter incremented when `_persistOrphanedStream` materializes
+a non-empty partial (the exact event the message count was proxying for), so
+compaction can never lower it. A turn that genuinely fails to advance still
+exhausts at the cap, and the 15-minute wall-clock ceiling is unchanged.
--- a/packages/ai-chat/src/index.ts
+++ b/packages/ai-chat/src/index.ts
@@ -114,16 +114,23 @@ type ChatRecoveryIncident = {
   lastAttemptAt: number;
   reason?: string;
   /**
-   * High-water mark of a monotonic recovery-progress signal (count of persisted
-   * assistant messages) observed for this incident. Used to distinguish a turn
-   * that is making forward progress but keeps getting interrupted by isolate
-   * resets (deploys) — which should NOT exhaust the budget — from one that
-   * genuinely fails to advance.
+   * High-water mark of the durable, monotonic recovery-progress counter (see
+   * `_chatRecoveryProgressMarker`) observed for this incident. Used to
+   * distinguish a turn that is making forward progress but keeps getting
+   * interrupted by isolate resets (deploys) — which should NOT exhaust the
+   * budget — from one that genuinely fails to advance. Sourced from a persisted
+   * counter rather than the live transcript so compaction cannot lower it
+   * (#1628).
    */
   progress?: number;
 };
 
 const CHAT_RECOVERY_INCIDENT_KEY_PREFIX = "cf:chat-recovery:incident:";
+// Durable, monotonic forward-progress counter for recovery budget resets.
+// Incremented once per non-empty partial materialized by
+// `_persistOrphanedStream`; never recomputed from the (compactable) transcript.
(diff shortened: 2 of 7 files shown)


threepointone:
/bonk do a deep review on this pr, look for edge cases, how's the ux on this

PR #1635, opened 2026-05-31
feat(think): add repairInterruptedToolPart hook for client-resolved tools (#1631)

threepointone:
## Stack

This is **PR 3 of 4** for #1631. Stacked; **base is `fix/chat-recovery-preserve-settled-work`** (PR #1634). Retargets to `main` as the stack merges. Diff shown is only this PR's commit.

1. #1633 — progress signal compaction-immune (#1628) · patch
2. #1634 — preserve settled work when a turn is given up · patch
3. **➡️ `repairInterruptedToolPart` hook (this PR)** · minor
4. incident identity + `onExhausted` payload · minor

This is a **public-API addition** (a new `protected` override point) — the one most worth a careful API-shape review. `@cloudflare/think` only (ai-chat has no transcript-repair pass).

---

## Problem (#1631, item 2)

Transcript repair (`_repairToolTranscriptParts`) flips a tool call with no settled result to an errored tool-result — preserving the record and keeping the provider from 400ing (`AI_MissingToolResultsError`). That's the right default for **server** tools.

It's wrong for **client-resolved** tools like `ask_user` — a question with no server `execute`, answered by the user's *next message*. There, the interrupted call **is** a question, and the right repair is to preserve it as a **text part carrying the prompt**, so the model sees a normal Q→A exchange and compaction keeps the question verbatim. Flipping it to `output-error` turns *\"the assistant asked X\"* into *\"a tool errored\"* and loses the prose.

Crucially, repair runs **and persists** before `beforeTurn`:

```ts
// _runInference
const history = await this._repairTranscriptForProvider(this.messages); // repairs + persists + broadcasts
const messages = await convertToModelMessages(truncated, { ... });       // builds model ctx
...
const subclassConfig = (await this.beforeTurn(ctx)) ?? {};               // too late: ctx.messages already built/persisted
```

So a subclass had **no way** to shape this for the current turn — any `beforeTurn` fix-up lands a turn late and races the framework's persisted `output-error`.

## Fix

Extract the per-orphan flip into a `protected` overridable hook that repair consults during the repair pass (before persist), defaulting to the existing behavior:

```ts
/** Repair a single interrupted tool call (no settled result). `input` is already
 *  normalized. Default flips it to an errored tool-result. Override to preserve
 *  client-resolved tools (e.g. ask_user -> a text part carrying the prompt) so the
 *  conversion shapes THIS turn, before repair persists + sends to the model. */
protected repairInterruptedToolPart(
  part: UIMessage["parts"][number]
): UIMessage["parts"][number] {
  return { ...part, state: "output-error", errorText: "The tool call was interrupted before a result was recorded." };
}
```

`_repairToolTranscriptParts` now calls `this.repairInterruptedToolPart({ ...part, input: normalizedInput })`. Default behavior is byte-for-byte unchanged. A subclass returning a text part for `ask_user` takes effect **on the same turn**, because it runs inside the repair pass.

Contract (documented on the method): a returned **tool** part must carry a settled result (so the `ignoreIncompleteToolCalls` backstop / `_incompleteToolCallIds` don't drop it); returning a **non-tool** part (e.g. text) is fine.

## Files

- `packages/think/src/think.ts` — new `protected repairInterruptedToolPart`; `_repairToolTranscriptParts` refactored to call it (default unchanged).
- `packages/think/src/tests/agents/client-tools.ts` — `ThinkClientToolsAgent` overrides the hook (ask_user → text) to demonstrate the intended usage + a `repairToolTranscriptPartsForTest` harness.
- `packages/think/src/tests/client-tools.test.ts` — tests.
- `.changeset/think-repair-interrupted-tool-part-hook.md` (minor · think).

## Tests

- *converts an interrupted `ask_user` into a text part carrying the prompt* (the override).
- *falls back to the default errored-result repair for non-`ask_user` tools* (default preserved).
- *leaves a settled `ask_user` untouched* (only interrupted calls are repaired).

The pre-existing repair tests (`message-reconciliation.test.ts`: orphan → `output-error`) still pass — the refactor is behavior-preserving for the default, and the demo override only matches `tool-ask_user` (other agents' orphans are unaffected).

## Verification

- `tsc --noEmit` clean (think); `oxlint` clean; `oxfmt` applied.
- `client-tools` + `message-reconciliation` suites: **58 passed**. Full think suite: **460 passed**.

## API review notes

- Name/shape up for discussion: `repairInterruptedToolPart(part) -> part`. Alternatives considered: a `beforeTranscriptRepair(messages)` pre-hook (rejected — needs extra persist-gating to avoid the repair's early-return swallowing hook-only changes; per-part is cleaner and persists naturally).
- This only resolves item 2 for `think`. ai-chat doesn't run this repair pass, so no change there.

## Test plan

- [ ] CI green
- [ ] API-shape sign-off on `repairInterruptedToolPart`
- [ ] Confirm default-path repair unchanged (existing repair tests green)

### DIFF
--- a/.changeset/think-repair-interrupted-tool-part-hook.md
+++ b/.changeset/think-repair-interrupted-tool-part-hook.md
@@ -1,0 +1,20 @@
+---
+"@cloudflare/think": minor
+---
+
+Add a `repairInterruptedToolPart` hook so subclasses can control how an
+interrupted tool call is repaired during transcript repair (#1631).
+
+Transcript repair flips a tool call with no settled result to an errored
+tool-result (preserving the record and keeping the provider from 400ing). That
+is the right default for server tools, but wrong for client-resolved tools like
+`ask_user` — a question with no server `execute`, answered by the user's next
+message — where the interrupted call _is_ a question and should be preserved as
+text so the model sees normal Q→A conversation and compaction keeps the prompt
+verbatim. Because repair runs (and persists) before `beforeTurn`, a subclass had
+no way to shape this for the current turn.
+
+`repairInterruptedToolPart(part)` defaults to the existing errored-result
+behavior and runs during repair, so an override (e.g. converting an interrupted
+`ask_user` into a text part carrying the prompt) takes effect on the same turn,
+not just the next one.
--- a/docs/chat-agents.md
+++ b/docs/chat-agents.md
@@ -651,7 +651,9 @@ Common return values:
 
 - `{}` — persist partial + auto-continue (default, works with providers that support assistant prefill)
 - `{ continue: false }` — persist partial but do not auto-continue (handle continuation yourself)
-- `{ persist: false, continue: false }` — handle everything yourself (e.g., retrieve a completed response from the provider)
+- `{ persist: false, continue: false }` — do not persist the unsettled remainder and handle everything yourself (e.g., retrieve a completed response from the provider)
+
+Settled work is never dropped: `persist: false` only suppresses persistence of a partial that has nothing settled to lose. A partial that already carries settled tool results (completed, often non-idempotent work) is persisted regardless, so an app cannot accidentally discard completed tool calls — and never needs `{ persist: true }` just to stay safe.
 
 When recovery happens before any stream chunks were written, there is no partial assistant message to continue. If the latest persisted message is still the unanswered user message from the interrupted turn, the framework retries that turn automatically unless `continue` is `false`.
 
--- a/docs/think/index.md
+++ b/docs/think/index.md
@@ -416,6 +416,33 @@ export class MyAgent extends Think<Env> {
 
 The same recovery events are available through `agents/observability` on the `chat` channel. Transcript repairs are emitted on the `transcript` channel.
 
+#### Repairing interrupted tool calls
+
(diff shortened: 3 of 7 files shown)


threepointone:
### Rebased onto `main` + docs added

- **Rebased** onto current `main` (dropped the now-merged #1633/#1634 commits; the diff is just this PR's change). Adapted three test-fixture accesses to the current type graph (the RPC stub return type for `repairToolTranscriptPartsForTest` stubifies the `UIMessage` union to `never[]` at the call site → explicit `as UIMessage` casts; test-only, production untouched).
- **Docs** for the new override:
  - `docs/think/index.md` — new "Repairing interrupted tool calls" subsection (default errored-result + the `repairInterruptedToolPart` override with the `ask_user`→text example). Think-only, so not added to `chat-agents.md` (ai-chat has no transcript-repair pass).
  - `packages/think/README.md` — brief recovery-section mention with the same example.
  - `docs/chat-agents.md` — corrected the now-stale `{ persist: false }` recovery-options wording to reflect the merged no-loss default (#1634/R1): settled work is never dropped.

Green: think **466**, client-tools **50**, `npm run check` clean (91 projects).

PR #1639, opened 2026-06-01
fix(mcp): safely route colliding streamable HTTP responses

### LINKED ISSUE #1632: StreamableHTTPServerTransport.send() misroutes result responses when two concurrent same-session requests share a JSON-RPC id
### Describe the bug

`StreamableHTTPServerTransport.send()` routes a JSON-RPC **result/error response** to a connection by *first match on request id*, with no tiebreak:

```js
// agents/dist/mcp/index.js — StreamableHTTPServerTransport.send()
let requestId = options?.relatedRequestId;
if (isJSONRPCResultResponse(message) || isJSONRPCErrorResponse(message)) requestId = message.id;
// ...
const connection = Array.from(agent.getConnections())
  .find((conn) => conn.state?.requestIds?.includes(requestId));   // first match wins
```

Each inbound POST to `/mcp` opens its own DO `Connection` (WebSocket) and is stamped in `handlePostRequest` with `connection.setState({ requestIds })`. When **two concurrent in-flight requests on the same session carry the same JSON-RPC id**, `find()` returns whichever connection appears first in `getConnections()` — which may be the *other* request's connection. The result is written to the wrong POST's SSE stream: one `tools/call` receives the response that belongs to a different, concurrent `tools/call`.

Because the misroute is by id only, the transport silently delivers a *plausible-looking* wrong response (a valid result for a different request) rather than erroring — so it surfaces as data crossing between requests, not as a transport failure.

The transport already has the information needed to route correctly: the originating connection is available in the `onmessage` async-context (it's exactly what `handlePostRequest` reads via `getCurrentAgent().connection`, and what #1514 tracked with ALS before being reverted). `send()` chooses to route by id-find instead.

### To Reproduce

*(Confirmed live on a deployed Worker, 2026-05-31 — see Additional context for the captured collision and the three affected-client reports.)*

1. Stand up an `McpAgent` served via `McpAgent.serve("/mcp")` (streamable-HTTP) with any tool that returns a distinguishable body per argument (e.g. a `read`/`echo` tool).
2. Initialize one session; capture its `mcp-session-id`.
3. On that one session, issue **two concurrent POSTs** to `/mcp`, each a `tools/call` for a *different* argument, but both using the **same JSON-RPC `id`** (e.g. both `"id": 1`). This is the spec-violating condition the transport does not defend against; it arises in practice when independent request-id sequences share one session id (see Additional context).
4. Read each POST's SSE response stream.

**Observed:** at least one POST receives the result generated for the *other* POST's arguments (responses cross). **Retrying serially returns the correct bodies.**

A focused regression test could assert: given two live connections whose `state.requestIds` both contain `1`, `send({id:1, result})` must reach the connection that originated id `1` — not merely the first match.

### Expected behavior

A result/error response is delivered to the connection that **originated** that request, never to a different concurrent connection. If two live requests genuinely share an id within one session, the transport should route by the originating-connection context (as #1514 attempted) and/or reject/log the collision — never silently misroute a result to the wrong stream.

### Screenshots

N/A — transport-level; reproducible via `wrangler tail` + per-POST SSE capture.

### Version

`agents@0.12.3`. **Verified unchanged in `agents@0.13.3`** (latest at time of writing): `StreamableHTTPServerTransport.send()`'s result-response `find`-by-id is byte-for-byte identical, and `McpAgent.serve()` still binds this transport. Independent of MCP SDK version (the logic is in `agents`' `send()`, not the SDK).

### Additional context

- **Real-world trigger — confirmed.** Multiple **Claude Desktop** conversations sharing one `mcp-session-id`, each numbering JSON-RPC requests from `1`. Server-side `wrangler tail` captured one shared session with many connections all carrying `requestIds:[1]` and two `mcp_request_id_collision` events (two id-`1` requests in flight at once). A spec-compliant single client keeps ids unique within a session and never collided in the same window — so the exposure is precisely session-id reuse across independent id sequences, which the first-match router turns into silent cross-request data exposure rather than an error.
- **Confirmed crossings from three affected conversations (identifiers genericized):** All three report writes landing on the correct note (id-checked re-reads) — the fault is response correlation only, and it crosses both note *and* operation schema:
  1. `replace_body` on note A (id-A) returned a **`read_note`-shaped** payload (raw body + `{permalink, frontmatter}`) for note B (id-B) — wrong note *and* wrong operation schema. The write committed correctly (verified by re-read).
  2. `replace_body` for note C (own id-C) returned a write-ack envelope for note D (id-D, etag-D). Note C re-read held its own payload/id intact.
  3. In ~9 calls during overlap: (a) `replace_body` on note D (id-D) returned a write-ack for note C (id-C); (b) `read_note` on note B (id-B) returned a bare write-ack JSON for note A (id-A) — a foreign `replace`-style response, no body. Six sequential writes with unique body tokens produced zero crossings; only overlapping windows crossed.
- **Detection signature** (useful for a regression assertion or a triage grep): a returned `id`/`etag`/`path` that does not belong to the requested target; the foreign id/etag identifies the session whose response was mis-delivered.
- **Severity framing.** In the observed case it was same-identity (one user's own notes crossing between their own sessions), not cross-tenant — but the same mechanism would expose responses across any two id-colliding requests routed to one DO, so it's worth hardening regardless of how the collision arises.
- **Relationship to existing work — and an honest caveat.** #1558 fixed the *RPC* transport's single pending-slot, which clobbered concurrent calls **even with unique ids** — an unambiguous bug. Streamable-HTTP's result-response `find`-by-id is **not** the same: under the JSON-RPC/MCP guarantee that request ids are unique within a session, `find`-by-id is correct. It misroutes only when ids collide, which requires a client to violate that guarantee (or independent id sequences to share one session id). So a maintainer could reasonably respond "the client must use unique request ids within a session" and decline. The case for hardening anyway: the failure mode is *silent cross-request data exposure* (a valid-looking wrong result), not an error; #1514 (reverted by #1532) already showed the transport can track the originating connection via ALS; and #1513 concluded the correlation belongs with whoever owns the parent-child relationship. Frame the ask as "route result responses to the originating connection (or detect + error on a duplicate in-flight id) instead of silently first-matching" — not as parity with #1558.
- **Both layers are implicated.** The trigger is a client (Claude Desktop) reusing one session id with non-unique request ids; the defect reported *here* is that this transport silently misroutes on the resulting collision instead of routing by originating connection or rejecting the duplicate. A parallel client-side report (Claude Desktop should not multiplex independent conversations onto one MCP session with colliding ids) addresses the trigger; this issue addresses the transport hardening.

whoiskatrin:
## Summary

Fixes #1632 by preventing `StreamableHTTPServerTransport.send()` from routing a JSON-RPC result/error response to an unrelated POST stream when concurrent requests on one MCP session reuse the same request id.

Although duplicate in-flight ids violate the client-side correlation contract, the current fallback is unsafe: it silently delivers a valid-looking response to whichever matching connection is returned first. This change hardens that failure mode rather than allowing responses to cross request streams.

## Approach

- Prefer the live connection from the current agent async context when multiple live streams claim the same request id.
- Preserve resumed POST-stream delivery: the originating connection is preferred only while it is still live, so a GET connection that inherited `requestIds` after `Last-Event-ID` resumption can continue receiving the response.
- When multiple live matching streams exist and no originating live connection can disambiguate them, return JSON-RPC `-32603 Internal error` responses to the affected streams instead of guessing or producing a transport-level failure.
- Track batched POST response completion by stream rather than globally by request id, so a collision on another POST cannot interfere with stream closure bookkeeping.

## Reviewer pointers

- `packages/agents/src/mcp/transport.ts`: the routing selection immediately after standalone SSE handling is the key behavioral change. The live-origin check is intentional because `handleGetRequest()` can replace a disconnected POST stream with a resumed GET stream carrying the same `requestIds`.
- `packages/agents/src/mcp/transport.ts`: `sendOnStream()` ensures the ambiguity-generated protocol error follows the same event-store and stream-completion path as ordinary responses.
- `packages/agents/src/mcp/transport.ts`: `_streamResponseIds` is separate hardening for batched requests; the previous request-id-only map shared completion state across colliding streams.
- `packages/agents/src/tests/mcp/transports/streamable-http.test.ts`: includes the same-id concurrent POST scenario from the issue plus the stale-origin/resumed-stream and no-origin protocol-error edge cases.
- `packages/agents/src/tests/agents/mcp.ts`: `collisionBarrierEcho` forces both colliding requests to remain in flight before either response is emitted, making the collision deterministic.

### DIFF
--- a/packages/agents/src/mcp/transport.ts
+++ b/packages/agents/src/mcp/transport.ts
@@ -376,6 +376,47 @@ export class StreamableHTTPServerTransport implements Transport {
     this.onclose?.();
   }
 
+  private async sendOnStream(
+    agent: McpAgent,
+    connection: Connection<TransportConnState>,
+    message: JSONRPCMessage,
+    requestId: RequestId
+  ): Promise<void> {
+    const streamId = connection.state?.streamId ?? connection.id;
+
+    let eventId: string | undefined;
+
+    if (this._eventStore) {
+      eventId = await this._eventStore.storeEvent(streamId, message);
+    }
+
+    let shouldClose = false;
+
+    if (isJSONRPCResultResponse(message) || isJSONRPCErrorResponse(message)) {
+      let responseIds = this._streamResponseIds.get(streamId);
+      if (!responseIds) {
+        responseIds = new Set<RequestId>();
+        this._streamResponseIds.set(streamId, responseIds);
+      }
+      responseIds.add(requestId);
+      const relatedIds = connection.state?.requestIds ?? [];
+      // Check if we have responses for all requests using this connection.
+      shouldClose = relatedIds.every((id) => responseIds.has(id));
+
+      if (shouldClose) {
+        this._streamResponseIds.delete(streamId);
+
+        // POST stream is fully responded — drop the persisted requestIds
+        // mapping and the stored events. No client should resume past
+        // this point.
+        await agent.deleteStreamRequestIds(streamId);
+        const clearable = this._eventStore as ClearableEventStore | undefined;
+        await clearable?.clearStream?.(streamId);
+      }
+    }
+    this.writeSSEEvent(connection, message, eventId, shouldClose);
+  }
+
   async send(
     message: JSONRPCMessage,
     options?: { relatedRequestId?: RequestId }
@@ -446,47 +487,26 @@ export class StreamableHTTPServerTransport implements Transport {
       (matchingConnections.length === 1 ? matchingConnections[0] : undefined);
     if (!connection) {
       if (matchingConnections.length > 1) {
-        throw new Error(
-          `Multiple connections established for request ID without an originating connection: ${String(requestId)}`
+        // No candidate can safely receive the original result. Terminate each
+        // affected request with a protocol error rather than returning a 500
+        // or exposing another request's plausible-looking response.
+        const routingError: JSONRPCMessage = {
+          jsonrpc: "2.0",
+          id: requestId,
+          error: { code: -32603, message: "Internal error" }
+        };
+        await Promise.all(
+          matchingConnections.map((candidate) =>
+            this.sendOnStream(agent, candidate, routingError, requestId)
+          )
         );
+        return;
       }
       throw new Error(
         `No connection established for request ID: ${String(requestId)}`
       );
     }
 
-    const streamId = connection.state?.streamId ?? connection.id;
-
-    let eventId: string | undefined;
-
-    if (this._eventStore) {
-      eventId = await this._eventStore.storeEvent(streamId, message);
-    }
-
-    let shouldClose = false;
-
-    if (isJSONRPCResultResponse(message) || isJSONRPCErrorResponse(message)) {
-      let responseIds = this._streamResponseIds.get(streamId);
-      if (!responseIds) {
-        responseIds = new Set<RequestId>();
-        this._streamResponseIds.set(streamId, responseIds);
-      }
-      responseIds.add(requestId);
-      const relatedIds = connection.state?.requestIds ?? [];
-      // Check if we have responses for all requests using this connection.
-      shouldClose = relatedIds.every((id) => responseIds.has(id));
-
-      if (shouldClose) {
-        this._streamResponseIds.delete(streamId);
-
-        // POST stream is fully responded — drop the persisted requestIds
-        // mapping and the stored events. No client should resume past
-        // this point.
-        await agent.deleteStreamRequestIds(streamId);
-        const clearable = this._eventStore as ClearableEventStore | undefined;
-        await clearable?.clearStream?.(streamId);
-      }
-    }
-    this.writeSSEEvent(connection, message, eventId, shouldClose);
+    await this.sendOnStream(agent, connection, message, requestId);
   }
 }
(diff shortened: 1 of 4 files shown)


packages/agents/src/mcp/transport.ts
```diff
@@ -424,14 +427,29 @@ export class StreamableHTTPServerTransport implements Transport {
...
+    );
+    const connection =
+      matchingConnections.find(
+        (conn) => conn.id === originatingConnection?.id
+      ) ??
+      (matchingConnections.length === 1 ? matchingConnections[0] : undefined);
     if (!connection) {
+      if (matchingConnections.length > 1) {
+        throw new Error(
+          `Multiple connections established for request ID without an originating connection: ${String(requestId)}`
```
mattzcarey: instead of throwing here I think better to return json-rpc error over the wire. 

```
Internal error, -32603
```

mattzcarey (approved):
tentative stamp. I would return that error but its probs not the end of the world either way. this will 500 instead.

packages/agents/src/mcp/transport.ts
whoiskatrin: Addressed in d0c6732f. The ambiguous no-origin path now sends a JSON-RPC `-32603 Internal error` response on each affected stream rather than throwing/producing a 500; it does not send the underlying result to either candidate. I factored response delivery through `sendOnStream()` so these protocol errors use the normal stream completion and event-store behavior, with regression coverage for the wire response.

PR #1640, opened 2026-06-01
fix(agents,think,ai-chat): re-attach to still-running sub-agent runs on parent recovery (#1630)

### LINKED ISSUE #1630: bug: sub-agent (agent-tool) runs are abandoned and re-run on parent recovery instead of re-attaching
## Summary

When a parent agent is interrupted (deploy / Durable Object eviction) **while a child agent-tool run is still in flight**, parent recovery marks the run `interrupted` within a ~5s budget and the parent re-issues the task — **re-running the child's already-completed work**. For long-running children (50–70s child turns observed in production) this is the common case under continuous deployment, and surfaces to users as "the agent went all the way back and lost the files it already wrote."

Surfaced while investigating a customer deploy-churn rollback report. (That customer's app turned out not to use sub-agents, so this is not their specific regression — but it is a genuine stock-Think bug, and it matches the production field-notes gap "a poison/long sub-agent blocks or is abandoned by its parent.")

## Root cause (all in `packages/agents/src/index.ts` unless noted)

1. **Single, short reconciliation pass.** On `onStart`, `_scheduleAgentToolRunRecovery` → `_reconcileAgentToolRuns` inspects each non-terminal child **once** within `DEFAULT_AGENT_TOOL_RECOVERY_TIMEOUT_MS = 2000` / `..._TOTAL_TIMEOUT_MS = 5000` (`~903`). A child still `running`/`starting` after that window is finalized `interrupted` (`~8005`):

   ```ts
   if (!inspection || inspection.status === "running" || inspection.status === "starting") {
     result = { runId: row.run_id, agentType: row.agent_type, status: "interrupted",
       error: "Agent tool run was still running, but live-tail reattachment is not supported in this runtime." };
   }
   ```

2. **The interrupted terminal write blocks later repair.** `_updateAgentToolTerminal` (`~7536`) only updates rows whose status is NOT already terminal, so once a row is `interrupted`, a child that *later* completes (via its own `chatRecovery`) never repairs the parent row.

3. **Re-issue can't re-attach.** `agentTool()` does not pass a `runId` (it only passes `parentToolCallId: executeOptions?.toolCallId`), so `runAgentTool` generates a fresh `nanoid(12)` → a brand-new child. Even with a *stable* runId (the documented "correct pattern"), `runAgentTool` on a duplicate **non-terminal** runId calls `_replayAndInterruptAgentToolRun` (`~7840`) — replays stored chunks then returns `interrupted` — rather than re-attaching to the live child.

4. **Lost in-memory promise.** The original `runAgentTool` await is gone after a restart; nothing bridges it (recovery runs in `ctx.waitUntil`, not by resuming the old promise). The child, however, is a **separate facet with its own `chatRecovery`**, so it self-completes its interrupted turn — the parent just never collects that result.

Net: the child's work is durable in the child facet (`cf_agent_tool_child_runs` in Think / `cf_ai_chat_agent_tool_runs` in ai-chat), but the parent abandons the run and the model re-runs it.

## Why the existing e2e looks "BOUNDED"

`packages/think/src/e2e-tests/task-amplification.test.ts` stays BOUNDED only because its `runTask` hard-codes a **stable** `CHILD_TASK_RUN_ID` ("the correct pattern") AND the child finishes inside the recovery window. A natural `agentTool()` re-issue (fresh runId) — or any child longer than ~5s — defeats both conditions and amplifies.

## Proposed fix (bounded re-attach)

Make a still-running **recoverable** child re-attach to its real terminal result instead of being abandoned. Both halves are needed — (b) keeps the row non-terminal so (a)'s re-issue can re-attach:

- **(a) `runAgentTool` duplicate non-terminal runId → re-attach.** Replace `_replayAndInterruptAgentToolRun` for the live-duplicate case with: resolve the child adapter, tail the live child to terminal (reuse `_forwardAgentToolStream` + `inspectAgentToolRun` + `_terminalResultFromInspection` + `_finishAgentToolRun`), return the real result. Fall back to `interrupted` only if there's no `tailAgentToolRun` adapter or it never reaches terminal. Makes the stable-runId pattern robust.
- **(b) Reconciliation re-attach.** For a still-running child whose adapter supports `tailAgentToolRun`, tail-to-terminal within a **generous bounded** budget (new `DEFAULT_AGENT_TOOL_REATTACH_TIMEOUT_MS`, e.g. 120_000) — or a `this.schedule()` defer-and-poll loop mirroring `_rescheduleRecoveryAfterStableTimeout` from #1623 — instead of marking `interrupted` at 5s. Only `interrupted` once the budget is exhausted.

Open design question: generous-bounded re-attach (e.g. 120s then `interrupted`) vs. a stricter cap; and whether the deferred poll should consume the same wall-clock bound as chat recovery.

## Scope / risk

- Changes agent-tool recovery **timing semantics** — re-attach can run for the child's lifetime inside `waitUntil`, so it MUST stay bounded so a genuinely hung child can't block forever.
- Reworks the tests that lock the current behavior:
  - `reconcileRunningThinkChildForTest` (in `packages/think/src/tests/agent-tools.test.ts` and `packages/ai-chat/src/tests/agent-tools.test.ts`) — currently assert `status: "interrupted"` with the "live-tail reattachment is not supported" message. New expectation: a child that completes during reattach → `completed`.
  - `reconcileStuckThinkChildWithTimeoutForTest` — asserts `elapsedMs < 1000`; a genuinely stuck child must still end `interrupted` but after the bounded budget (thread a small timeout through the test seam so the test stays fast).
- Think and ai-chat both implement `tailAgentToolRun` and child `chatRecovery`, so both can self-complete and be tailed.

## TDD plan

1. **e2e repro (the gate):** a variant of `task-amplification` where `runTask` uses a **natural fresh runId** (the `agentTool()` path) and/or a child longer than the recovery window → assert it currently **AMPLIFIES**, BOUNDED after the fix.
2. Implement (a), then (b); rework the two reconcile unit tests; add a "child completes during reattach → parent row completed" unit test.
3. Confirm the existing `task-amplification` (stable-runId) test still BOUNDED.

## References

- `packages/agents/src/index.ts`: `runAgentTool` (~7102), `_replayAndInterruptAgentToolRun` (~7840), `_reconcileAgentToolRuns` (~7946) + still-running branch (~8005), `_inspectAgentToolRunForRecovery` (~8068), `_scheduleAgentToolRunRecovery` (~8094), `_finishAgentToolRun` (~7491), `_updateAgentToolTerminal` (~7536), `_forwardAgentToolStream` (~7671), constants (~903).
- Child adapters: `packages/think/src/think.ts` (`cf_agent_tool_child_runs`, `tailAgentToolRun`, child `chatRecovery`), `packages/ai-chat/src/index.ts` (`cf_ai_chat_agent_tool_runs`, force-terminal-on-inspect when no abort controller).
- Tests: `packages/think/src/tests/agent-tools.test.ts`, `packages/ai-chat/src/tests/agent-tools.test.ts`, `packages/think/src/e2e-tests/task-amplification.test.ts`.
- Related: #1623 (recovery hardening; `AgentToolFailure.retryable` — `interrupted → retryable: true` — is the parent-facing half of this), and the field-notes gap on synchronous child agents.

threepointone:
Closes #1630.

## Problem

When a parent agent is interrupted (deploy / Durable Object eviction) **while a child `agentTool()` run is still in flight**, parent recovery sealed the run `interrupted` within a ~5s window and the parent re-issued the task — **re-running the child's already-completed work**. For long-running children under continuous deploys this surfaced to users as *"the agent went all the way back and lost the files it already wrote."*

Root causes (all in `packages/agents/src/index.ts` unless noted):

1. **Fresh runId per call.** `agentTool()` minted a fresh `nanoid` per call, so a turn re-run by chat recovery spawned a **brand-new child** instead of re-using the in-flight one.
2. **Abandon, don't re-attach.** A duplicate non-terminal run (`runAgentTool`) and a still-running child during startup reconciliation were sealed `interrupted` at ~5s instead of being tailed to their real terminal result.
3. **No durable child-run reconcile.** A child facet self-heals its interrupted turn via its own `chatRecovery`, but that path never wrote the child's agent-tool run row, so the row stranded `running` (think) / was force-errored (ai-chat) — the parent could never collect the recovered result.
4. **`interrupted` was an immutable terminal.** Once sealed `interrupted`, a later child completion could never repair the parent row, so a re-issue returned stale `interrupted` and the model retried with a new toolCallId → fresh child → re-ran work.

### 1. Stable child runId — `agentTool()` (`agent-tools.ts`)
Derive the child `runId` from the (recovery-preserved) tool call id: `agent-tool:<toolCallId>` instead of a fresh `nanoid`. A turn re-run by recovery resolves to the **same idempotent child facet**, so completed child work is never re-run. Falls back to a fresh id only when there's no tool call id.

### 2. Bounded re-attach (`runAgentTool` + `_reconcileAgentToolRuns`)
A duplicate non-terminal `runId` and a still-running child during startup reconciliation now **tail the live child to its real terminal result** and collect it, instead of immediately sealing `interrupted`. Bounded by a generous internal wall-clock budget `DEFAULT_AGENT_TOOL_REATTACH_TIMEOUT_MS` (120s), consistent with the wall-clock-dominant chat-recovery budget (#1638): a child that keeps advancing toward terminal within the window is collected; a genuinely hung child still seals `interrupted` so recovery can never block forever.

`_reconcileAgentToolRuns` is a **two-pass** sweep: a deadline-bounded inspect/classify, then **parallel** re-attach of still-running children (each with its own budget) — so one slow/hung child can't starve every later sibling against the shared deadline.

### 3. Durable child-run reconcile — think + ai-chat (transcript-based, symmetric)
Both packages reconcile a stale child-run row (`running`, no live abort controller = the original isolate is gone) from the **durable transcript** on `inspectAgentToolRun`, gated on recovery state via `_classifyAgentToolChildRecovery()` (lists chat-recovery incidents on the 1:1 child facet):

- recovery `in-progress` → keep `running` so the parent's bounded re-attach keeps waiting;
- settled with an assistant turn (text **or** tool-only) → `completed` (the parent collects the real result);
- `failed`/empty → `error`.

This keeps the child's own (working) `saveMessages` recovery path **untouched**.

> **Why not the durable submission registry?** We first tried running the child via `submitMessages({ submissionId: runId })` so the durable submission row would be the recovery-reconciled terminal marker. The unit suite stayed green, but the **real-eviction e2e caught a regression**: `_recoverSubmissionsOnStart`'s "error a `running` submission whose messages were applied" policy fights the chat-recovery fiber resuming a multi-restart, multi-step tool loop — the child stalled at 10/30 steps and the parent collected `error`. Submissions are a clean *single-turn* durable primitive but the wrong execution path for resumable agent-tool loops. **Reverted** in favor of the transcript reconcile, which never touches the recovery path. (Tracked as base-API follow-ups N7/N8 in the working doc.)

### 4. `interrupted` is now a soft, repairable terminal (`_updateAgentToolTerminal`)
`_updateAgentToolTerminal` no longer excludes `interrupted` from its overwrite guard (only `completed`/`error`/`aborted` are hard/immutable), and `runAgentTool` routes an existing `interrupted` run through the re-attach path. So a re-issue (or later child completion) **repairs** the row to the child's real result instead of returning stale `interrupted` — implements #1630's "fix `_updateAgentToolTerminal` so a later child completion can repair the parent row."

## Edge cases handled (from review)

- **Multi-child starvation** — parallel re-attach (above).
- **Text-less completed runs** — collected as `completed` when the recovered turn produced **any** assistant message (text or tool-only), not only when there's final text.
- **Chunk re-replay** — re-attach tails from the child's **current max chunk sequence**, not `-1`, so it forwards only new chunks (a reconnected client already replays stored chunks via `_replayAgentToolRuns`; the client reducer appends by arrival order).
- **Hung-child reader** — on budget-abort the tail reader is abandoned (a pre-caught read), not cancelled: cancelling a remote child-facet RPC stream surfaces an unswallowable "Stream was cancelled" pump rejection (verified). The hold is bounded — the child reaches terminal within its own chat-recovery ceiling, firing the tail's closer which releases the reader.
- **No premature terminalize** — `_checkRunFibers()` (creates the recovery incident) is `await`ed in `onStart` before the facet serves RPC, so a stale-row inspect always sees the in-progress incident.

## API / surface

- No new public configuration (budget is an internal constant; the existing `maxAttempts` stays the one knob).
- Adds an internal `agent_tool:recovery:reattach` observability event.
- `agentTool()` runId derivation and the re-attach behavior are internal.

## Tests

- **think unit (`agent-tools.test.ts`):** reworked still-running → `completed` via re-attach; new bounded-tail-able-stuck → `interrupted` after budget; new **parallel** reconcile (hung child can't starve a sibling); new **interrupted-repair** (re-issue collects `completed`). Plus the `agents` test that `agentTool()` derives a stable runId.
- **ai-chat unit:** the reconcile classify/`completed`/`error`/`in-progress` paths.
- **Real-eviction e2e (`task-amplification.test.ts`):** drives both the hand-picked stable-runId parent **and** the natural `agentTool()` parent through real kill/restart churn, asserting the child reaches all 30 steps **and** the parent's run row settles `completed` (`parentChildStatus`). Both scenarios pass.
- **Real-deploy harness (`examples/deploy-churn`, operator-run):** a new `--mode subagent` drives a child via `agentTool()` and fires real `wrangler deploy`s mid-child-loop. Run with 2 mid-loop deploys, it confirms the fix holds under *real* deploys — the child completed all 30 steps with **1** re-run (**no amplification, no data loss**). It also surfaced the follow-up below.

## Verification

- `tsc` clean (agents + think + ai-chat); `oxlint` + `oxfmt` clean.
- Full suites green: think **458**, ai-chat **480**; `agents` agent-tool tests green.
- Both real-eviction e2e scenarios pass (~224s).

## Known follow-ups (tracked, not in this PR)

- **Orchestrating-parent recovery progress (surfaced by the real-deploy harness).** A parent turn that merely `await`s a sub-agent makes no forward progress of its own, so under heavy deploy churn the **parent's own** chat recovery can exhaust (`stable_timeout`, `progress: 1`) *before the child finishes* — the parent then collects `interrupted` even though the child completed (its work preserved, no amplification). Fix (= plan's N1 "apply recovery to the sub-agent path"): count the child's forwarded progress as parent progress, and/or re-attach a still-recoverable `interrupted` row on a later restart. Beyond this PR's re-attach scope → separate PR.
- **ai-chat real-eviction e2e** — blocked on building a deterministic, LLM-free, `streamText`-compatible mock child that completes after recovery (an AI-SDK harness problem orthogonal to this change; ai-chat's reconcile is unit-validated).
- **N7** — durable submissions can double-recover / prematurely error a `chatRecovery` turn under churn (the latent base-API bug behind the reverted submission experiment).
- **N8** — unify the durable "turn status/result" record (transcript vs deleted-on-success incident vs submissions).

## Test plan

- [ ] CI green
- [ ] Review the re-attach budget (120s) and the `agent-tool:<toolCallId>` runId derivation
- [ ] Validate against `examples/deploy-churn` + the customer chaos harness

Made with Cursor


threepointone:
/bonk do a deep review, look for edge cases, is the ux/dx good

Issue #1637, opened 2026-06-01
recovery budget should be wall-clock-keyed-to-progress + alarm-debounced, not raw attempt count (deploy churn)


threepointone:
## Problem

Bounded chat recovery (#1623) caps retries by **attempt count** (`attempt > maxAttempts`, default 6) plus a 15-min incident-age wall-clock, with a progress-aware reset (#1628/#1633). Under realistic deploy churn the **attempt count is the wrong primary bound**, and a healthy in-flight turn gets sealed even though it's advancing.

### Producer-side evidence (customer session)
One prompt → **23 successful LLM calls / 16.8k output tokens / ~6 min**, **0/23 errored**, 0 MB (not OOM). It exhausted recovery and sealed. Why:

- **One deploy ≠ one alarm.** A rollout takes ~11–22s and the WebSocket drops/reconnects several times — ~15 real disconnects from ~10 deploys; **4 disconnects in 37 s** when deploys bunched (dashboard 20:39:19 → project 20:39:39 → agent 20:40:01). Each reconnect/eviction can fire a recovery alarm, so the attempt counter inflates far faster than the real interruption rate.
- **Retries aren't independent.** Each eviction forces a cold re-run with growing context and no prompt cache → each attempt is *slower* (20→35 s) and less likely to complete a unit of work before the next deploy. So the geometric "raise the cap" intuition doesn't hold — more attempts of a degrading process buys little.

Net: the cap trips on a reconnect/eviction *storm*, not on genuine repeated failure. (Companion to #1631; #1628 fixed the progress *signal*, this fixes the *budget shape*.)

## Proposed change (decided design)

Make the budget **wall-clock-keyed to forward progress**, with the attempt cap demoted to a secondary backstop:

1. **Primary bound: no-progress wall clock = 5 min.** Track `lastProgressAt` on the incident; reset it to `now` on every forward-progress attempt. Exhaust when `now - lastProgressAt > 5min`. A turn that produces *anything* within 5-min windows survives indefinitely under churn; a turn that goes quiet for 5 min is sealed.
2. **Alarm debounce ≈ 30 s.** Consecutive recovery alarms within the debounce window (covers a rollout's reconnect storm) collapse into **one** attempt instead of N.
3. **Attempt cap = high secondary backstop** (raise default, e.g. ~10) so the 5-min window binds first under normal cadence; the cap only catches a pathological tight alarm-loop. It still resets on progress.
[...]

PR #1641, opened 2026-06-01
fix(agents,think,ai-chat): count a sub-agent's progress as the orchestrating parent's recovery progress (N9)

threepointone:
## Summary

A parent turn whose work is *"run a sub-agent and `await` its result"* produces **no recoverable content of its own**. Under deploy churn, the **parent's own** chat-recovery no-progress window could therefore exhaust while the child was still healthily streaming — the parent abandoned its turn as `interrupted` and collected an interrupted result *even though the child went on to complete*.

This is the other half of the sub-agent recovery story: #1630/#1640 made the child **self-heal** and made the parent **re-attach** to it, but the parent's *own* recovery budget still expired while it waited. It's what still forced a customer-side `MAX_TASK_TURN_RETRIES` cap.

Reproduced by the `examples/deploy-churn --mode subagent` harness with real `wrangler deploy`s mid-loop: the parent exhausted at **`attempt 6/6`, `progress: 1`** while the child self-healed all 30 steps with no amplification.

## Root cause

`_forwardAgentToolStream` (base `Agent`) only **relayed** child chunks to the parent's connections via `_broadcastAgentToolEvent`. It never touched the parent's durable recovery-progress marker (`_bumpChatRecoveryProgress`), which lives in the chat layer (`Think` / `AIChatAgent`), not in base `Agent`. So an awaiting parent banked **zero** forward progress, and the no-progress window (`lastProgressAt`, #1638) tripped.

## Fix

- Add a base-`Agent` seam **`_onAgentToolStreamProgress()`** (no-op by default), invoked by `_forwardAgentToolStream` **once per read iteration that actually forwarded ≥1 child chunk**.
- `Think` and `AIChatAgent` **override** it to bump their recovery-progress marker, throttled in-memory (`AGENT_TOOL_STREAM_PROGRESS_BUMP_THROTTLE_MS = 5s`, reset per isolate — so the **first forwarded chunk after any restart always credits**, guaranteeing each recovery attempt that observes child output registers progress).
- Covers **both** the live `runAgentTool` path and the recovery `_reconcileAgentToolRuns` / re-attach path (all route through `_forwardAgentToolStream`).
- The bump is **best-effort** — wrapped so a transient failure can never break the child stream the user is watching.

### Design guard (the important invariant)

Credit is granted **only on a genuinely-forwarded chunk** — never merely because a child is attached. A **silent / hung child forwards nothing → no credit → the parent still exhausts on its own no-progress timer**, so a stuck sub-agent can never pin a parent's recovery open forever. (Covered by a dedicated unit test.)

## No new public API / config

Zero new app-facing surface — a parent using `agentTool()` / `runAgentTool()` now survives churn while its child streams, automatically. The seam is an internal (`_`-prefixed) protected hook for the framework's own chat-recovery agents.

## Validation (all three layers green)

- **Unit (think + ai-chat, deterministic):** forwarding a child stream via the real `_forwardAgentToolStream` advances the durable marker and **resets the parent's attempt budget** exactly like in-band content (mirrors the existing #1637/#1638 budget-reset test); a **silent** child (0 chunks) does **not** credit, so the cap still binds. New harness seam `forwardChildStreamProgressForTest` in both packages. Full suites green (think 460, ai-chat 482).
- **SIGKILL e2e:** the `task-amplification` natural-`agentTool()` scenario stays **BOUNDED** (child 30/30, 1 bounded re-run) and the parent collects `completed` — **no regression** to #1640's re-attach.
- **Real-deploy (`deploy-churn --mode subagent`, 3 mid-loop `wrangler deploy`s):** verdict flipped **`NOT COLLECTED` → `RE-ATTACHED`** — `parentChildStatus: completed`, child 30/30, and the parent's recovery incident shows **`progress: 8`, `attempt: 5/6` (did not exhaust)** vs the prior `progress: 1, attempt 6/6 → interrupted`.
- Also fixed a harness gap: `deploy-churn`'s `reset()` now clears the parent's chat transcript, so a **reused** session no longer silently short-circuits the sub-agent mock model to "DONE".
- `npm run check` clean (sherif, exports, oxfmt, oxlint, typecheck — 91 projects).

## Follow-up (deliberately out of scope — tracked as N10)

N9 keeps the parent alive *while the child streams new output*; it is not an infinite hold (by design — a silent child must let the parent exhaust). Under **extreme** churn where the child itself is repeatedly reset and produces no *new* forwarded chunks between two parent attempts, the parent can still climb to the cap (the real-deploy run collected at `attempt 5/6` — one beat from exhausting). The hardening backstop — **re-attach a still-recoverable `interrupted` row on a later restart** in `_reconcileAgentToolRuns` (using #1640's soft-`interrupted` + N6's transcript-reconcile to repair it to `completed`) — is left as a separate follow-up because it touches the same fragile reconcile code and needs its own real-deploy gate.

## Semver

Marked all three packages `patch` — this is a behavior bug fix with no new public API (the seam is an internal protected hook). Happy to bump `agents` to `minor` if reviewers prefer to treat the new protected hook as additive.

## Test plan

- [x] think + ai-chat unit suites green
- [x] SIGKILL `task-amplification` e2e BOUNDED + parent collects `completed`
- [x] `deploy-churn --mode subagent` real-deploy verdict = RE-ATTACHED
- [x] `npm run check` clean

### DIFF
--- a/.changeset/parent-progress-on-child-stream.md
+++ b/.changeset/parent-progress-on-child-stream.md
@@ -1,0 +1,28 @@
+---
+"@cloudflare/think": patch
+"@cloudflare/ai-chat": patch
+"agents": patch
+---
+
+Count a sub-agent's progress as the orchestrating parent's recovery progress
+
+A parent turn whose work is "run a sub-agent and await its result" produced no
+recoverable content of its own, so under deploy churn the **parent's** own
+chat-recovery no-progress window could exhaust while the child was still
+healthily streaming — abandoning the turn as `interrupted` and collecting an
+interrupted result even though the child went on to complete. (Reproduced by
+the `examples/deploy-churn --mode subagent` harness: the parent exhausted at
+`attempt 6/6` with `progress: 1` while the child self-healed all 30 steps.)
+
+Forwarding a child's stream to the parent's connections is now treated as
+genuine forward progress for the parent's recovery budget: `Think` and
+`AIChatAgent` advance their durable recovery-progress marker (throttled) each
+time `_forwardAgentToolStream` forwards child output, so a parent that keeps
+re-attaching to and streaming a live child survives churn indefinitely. The
+credit is only granted when the child actually produces output — a silent or
+hung child still lets the parent exhaust on its own no-progress timer, so a
+stuck sub-agent can never pin a parent's recovery open forever.
+
+This completes the sub-agent recovery story started by the stable-runId +
+bounded re-attach fix (#1630): the child self-heals and the parent both
+re-attaches to it _and_ keeps its own recovery alive while doing so.
--- a/examples/deploy-churn/src/server.ts
+++ b/examples/deploy-churn/src/server.ts
@@ -958,8 +958,11 @@ export class DeployChurnAgent extends Think<Env> {
       );
     }
     // Wipe submissions, the tool ledger, and tool config so a fresh run starts
-    // from zero. (Each orchestrator run also uses a fresh session/DO instance,
-    // so the durable transcript starts empty too.)
+    // from zero. Also clear the durable chat transcript: a REUSED session would
+    // otherwise keep the prior turn's tool result, so the sub-agent mock model
+    // (which only emits `runTask` when no tool result is present) would short-
+    // circuit to "DONE" and never dispatch the child.
+    await this.clearMessages();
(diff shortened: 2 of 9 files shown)


threepointone:
/bonk do a deep review on this, look for edge cases, is the ux/dx good

PR #1643, opened 2026-06-01
fix(think): route a stream-stall watchdog abort into bounded recovery (#1626)

### LINKED ISSUE #1626: discussion: route a stream-stall watchdog abort into bounded recovery instead of a terminal error
## Summary

The inactivity watchdog added in #1623 (`Think.chatStreamStallTimeoutMs`) currently turns a hung model stream into a **terminal** error. That fixes the "infinite spinner" — but for the common cause (a transient provider/transport hang, often during the same deploy churn the rest of #1623 hardens against), a bounded **retry/continue** would recover the turn instead of failing it. This issue is to discuss whether and how to route a watchdog stall into the existing recovery path.

Filing as a discussion because there are real trade-offs (idempotency, distinguishing a stall from a genuine model error, opt-in vs default).

Related: #1623 (watchdog + recovery hardening).

## Current behavior

When no UI-message-stream chunk arrives within `chatStreamStallTimeoutMs`, the watchdog (`_iterateWithStallWatchdog`):

1. emits a `chat:stream:stalled` observability event,
2. calls `abortRequest(requestId)` to tear down the upstream model stream,
3. throws, so the read loop (`_streamResult` / `_streamResultToRpcCallback`) exits its `catch`, which:
   - marks the resumable stream errored and broadcasts a terminal `done + error` frame,
   - calls `onChatError(error, { stage: "stream", messagesPersisted: true })`,
   - fires `onChatResponse` with `status: "error"`.

So the turn ends terminally. The user must re-send.

## Why a stall is *not* currently recovered

Think's chat recovery (`_beginChatRecoveryIncident` → `_chatRecoveryContinue` / `_chatRecoveryRetry`) is triggered by **Durable Object interruption**: an interrupted fiber snapshot is detected on restart (deploy / eviction / process restart). A stall is different — it happens **inside a live isolate** that never restarts, so no fiber-recovery scan runs and the bounded-recovery machinery is never entered.

## Proposal (for discussion)

Treat a watchdog stall like a recoverable interruption: instead of failing terminally, schedule a continuation through the existing bounded-recovery path so a transient hang is retried within budget.

The pieces to make this safe mostly already exist from #1623:

- **Settled tool durability** — `tool-output-available` / `tool-output-error` chunks are flushed immediately (#1621), so a continuation re-anchors without re-running completed (often non-idempotent) tool calls.
- **Transcript repair** — an interrupted/abandoned tool call is preserved as an errored result, so the continuation transcript is provider-valid.
- **Trailing-assistant checkpoint** — `ensureValidContinueCheckpoint` already lets a continuation resume after a partial assistant turn without a prefill 400.
- **Bounded budget** — `maxAttempts` + the 15-min wall-clock window + progress-aware reset already cap retries, so a persistently-hanging provider can't loop forever; a stall-retry should consume the same incident budget.

Sketch: on stall, persist the partial assistant message (already happens in the error path), open/advance a recovery incident with a `recoveryKind` like `"stall"` (or reuse `"continue"`), and schedule `_chatRecoveryContinue` rather than emitting the terminal error — falling back to terminal only once the budget is exhausted (mirroring how `stable_timeout` now reschedules).

## Open questions

- **Opt-in or default?** The watchdog itself is opt-in (`chatStreamStallTimeoutMs = 0`). Should stall-recovery be automatic once the watchdog is enabled, or a separate flag (e.g. `chatStreamStallTimeoutMs: { ms, recover: true }`)?
- **Stall vs genuine model error.** A provider that returns a hard error mid-stream is different from a silent hang. Today both hit the same terminal path. Should only the *watchdog-triggered* abort be retried, while in-band errors stay terminal? (The watchdog already has a distinct signal — `chat:stream:stalled` — so they're distinguishable.)
- **Idempotency of the in-flight step.** Settled tool calls are durable, but a tool that was *mid-execution* when the stall fired is not. Re-anchoring would re-run it. Is that acceptable (it's the same guarantee as a deploy-interrupted turn), or should stall-recovery only continue when no tool call is in flight?
- **Interaction with the abort.** The watchdog calls `abortRequest`, which aborts the turn's signal. A continuation would need a fresh request context — same as the existing continuation path, but worth confirming the aborted signal doesn't leak into the retry.
- **AIChatAgent.** The watchdog is Think-only (AIChatAgent delegates the read loop to `onChatMessage`). Stall-recovery would be Think-only too. Acceptable?

## Scope / non-goals

- Not changing the default (`chatStreamStallTimeoutMs` stays `0`/off).
- Not adding a new recovery budget — reuse the existing incident budget and wall-clock ceiling.
- Not auto-retrying genuine, fast model errors (those are usually deterministic and would just re-fail).

## References

- `packages/think/src/think.ts` — `_iterateWithStallWatchdog`, `_streamResult` / `_streamResultToRpcCallback`, `_beginChatRecoveryIncident`, `_chatRecoveryContinue`, `_rescheduleRecoveryAfterStableTimeout`, `ensureValidContinueCheckpoint`
- `chatStreamStallTimeoutMs` config and `chat:stream:stalled` observability event (#1623)
- Settled-tool durability (#1621) and transcript-repair behavior that make a continuation safe

threepointone:
## Summary

Closes #1626. When `chatStreamStallTimeoutMs` is set and the inactivity watchdog fires on a **hung model/transport stream** (a provider that parks without ever throwing), the turn was failed **terminally**. A stall is really just another interruption — like a deploy or eviction — so it is now routed into the **same bounded chat-recovery path**: the settled partial is preserved, a continuation is scheduled, and the turn resumes. A transient hang (the common cause under deploy churn) recovers automatically; a persistently hanging provider still terminalizes once the recovery budget is exhausted — the watchdog's original "kill the infinite spinner" guarantee, now after bounded retries.

This removes the last framework gap behind a customer's stall-watchdog→terminal hack (g3's `classifyWatchdogTick` + manual stall→terminal path).

## Design decisions

- **Automatic when the watchdog is enabled + `chatRecovery` is on (the Think default) — no new config.** A `{ ms, recover }` flag would be exactly the new-API knob we avoid; a stall routing through the existing recovery (like every other interruption) is the consistent default. The "fail fast" knob is the existing `maxAttempts`/budget. With `chatRecovery` **off**, a stall stays terminal as before.
- **Reuses the existing incident budget** (`maxAttempts` + 15-min wall-clock + no-progress window + progress-aware reset). No new budget. Boundedness is inherited from `_beginChatRecoveryIncident`.
- **Only the watchdog abort routes to recovery** — in-band model errors stay terminal (distinguished by a dedicated `ChatStreamStalledError`).
- **Idempotency matches deploy recovery:** settled tool results are durable (not re-run); a tool mid-execution when the stall fired re-runs on the continuation.
- **Think-only** (the watchdog is Think-only; `AIChatAgent` delegates the read loop).

## Implementation

- `_iterateWithStallWatchdog` throws a distinct **`ChatStreamStalledError`** so the read-loop catch can tell a stall from an in-band error.
- New **`_routeStallToBoundedRecovery`** — gated on chatRecovery enabled — opens/advances a recovery incident and schedules `_chatRecoveryContinue` (with `recoveredRequestId` when a submission is running), or returns `false` when the budget is exhausted so the caller falls through to the existing terminal path.
- Both stream paths (`_streamResult`, `_streamResultToRpcCallback`) branch their catch: persist the settled partial (re-anchor), route to recovery, and suppress the terminal error frame when a continuation is scheduled (the continuation owns the real terminal outcome).

## Tests

- **Unit** (`think-session.test.ts`): a stall on the first inference recovers via the scheduled continuation (no terminal error, the continuation streams to completion). The existing watchdog→terminal test now runs with `chatRecovery` off — the no-recovery gate. (Full think suite green: **469**.)
- **E2E** (`stall-recovery.test.ts`, real `wrangler dev` — **no process kill**, since a stall is an in-isolate hang): the model stalls the first inference then streams; the watchdog fires, the framework's **real alarm** fires the scheduled continuation, and the turn completes with the continuation's output and **no orphaned fiber rows**.

## Verification

- `npm run check` clean (91 projects); think unit suite **469**; new stall-recovery e2e green.

## Test plan

- [x] Unit: stall → recovery (recover) + watchdog → terminal (recovery off)
- [x] E2E: real-alarm stall→recovery, no orphaned fibers
- [x] `npm run check` clean


threepointone:
/bonk do a deep review on this, look for edge cases, is the ux/dx good and tuneable

Issue #1645, opened 2026-06-01
ai-chat: chat-recovery exhaustion has no terminal-status hydration (disconnected client misses the terminal banner on reconnect)


threepointone:
## Summary

In `@cloudflare/ai-chat`, when chat recovery is **exhausted**, `_exhaustChatRecovery` delivers the terminal UX **only** as a transient broadcast (`_broadcastChatMessage` with `done: true, error: true`). It does **not** durably record the terminal status. A client that is disconnected at the moment recovery gives up (e.g. mid deploy-churn / WS reconnect storm — the exact condition under which recovery exhausts) will **never** see the terminal banner on reconnect, and the turn appears frozen.

`@cloudflare/think` does not have this gap: its `_exhaustChatRecovery` additionally calls `_recordTerminalChatStatus("interrupted", …)`, which persists `CHAT_LAST_TERMINAL_KEY`, and `_buildIdleConnectMessages` replays that terminal error to a client on (re)connect. ai-chat has no equivalent.

## Scope

This affects **all** ai-chat recovery exhaustion paths (deploy/eviction exhaustion via `_handleInternalFiberRecovery`, and the stable-state-timeout give-up), not any single path. It is pre-existing.

## Where

- `packages/ai-chat/src/index.ts` — `_exhaustChatRecovery` only broadcasts; no durable terminal record.
- Compare `packages/think/src/think.ts` — `_exhaustChatRecovery` → `_recordTerminalChatStatus("interrupted", …)`, plus `_buildIdleConnectMessages` / `_pendingTerminalChatResponse` for replay on idle connect.

## Repro sketch

1. Start a turn on an ai-chat agent with `chatRecovery` enabled and a custom `terminalMessage`.
2. Drive recovery to exhaustion while the client socket is closed (e.g. simulate deploy churn so the budget drains while disconnected).
3. Reconnect the client.
4. **Expected:** the client receives a terminal error message (the configured `terminalMessage`) on connect.
5. **Actual:** no terminal frame is replayed; the UI shows no resolution for the abandoned turn.

## Proposed fix

Give ai-chat a durable last-terminal record analogous to think's: persist the terminal message/requestId in `_exhaustChatRecovery` and replay it in the agent's on-connect message build, so an exhausted turn surfaces on reconnect rather than freezing.

## Context

[...]

PR #1646, opened 2026-06-01
fix(think,ai-chat): terminalize via onExhausted when recovery gives up waiting for stable state

threepointone:
## Summary

Under extreme deploy churn (a long multi-tool turn interrupted many times in quick succession), a chat-recovery callback (`_chatRecoveryRetry` / `_chatRecoveryContinue`) could keep timing out waiting for the isolate to reach **stable state** until its retry budget drained. The give-up path only marked the incident `failed` and completed the recovered submission as `error` — it **bypassed `_exhaustChatRecovery` entirely**. So:

- `onExhausted` never fired,
- `chat:recovery:exhausted` was never emitted,
- the configured `terminalMessage` banner was never delivered,
- the terminal chat status was never recorded.

An app that trusts the framework's progress-keyed budget and relies on `onExhausted` for the terminal banner regressed to an **eternal spinner** with no terminal signal. In the worst case (incident record missing/swept before a stale alarm fired) the turn was dropped with **no terminalization at all**.

This is the give-up path in the recovery callbacks — distinct from, and **not** covered by, #1626, which fixed the sibling stall-watchdog exhaustion path to route through `_exhaustChatRecovery`. This PR brings the stable-state-timeout give-up in line with every other "recovery gave up" path.

Reported against the `@1636` preview pin (the give-up was unchanged at that commit and at the #1626 head).

## What changed

- **New shared helper `_exhaustRecoveryAfterStableTimeout(callback, data)`** in both `@cloudflare/think` and `@cloudflare/ai-chat`. The budget-spent stable-state give-up now routes through the **same** `_exhaustChatRecovery` path as deploy-recovery and stall exhaustion: fires `onExhausted` (with `reason: \"stable_timeout\"`), emits `chat:recovery:exhausted`, marks the durable submission interrupted (think), records the terminal chat status (think), and delivers the configured `terminalMessage`.
- **Silent-drop backstop:** when the incident record is missing (no `incidentId`, or it was swept/deleted before a stale alarm fired), the give-up **synthesizes** a terminal incident from the recovery-root request id so the turn STILL terminalizes through `onExhausted` instead of being dropped with no terminal UX.
- **Re-entry guard:** once an incident is sealed `exhausted`, a duplicate stale alarm (or retried callback) returns early so `onExhausted` / the terminal banner cannot fire twice. The durable-submission paths already short-circuit via the submission-not-running check; this guard additionally covers callers with no submission layer (notably `@cloudflare/ai-chat`).
- **`_resolveRecoveryStreamId` helper** surfaces the turn's partial into the `onExhausted` context best-effort (read-only; no orphan re-persist — matching the prior give-up behavior, since `waitUntilStable` failed before any turn ran so there is no new settled work).
- Both give-up branches in both callbacks, in both packages, now call the helper instead of `_updateChatRecoveryIncident(\"failed\")` + `_completeRecoveredSubmission(\"error\")`.
- **Docs:** `ChatRecoveryExhaustedContext.reason` JSDoc now enumerates all four emittable reasons (`max_attempts_exceeded`, `no_progress_timeout`, `max_recovery_window_exceeded`, `stable_timeout`) and notes it is an open string.

## Behavior notes (observable changes)

- The stable-timeout give-up now emits **`chat:recovery:exhausted`** instead of `chat:recovery:failed`. This is the correct semantics (it *is* exhaustion) and aligns with deploy/stall exhaustion — exactly what an `onExhausted`-based app expects.
- The recovered submission's error message is now the user-facing `terminalMessage` (via `_markRecoveredSubmissionInterrupted`) rather than an internal diagnostic string; status is still `error`.
- The incident is retained as `exhausted` (was `failed`); same TTL-sweep retention, no storage leak.

## Deliberate scope choice

The give-up terminalizes unconditionally (it is reached only after `waitUntilStable` fails, before the leaf/`targetUserId` checks). The existing `_exhaustChatRecovery` callers (fiber-recovery + stall exhaustion) also terminalize unconditionally, so adding a superseded-turn guard *only here* would make exhaustion semantics inconsistent across paths. The progress-keyed budget makes a superseded-during-give-up scenario contrived; left as a possible future unification across all exhaustion paths.

## Tests

- **think** (`think-session.test.ts`): rewrote the old "fails terminally" test to assert `exhausted` + `onExhausted` + persisted terminal banner; added retry-path coverage, a missing-incident silent-drop guard, a no-`incidentId` case, and a no-submission duplicate-alarm re-entry guard. **159 pass.**
- **ai-chat** (`durable-chat-recovery.test.ts`): mirrored continue + retry exhaustion, the silent-drop guard, and the duplicate-alarm re-entry guard. **33 pass** (489 across the full workers suite).
- Extended harnesses: `enableExhaustedCaptureForTest` gains an optional terminal-message arg; added `runChatRecoveryRetryDirectForTest` (ai-chat).
- No e2e added: forcing repeated real stable-state timeouts to drain the budget over real alarms is slow/flaky, and the exhausted branch already flows through the same `_exhaustChatRecovery` the existing stall-recovery e2e and deploy-churn harness exercise end-to-end. Covered deterministically at the seam.

## Changeset

`@cloudflare/think` + `@cloudflare/ai-chat` patch.

## Follow-up

- ai-chat has weaker terminal hydration than think (its `_exhaustChatRecovery` only broadcasts transiently; no durable last-terminal record replayed on reconnect). Pre-existing and affects **all** ai-chat exhaustion paths — tracked in #1645, out of scope here.

## Test plan

- [x] `npm run check` (all 91 projects typecheck; format + lint clean)
- [x] `packages/think` `think-session.test.ts` — 159 pass
- [x] `packages/ai-chat` full workers suite — 489 pass
- [ ] CI green

### DIFF
--- a/.changeset/recovery-stable-timeout-exhaustion.md
+++ b/.changeset/recovery-stable-timeout-exhaustion.md
@@ -1,0 +1,26 @@
+---
+"@cloudflare/think": patch
+"@cloudflare/ai-chat": patch
+---
+
+Terminalize a chat-recovery turn through `onExhausted` when it gives up waiting for stable state
+
+Under extreme churn (a long turn interrupted many times in quick succession), a
+recovery callback (`_chatRecoveryRetry` / `_chatRecoveryContinue`) could keep
+timing out waiting for the isolate to reach stable state until its retry budget
+drained. The give-up path only marked the incident `failed` and completed the
+recovered submission as `error` — it **bypassed `_exhaustChatRecovery`**, so
+`onExhausted` never fired, the `chat:recovery:exhausted` event was not emitted,
+the configured `terminalMessage` banner was never delivered, and the terminal
+chat status was not recorded. Apps relying on `onExhausted` for the terminal
+banner saw an eternal spinner with no terminal signal.
+
+The stable-state-timeout give-up now routes through the **same**
+`_exhaustChatRecovery` path as deploy-recovery and stall exhaustion: it fires
+`onExhausted` (with `reason: "stable_timeout"`), emits `chat:recovery:exhausted`,
+marks the durable submission interrupted, records the terminal chat status, and
+delivers the `terminalMessage`. As an extra backstop against silent drops, the
+give-up also terminalizes when the incident record is missing (no `incidentId`,
+or it was swept/deleted before a stale alarm fired) by synthesizing a terminal
+incident from the recovery-root request id — so a turn can never be dropped with
+no terminal UX.
--- a/packages/agents/src/chat/lifecycle.ts
+++ b/packages/agents/src/chat/lifecycle.ts
@@ -157,7 +157,16 @@ export type ChatRecoveryExhaustedContext = Pick<
   | "partialText"
   | "partialParts"
 > & {
-  /** Why recovery stopped: `max_attempts_exceeded` or `max_recovery_window_exceeded`. */
+  /**
+   * Why recovery stopped. One of:
+   * - `max_attempts_exceeded` — the per-incident attempt budget was spent.
+   * - `no_progress_timeout` — no forward progress within the no-progress window.
+   * - `max_recovery_window_exceeded` — the absolute incident-age ceiling was hit.
+   * - `stable_timeout` — a recovery attempt kept timing out waiting for the
+   *   isolate to reach stable state until the budget drained (extreme churn).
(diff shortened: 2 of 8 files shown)


threepointone:
/bonk review deeply, look for edge cases, and whether ux/dx is good

Issue #1650, opened 2026-06-02
Auto-continuation parallel-tool barrier: make Think event-driven (avoid timeout fire-through for human-in-the-loop tools + orphan keepAlive) — follow-up to #1649


threepointone:
## Context

#1649 fixed auto-continuation firing before all of a step's parallel client-tool results arrive. The fix adds a **barrier** in both `@cloudflare/think` and `@cloudflare/ai-chat`: when the model emits parallel tool calls and the client answers them independently (each `addToolOutput` sends `cf_agent_tool_result` with `autoContinue: true`), the server now holds the continuation until the step's batch is fully answered instead of firing on the first result.

The barrier is detected via `_hasIncompleteToolBatch()` — the **leaf** assistant message has both a settled tool result *and* a still-`input-available`/`approval-requested` sibling — and the wait is **bounded** by `AUTO_CONTINUATION_PENDING_TOOL_TIMEOUT_MS` (currently 60s). On timeout it fires through (Think repairs the unanswered call to errored; ai-chat surfaces the provider error), and logs a `console.warn`.

Key code (post-#1649):
- Think: `_fireAutoContinuationWhenStable`, `_awaitToolBatch`, `_hasIncompleteToolBatch` in `packages/think/src/think.ts`. The barrier runs **before** enqueuing the continuation turn (so it does not occupy the turn queue), wrapped in `keepAliveWhile`.
- ai-chat: `_awaitPendingInteractionBarrier`, `_hasIncompleteToolBatch` in `packages/ai-chat/src/index.ts`. The barrier runs **inside** the queued continuation turn (after `_awaitPendingAutoContinuationPrerequisite`).

## Problem with the bounded-wait approach

The fixed 60s timeout + fire-through is correct for the common case (parallel data-fetch tools resolving at different sub-second/second latencies) but has two rough edges:

1. **Human-in-the-loop tools emitted in parallel.** A client-resolved tool with no `execute` (e.g. an `ask_user`/`display_ui`-style prompt) parks at `input-available` until a human answers — potentially minutes. If the model emits such a tool **in parallel** with another tool that resolves quickly, the barrier waits 60s and then **fires through, repairing the still-open human tool to errored** — even though the user is legitimately still answering. (`ask_user` is usually emitted alone, so this is an edge, but `display_ui`-heavy apps like the g3 customer make it plausible.)

[...]

Issue #1642, opened 2026-06-01
refactor: unify duplicated chat-recovery/repair machinery into the shared agents/chat layer (N3)


threepointone:
## Problem

The chat-recovery + transcript-repair machinery is **duplicated, nearly verbatim**, between `packages/think/src/think.ts` and `packages/ai-chat/src/index.ts`. Every recovery fix has to be written **twice**, the two copies drift, and the drift is a real source of bugs and confusion.

This was painful and concrete across the recent recovery stack (#1633, #1634, #1635, #1636, #1638, #1640, #1641): each touched think **and** ai-chat in lockstep, and the divergence caused issues — e.g. ai-chat's persist gate is `streamStillActive`-only while think uses `_shouldPersistOrphanedPartial` (an intentional asymmetry, but one that has to be re-reasoned every time), and ai-chat has **no transcript-repair pass at all** while think does.

**The drift is now actively growing.** #1650 made Think's parallel-tool auto-continuation barrier **event-driven** (#1667, #1671), while ai-chat deliberately stayed on the **bounded-wait** barrier from #1649/#1651 — so the two now use *different mechanisms* for the same job. This is the latest concrete example of the cost of duplication and is explicitly deferred here (see the new asymmetry below).

## Goal

Hoist the shared recovery/repair machinery into `packages/agents/src/chat/` so it is written once and both `Think` and `AIChatAgent` consume it. **Pure refactor — no behavior change.** Design home: `design/chat-shared-layer.md`.

## Duplicated surface to unify (audit, not exhaustive)

- **Budget / incident:** `_beginChatRecoveryIncident`, `_handleInternalFiberRecovery`, `_exhaustChatRecovery`, `_sweepStaleChatRecoveryIncidents`, the `ChatRecoveryIncident` shape, and the constants (`CHAT_RECOVERY_NO_PROGRESS_WINDOW_MS`, `CHAT_RECOVERY_ALARM_DEBOUNCE_MS`, `CHAT_RECOVERY_MAX_WINDOW_MS`, `DEFAULT_CHAT_RECOVERY_MAX_ATTEMPTS`, `DEFAULT_CHAT_RECOVERY_STABLE_TIMEOUT_MS`, `CHAT_RECOVERY_PROGRESS_KEY`, `AGENT_TOOL_STREAM_PROGRESS_BUMP_THROTTLE_MS`).
[...]

Issue #1160, opened 2026-03-23
ai-chat: tool continuations after client-side tools lose earlier steps' parts


alexander-zuev:
## Summary

When an agent uses **tools that require user input** — client-side tools without `execute` (e.g. `ask_choice`, `collect_credentials`) or tools with `needsApproval` (e.g. `edit_workflow`, `deploy_workflow`) — the assistant message loses parts from earlier steps after each continuation. Only the most recent continuation's parts survive in the broadcast.

The key pattern: any tool that **pauses execution and waits for user interaction** before the SDK continues. This includes both client-side tools and server-side tools with approval gates.

This does NOT happen with pure server-side tool chains that execute without user interaction (e.g. `read_workflow` → `validate_workflow` → `get_mcp_status` in one turn).

## Reproduction scenario

1. User sends a message
2. Agent calls tool A (`ask_choice`, no `execute`) — step finishes, SDK pauses for user input
3. User provides choice → tool continuation starts
4. Agent calls tool B (`collect_credentials`, no `execute`) — step finishes, SDK pauses for user input
5. User provides credentials → tool continuation starts
6. Agent calls tool C (`read_workflow`) + tool D (`edit_workflow` with `needsApproval`) — step finishes, SDK pauses for approval
7. User approves → tool continuation starts
8. **At this point, the `messages` argument passed to `persistMessages` only contains parts from the latest continuation — parts from steps A and B are gone**

## Evidence

We logged both `this.messages` and the `messages` argument inside `persistMessages` at call time:

```
persist_debug {
  inMemoryAssistant: [{ id: "assistant_177428", parts: 14 }]   ← this.messages (full)
  incomingAssistant: [{ id: "assistant_177428", parts: 6 }]    ← messages arg (truncated)
}
```

**`this.messages` has 14 parts (the complete accumulated message). The `messages` argument passed to `persistMessages` by the SDK has only 6 parts (only the latest continuation step).** The SDK is calling `persistMessages` with a truncated copy.

After `super.persistMessages` writes the truncated 6 parts to SQLite and reloads, `this.messages` gets overwritten with the 6-part version. The full 14-part in-memory state is lost.

We verified this is not caused by our `persistMessages` override — the issue reproduces with the override reduced to a near no-op (metadata stamp + `super.persistMessages` passthrough with no filtering).

## Root cause

[...]

PR #1651, opened 2026-06-02
fix(think,ai-chat): wait for all parallel client-tool results before auto-continuing (#1649)

### LINKED ISSUE #1649: _scheduleAutoContinuation 50ms timer fires before all parallel client-tool results arrive → MissingToolResultsError
When the AI returns multiple parallel tool calls and some are resolved by the client via onToolCall (client-resolved tools), the server's auto-continuation fires before all tool results arrive, causing MissingToolResultsError.

The root cause: each addToolOutput() call sends cf_agent_tool_result with autoContinue: true to the server independently. The server's _scheduleAutoContinuation in @cloudflare/think uses a 50ms coalesce timer (think). If the first tool resolves quickly (e.g., a synchronous client-side state mutation) but another tool takes longer (e.g., calls a backend RPC via agent.call()), the 50ms window expires and the server fires _runInferenceLoop with incomplete tool results.

I thought maybe there's something wrong with the implementation so I created a new project from the starter template and the bug is still there. (Although Claude had to make some patches to make the project work, idk what those are, i'm exhausted)

https://github.com/abhagsain/cf-think-bug

Image

Image

To Reproduce

1. Create a Think agent with multiple client-resolved tools (no execute function — resolved via onToolCall on the client)
2. Have at least one tool that resolves synchronously (e.g., reads from local state) and one that resolves asynchronously (e.g., calls a backend @callable() that takes >50ms)
3. Prompt the AI to use both tools in a single step (e.g., "delete item A and modify item B" — the AI emits parallel tool calls)
4. The fast tool's result triggers _scheduleAutoContinuation with its 50ms timer
5. The slow tool hasn't resolved yet when the timer fires
6. Server throws: MissingToolResultsError: Tool result is missing for tool call <id>

Note: We're using `onToolCall` witth `useAgentChat`, not the exectute with client-tools

Minimal reproduction:

// Server: Think agent with client-resolved tools
```ts
class MyAgent extends Think<Env, State> {
  getConfig() {
    return {
      tools: {
        fastTool: tool({ description: "Fast", inputSchema: z.object({}) }),
        // no execute — client-resolved
        slowTool: tool({ description: "Slow", inputSchema: z.object({}) }),
        // no execute — client-resolved
      }
    };
  }
}

// Client:
useAgentChat({
  agent,
  onToolCall: async ({ toolCall, addToolOutput }) => {
    if (toolCall.toolName === "fastTool") {
      // Resolves immediately
      addToolOutput({ toolCallId: toolCall.toolCallId, output: { ok: true } });
    }
    if (toolCall.toolName === "slowTool") {
      // Takes >50ms (e.g., backend RPC)
      const result = await agent.call("someCallable", [toolCall.input]);
      addToolOutput({ toolCallId: toolCall.toolCallId, output: result });
    }
  },
});
```
Prompt the AI to call both fastTool and slowTool in a single step. The server auto-continues after fastTool's result + 50ms, before slowTool resolves.

### Expected behavior

The server should wait for ALL tool results in a step before auto-continuing, not just 50ms after the last received result. The server knows how many tool calls it emitted — it
should use that count as a barrier.

Screenshots

Server error:
MissingToolResultsError [AI_MissingToolResultsError]: Tool result is missing for tool call <id>

Client error (secondary — activeResponse is undefined because the continuation request failed):
TypeError: Cannot read properties of undefined (reading 'state') https://github.com/vercel/ai/issues/15668 at Chat.makeRequest (chat.ts:761)

Version

- @cloudflare/think: 0.6.1
- @cloudflare/ai-chat: 0.7.2
- agents: 0.13.3
- ai: 6.0.194

Additional context

- The 50ms coalesce timer is at think.js:2581 in _scheduleAutoContinuation
- Each tool result resets the timer, but async tools that take seconds will always lose this race
- Setting autoContinueAfterToolResult: false is a workaround but requires manually managing continuation
- The client SDK (@cloudflare/ai-chat) sends autoContinue: true per-result — it has no concept of "this is the last result in the batch"
- The activeResponse crash is a separate issue in ai SDK (filed as vercel/ai#15668), triggered as a secondary effect of this bug

Claude's Suggested fix

Either:
1. Server-side barrier: The server knows how many tool calls the assistant emitted. Wait for that many results (with a timeout) before auto-continuing, instead of using a fixed 50ms coalesce window.
2. Client-side batching: The client SDK should only send autoContinue: true on the last tool result in a step, not on every result.

> I got the issue where parallel tools were breaking on the client for vercel's sdk and I couldn't understand what was wrong and I asked claude to investigate.

threepointone:
Fixes #1649.

## Problem

When the model emits **multiple tool calls in a single step** and the client resolves them independently (each `addToolOutput` sends `cf_agent_tool_result` with `autoContinue: true`), a **fast** tool's result could trigger the next inference while a **slower** sibling was still `input-available`.

The server's auto-continuation fired on a short coalesce timer (think `50ms`, ai-chat `10ms`), not on "is the model's tool-call batch actually complete." So the first result's `autoContinue` would continue the turn with an incomplete tool-result set:

- On a build without transcript repair → `AI_MissingToolResultsError` (the crash the reporter saw).
- On current `main` (Think repairs + `convertToModelMessages(..., { ignoreIncompleteToolCalls: true })`) → arguably worse: the still-in-flight sibling is silently flipped to errored and a **spurious extra continuation** runs, then the real result lands and churns the transcript.

Repro shape (from the issue):

```ts
onToolCall: async ({ toolCall, addToolOutput }) => {
  if (toolCall.toolName === "fastTool") {
    addToolOutput({ toolCallId: toolCall.toolCallId, output: { ok: true } }); // sync
  }
  if (toolCall.toolName === "slowTool") {
    const result = await agent.call("someCallable", [toolCall.input]); // >50ms
    addToolOutput({ toolCallId: toolCall.toolCallId, output: result });
  }
}
```

## Fix

Add a **server-side barrier**: auto-continuation now holds until the model's parallel tool batch is fully answered, then runs a single continuation — regardless of result arrival order. This uses the transcript state the framework already tracks, so it covers tool results *and* approvals and needs no counting.

### Scoped to the leaf (the key design choice)

The barrier fires only when the **latest assistant message** is *mid-batch* — it carries **both** a settled tool result **and** a still-pending (`input-available`/`approval-requested`) sibling (`_hasIncompleteToolBatch()`). That is precisely the #1649 signature (the model fanned out parallel calls and only some are answered).

It is intentionally **not** "any pending interaction anywhere in the transcript." A broad barrier over-blocked a legitimate, pre-existing path — a follow-up continuation when a result lands in a *separate* assistant message before a stream starts (`chat-turn-serialization` test). Leaf-scoping fixes #1649 without disturbing that.

### Mechanics

- **think** (`packages/think/src/think.ts`): `_scheduleAutoContinuation`'s timer now calls `_fireAutoContinuationWhenStable`, which gates `_fireAutoContinuation` behind `_awaitToolBatch`. The wait runs **before** the turn is enqueued, so it never occupies the turn queue, and is wrapped in `keepAliveWhile` so the (typically sub-second) remainder of the batch can land. It awaits the in-flight tool-result apply (`_pendingInteractionPromise`) so a sibling that lands mid-wait is observed promptly.
- **ai-chat** (`packages/ai-chat/src/index.ts`): `_awaitPendingInteractionBarrier` runs **inside** the queued continuation (after `_awaitPendingAutoContinuationPrerequisite`), because that path runs in-queue and so can't reuse the queue-idle `waitUntilStable` (it would deadlock). Same `_hasIncompleteToolBatch` predicate.
- **Bounded** by `AUTO_CONTINUATION_PENDING_TOOL_TIMEOUT_MS` (60s): on timeout the continuation proceeds — Think's repair backstop errors the unanswered call, ai-chat surfaces the provider error — so a genuinely orphaned tool (client disconnected mid-batch) never pins the continuation open. A `console.warn` makes that fall-through visible.

### Hardening found during review

- **Double-fire race:** during a barrier wait, each sibling result re-arms the `50ms` coalesce timer; when the barrier fired the continuation, that leftover timer could enqueue a **second** continuation in the window before `pastCoalesce` was set. `_fireAutoContinuation` now cancels any still-armed timer when it fires, and `_fireAutoContinuationWhenStable` bails if the continuation is already `pastCoalesce`. (ai-chat is structurally immune — a sibling during a pending continuation hits the merge branch, not a re-queue.)
- **Stale barrier on clear:** `resetTurnState` now clears the barrier flag, and both wait loops bail when `_continuation.pending` is gone, so a chat clear / turn reset can't strand the wait holding the isolate.

## Why server-side (not the client-side option from the issue)

The issue floated "client only sends `autoContinue` on the last result." That's fragile (tools resolve independently in `onToolCall`; the client isn't the transcript's source of truth) and wouldn't protect custom / non-`useAgentChat` clients. The server already owns the transcript and the completeness invariant, so the barrier belongs there.

## Tests

- **New** parallel fast/slow coverage in both packages: assert exactly **one** continuation fires after **both** results land, and the slow tool is never flipped to errored by a premature repair.
- **Updated** the think `deferred continuation` test and the ai-chat continuation-stream-error test — both previously encoded the old partial-continuation behavior (continue after the first of two parallel tools).
- Full suites green: **think 483**, **ai-chat 492**. Typecheck, oxlint, oxfmt clean.

## Follow-up

#1650 tracks making the Think barrier **event-driven** (return-and-wait-for-the-next-result-event instead of a bounded timer) to avoid the 60s fire-through for human-in-the-loop tools (e.g. `ask_user`/`display_ui` emitted in parallel) and the orphan keepAlive hold. Kept symmetric (bounded-wait) with ai-chat here on purpose, since #1642 (N3) wants to unify the duplicated think↔ai-chat layer.

## Test plan

- [ ] CI green (`npm run build` + `npm run check` + affected tests)
- [ ] Manual: parallel client tools where one resolves synchronously and one via a >50ms backend RPC → single clean continuation, no `MissingToolResultsError`, no spurious repair
- [ ] Sanity: a single client tool still auto-continues immediately (no added latency)
- [ ] Sanity: server-approval continuation (#1627) and chained approvals (#1160) still work

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1651" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/fix-parallel-client-tool-continuation.md
+++ b/.changeset/fix-parallel-client-tool-continuation.md
@@ -1,0 +1,19 @@
+---
+"@cloudflare/think": patch
+"@cloudflare/ai-chat": patch
+---
+
+Fix auto-continuation firing before all parallel client-tool results arrive
+(#1649). When the model emitted multiple tool calls in one step and the client
+resolved them independently via `addToolOutput`, a fast result's `autoContinue`
+could trigger the next inference while a slower sibling was still
+`input-available`. That fed the provider an incomplete tool-result set
+(`MissingToolResultsError`) or — via the transcript-repair backstop — silently
+flipped the in-flight sibling to errored and ran a spurious extra continuation.
+
+Auto-continuation now waits until the transcript is stable (no
+`input-available`/`approval-requested` parts) before continuing, so a fanned-out
+tool batch coalesces into a single continuation regardless of result arrival
+order. The wait is bounded, so a genuinely orphaned tool call (e.g. the client
+disconnected mid-batch) still falls through to the existing backstop instead of
+pinning the continuation open.
--- a/packages/ai-chat/src/index.ts
+++ b/packages/ai-chat/src/index.ts
@@ -148,6 +148,15 @@ const CHAT_RECOVERY_PROGRESS_KEY = "cf:chat-recovery:progress";
 // first under normal deploy cadence (#1637).
 const DEFAULT_CHAT_RECOVERY_MAX_ATTEMPTS = 10;
 const DEFAULT_CHAT_RECOVERY_STABLE_TIMEOUT_MS = 10_000;
+// Auto-continuation barrier (#1649): when the model emits parallel tool calls,
+// the client answers each one independently and sends a tool result with
+// `autoContinue` per result. A fast tool's result must NOT trigger inference
+// while a slower sibling is still `input-available` — that feeds the provider
+// an incomplete tool-result set. So we wait until the transcript is stable (no
+// `input-available`/`approval-requested` parts) before continuing, bounded by
+// this timeout so a genuinely orphaned tool call (e.g. the client disconnected
+// mid-batch) still falls through instead of pinning the continuation open.
+const AUTO_CONTINUATION_PENDING_TOOL_TIMEOUT_MS = 60_000;
 // Delay before retrying a recovery that timed out waiting for stable state.
 // Gives an actively-churning isolate (e.g. a deploy in flight) time to settle.
 const CHAT_RECOVERY_STABLE_RETRY_DELAY_SECONDS = 3;
@@ -2090,7 +2099,7 @@ export class AIChatAgent<
     while (true) {
       const prerequisite = this._continuation.pending?.prerequisite;
       if (!prerequisite) {
-        return true;
+        break;
       }
 
       const applied = await prerequisite;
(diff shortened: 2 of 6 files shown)


threepointone:
/bonk do a deep review here, look for edge cases and things we might've missed

PR #1663, opened 2026-06-03
fix(think): hold continuation barrier for a slow tool sibling pending only in the streaming accumulator (#1649 follow-up)

threepointone:
## Summary

Follow-up to #1657. A customer reported that #1657 fixed the **fast** parallel tool case but they still hit `"The tool call was interrupted before a result was recorded."` on a **slow** client-resolved sibling (an RPC taking 2–5s) running beside a fast one (e.g. _"delete the 4th slide and rewrite the 1st"_ → `deleteSlides` fast + `modifySlide` slow).

Their root-cause analysis was correct: `_hasIncompleteToolBatch` didn't account for the in-flight streaming accumulator.

## Root cause

When the model fans out parallel client tool calls in one streaming step:

1. The fast sibling resolves mid-stream and is applied to `_streamingAssistant`.
2. Its `autoContinue` schedules the continuation and arms the 50ms barrier timer.
3. The timer fires while the assistant message lives **only** in `_streamingAssistant` — not yet persisted.
4. `_hasIncompleteToolBatch()` scanned only `this.messages`, so it inspected a stale prior leaf, reported "not mid-batch", and the fast path fired the continuation — **bypassing the barrier**.
5. The stream ends and persists with the slow sibling still `input-available`; the continuation's transcript repair errors it.
6. The slow RPC result arrives 2–5s later — too late.

A `NOTE` comment had claimed this was safe because "the turn queue persists the (now settled) result before the continuation's repair runs". That holds for a fast sibling that settles before end-of-stream, but **not** for a slow RPC sibling that settles after.

## Fix

`_hasIncompleteToolBatch()` now inspects the in-flight `_streamingAssistant` accumulator first (the true current leaf), falling back to the latest persisted assistant message. A slow sibling still `input-available` in the accumulator now keeps the barrier closed until its result lands or the existing 60s timeout elapses. The part-scan is extracted into a shared `_partsAreMidBatch` helper, and the misleading `NOTE` is corrected.

## ai-chat (defensive)

The same guard is applied to `@cloudflare/ai-chat`'s `_hasIncompleteToolBatch`. ai-chat's barrier runs in the **post-stream** continuation turn (where `_streamingMessage` is already null), so it does **not** exhibit the bypass today — this keeps the two implementations aligned and mirrors the existing `hasPendingInteraction()`, which already checks `_streamingMessage`.

## Tests

- **Deterministic guard** (`detectsMidBatchInStreamingAccumulator`): a settled `tc-fast` beside a pending `tc-slow` exposed only in the accumulator must report mid-batch; a cleared accumulator + empty history must not (no false positive).
- **Integration test**: two parallel client tools streamed, fast answered mid-stream with `autoContinue`, slow answered *after* the stream ends. Asserts the slow tool ends `output-available` (not errored) with exactly one continuation.
- Both verified to **fail without the fix** — the integration test reproduces the exact customer signature (`expected 'output-error' to be 'output-available'`).

## Verification

- `npm run format`, `oxlint` — 0 warnings / 0 errors
- `npm run typecheck` — 91/91 projects
- `@cloudflare/think` — 495 passed
- `@cloudflare/ai-chat` — 566 passed

## Changeset

`patch` for `@cloudflare/think` (real fix) and `@cloudflare/ai-chat` (defensive symmetry).

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1663" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/streaming-accumulator-tool-batch-barrier.md
+++ b/.changeset/streaming-accumulator-tool-batch-barrier.md
@@ -1,0 +1,27 @@
+---
+"@cloudflare/think": patch
+"@cloudflare/ai-chat": patch
+---
+
+fix(think): hold the auto-continuation barrier for a slow tool sibling pending only in the streaming accumulator (#1649 follow-up)
+
+When the model fans out parallel client tool calls in one streaming step and a
+fast sibling resolves mid-stream, its `autoContinue` schedules the continuation
+and arms the 50ms barrier timer. That timer can fire while the assistant
+message still lives only in the streaming accumulator (`_streamingAssistant`)
+and has not yet been persisted. `_hasIncompleteToolBatch()` only scanned
+`this.messages`, so it saw a stale (prior) leaf, reported "not mid-batch", and
+the fast path fired the continuation — bypassing the barrier. When the stream
+then persisted with the slow sibling still `input-available`, the continuation's
+transcript repair errored it with "The tool call was interrupted before a result
+was recorded", and the slow RPC result (2–5s later) arrived too late.
+
+`_hasIncompleteToolBatch()` now inspects the streaming accumulator first (the
+true in-flight leaf), falling back to the latest persisted assistant message,
+so a slow sibling keeps the barrier closed until its result lands or the
+existing 60s timeout elapses.
+
+The same guard is applied to `@cloudflare/ai-chat`'s `_hasIncompleteToolBatch`
+for symmetry. ai-chat's barrier runs in the post-stream continuation turn (where
+`_streamingMessage` is already null), so it does not exhibit the bypass today —
+this keeps the two implementations aligned and safe against future flow changes.
--- a/packages/ai-chat/src/index.ts
+++ b/packages/ai-chat/src/index.ts
@@ -2189,7 +2189,8 @@ export class AIChatAgent<
     const streamingParts = this._streamingMessage?.parts;
     if (streamingParts) {
       return AIChatAgent._partsAreMidBatch(
-        streamingParts as ReadonlyArray<Record<string, unknown>>
+        streamingParts as ReadonlyArray<Record<string, unknown>>,
+        { streaming: true }
       );
     }
     // Zero-allocation backward scan for the latest assistant message — this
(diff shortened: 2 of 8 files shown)


threepointone:
## Follow-up: deeper review + added coverage

Did a full edge-case review and added two more layers of coverage (pushed in `00d50bb9`).

### Edge cases reviewed
- **Post-stream cache-coherence window** — at the instant `_streamingAssistant` is nulled, the barrier falls back to `this.messages`. There's a *theoretical* window where the persisted leaf isn't in the cache yet, but it's (a) pre-existing (old code only ever read `this.messages`), and (b) empirically closed: the integration/e2e tests answer the slow tool **after** stream-end and hold across 10+ barrier poll ticks without firing early. The persist→null ordering plus the intervening awaits keep the cache coherent.
- **`hasPendingInteraction` / `waitUntilStable`** scan only `this.messages`, but they're **recovery-only** paths that run post-restart with `_streamingAssistant === null` — no analogous gap.
- **ai-chat** — its only `_hasIncompleteToolBatch` caller (`_awaitPendingInteractionBarrier`) runs in the **post-stream** continuation turn where `_streamingMessage` is already null, so it isn't vulnerable. The mirror remains defensive-only (as labeled).

### Added coverage
1. **Approval-path unit test** — the slow sibling is `approval-requested` (not a result); confirms the barrier holds there too.
2. **Full `wrangler dev` e2e** (`parallel-client-tool.test.ts`) — a new `ThinkParallelClientToolE2EAgent` (deterministic mock model, two parallel client tools) driven over a **real WebSocket against a real worker process**: fast tool answered mid-stream, slow tool answered after stream-end. Asserts the slow tool settles `output-available` with exactly one continuation.

Both **fail without the fix** — the e2e reproduces the exact customer signature (`expected 'output-error' to be 'output-available'`) deterministically across all 3 retries.

Verification: in-process #1649 suite 11 passed, typecheck 91/91, lint clean, e2e passes against real `wrangler dev`.

threepointone:
clsoing this in facour of https://github.com/cloudflare/agents/pull/1667 for now

PR #1666, opened 2026-06-03
fix(mcp): bind OAuth PKCE verifiers to callback state

dcartertwo:
## Problem

`DurableObjectOAuthClientProvider` stores OAuth state per nonce, but stored the PKCE `code_verifier` in a single client/server slot.

When multiple OAuth attempts overlap for the same MCP server/client, a callback can arrive for one state while the provider reads the verifier from another attempt. The token exchange then fails with the upstream OAuth error:

```
Invalid PKCE code_verifier
```

This can happen when a user retries auth, opens multiple auth popups, completes an older popup, or hits reauth churn.

## Fix

Store PKCE verifier data by OAuth callback state instead of by client/server.

The MCP SDK calls `saveCodeVerifier(verifier)` without passing `state`, then later calls `redirectToAuthorization(authUrl)` with an authorization URL that contains both `state` and `code_challenge`. The provider now uses that `code_challenge` as the bridge to bind the saved verifier to the generated OAuth state without changing the MCP SDK provider interface.

Callback handling now runs token exchange in the returned state verifier context, so `codeVerifier()` resolves the verifier for that exact callback.

Stale/duplicate callbacks are treated idempotently once auth is already accepted or progressing. Their state is consumed to prevent replay, but verifier cleanup is left to the active completion path, direct reconnect cleanup, invalidation, or expiry so stale callbacks cannot delete the verifier currently in use.

## Tests

Adds coverage for:

- resolving PKCE verifiers by OAuth callback state;
- overlapping OAuth states for the same server/client;
- stale valid callbacks after auth is already progressing;
- callbacks that become stale while state validation is pending;
- stale OAuth error callbacks after auth is accepted/progressing;
- duplicate callbacks with the same active state;
- deprecated direct reconnect verifier cleanup;
- connection state while OAuth token exchange is pending.

Verification:

- `npm --workspace packages/agents run test:workers -- --run src/tests/mcp/create-oauth-provider.test.ts src/tests/mcp/client-manager.test.ts src/tests/mcp/client-connection.test.ts`
- `npm run check`

## Context

This fixes the MCP OAuth failure mode where downstream clients surface `Invalid PKCE code_verifier` after overlapping or stale auth windows. The manager-driven callback path is now concurrency-safe for overlapping attempts.

The deprecated direct `connect(..., { reconnect: { oauthCode } })` path remains best-effort because it does not carry OAuth state, but it now performs verifier cleanup after completion.

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1666" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/stateful-pkce-verifiers.md
+++ b/.changeset/stateful-pkce-verifiers.md
@@ -1,0 +1,7 @@
+---
+"agents": patch
+---
+
+Fix MCP OAuth PKCE verifier lookup for overlapping authorization attempts.
+
+`DurableObjectOAuthClientProvider` now binds pending PKCE verifiers to the OAuth callback state instead of storing a single verifier per client/server. Callback handling runs token exchange and verifier cleanup in the returned state's context, so older auth windows and retry churn no longer exchange an authorization code with another attempt's verifier.
--- a/packages/agents/src/mcp/client-connection.ts
+++ b/packages/agents/src/mcp/client-connection.ts
@@ -252,19 +252,26 @@ export class MCPClientConnection {
   /**
    * Complete OAuth authorization
    */
-  async completeAuthorization(code: string): Promise<void> {
-    if (this.connectionState !== MCPConnectionState.AUTHENTICATING) {
+  async completeAuthorization(
+    code: string,
+    options: { alreadyAccepted?: boolean } = {}
+  ): Promise<void> {
+    const expectedState = options.alreadyAccepted
+      ? MCPConnectionState.CONNECTING
+      : MCPConnectionState.AUTHENTICATING;
+    if (this.connectionState !== expectedState) {
       throw new Error(
-        "Connection must be in authenticating state to complete authorization"
+        `Connection must be in ${expectedState} state to complete authorization`
       );
+    }
+
+    if (!options.alreadyAccepted) {
+      this.connectionState = MCPConnectionState.CONNECTING;
     }
 
     try {
       // Finish OAuth by probing transports per configuration
       await this.finishAuthProbe(code);
-
-      // Mark as connecting
-      this.connectionState = MCPConnectionState.CONNECTING;
     } catch (error) {
       this.connectionState = MCPConnectionState.FAILED;
       throw error;
--- a/packages/agents/src/mcp/client.ts
+++ b/packages/agents/src/mcp/client.ts
@@ -641,6 +641,13 @@ export class MCPClientManager {
     state: string
   ): Promise<void> {
     try {
+      const stateValidation = await authProvider.checkState(state);
+      if (!stateValidation.valid) {
+        console.warn(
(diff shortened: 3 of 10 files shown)


mattzcarey:
cant find any issues regarding mcp or oauth spec. clanker came up with these 3 things which are kinda nits but seem nice to fix:

**1. Orphaned challenge-key verifiers never self-expire**

`saveCodeVerifier(v)` always writes `…/code_verifier_challenge/<challenge>`, and `redirectToAuthorization` only promotes it to `…/code_verifier/<nonce>` (and deletes the challenge key) when `state` + `code_challenge` are present *and* `parsed.serverId === this.serverId`. Any flow that starts (`saveCodeVerifier`) but never cleanly redirects — serverId mismatch, missing params, or redirect never happens — orphans a challenge verifier.

The expiry sweeps don't catch these:

| Path | Prefix listed | Sweeps challenge orphans? |
|---|---|---|
| `codeVerifier()` no-context expiry sweep | `…/code_verifier/` (trailing slash) | No — doesn't match `…/code_verifier_challenge/` |
| `checkState` expired | specific state key | No |
| `deleteCodeVerifier()` / `invalidateCredentials("verifier")` | `…/code_verifier` (no slash) | Yes |

So orphaned challenge verifiers have no TTL and are only cleaned by an explicit `deleteCodeVerifier`/`invalidate`, which an aborted flow may never reach — they can accumulate unbounded. Low severity (an unredeemed verifier is useless), but it means the 10-min expiry only really covers *state-bound* verifiers, not orphans. Cheapest fix: opportunistically drop expired challenge entries in `saveCodeVerifier`, or include the challenge prefix in the no-context expiry sweep.

**2. Stale-callback early-return consumes `state` without `checkState`**

The first early-return in `handleCallbackRequest` (the `!validation.valid` branch — `?error=…` or no code) on an already-accepted connection calls `consumeStaleOAuthState(state)` and returns `authSuccess: true` *without* ever calling `checkState`. Not exploitable (nonce is an unguessable `nanoid`, state keys are namespaced by `serverId`, nothing new is minted on an already-authenticated connection), but it reports success on a forged callback and deletes a state it never validated. Suggest running `checkState` on this path too — keep the idempotent success (don't tear down a working connection over a stale popup), but only `consumeState` when state validates, and `console.warn` on mismatch. The other two early-returns (post-`checkState`) are fine.

**3. `redirectToAuthorization` can now throw via getters**

The old `redirectToAuthorization` never touched `clientId`/`serverId`; the new one reads both through getters that throw if unset. Unreachable in the SDK flow (`saveCodeVerifier` dereferences both first, so it'd throw earlier), but it's a new throw-site in a method that previously couldn't fail. Belt-and-suspenders: read `this._serverId_`/`this._clientId_` directly and `return` early when unset, consistent with the existing fail-soft `if (!state || !codeChallenge) return`.

mattzcarey (approved):
looks good

PR #1667, opened 2026-06-03
fix(think): event-driven auto-continuation barrier + stream-active gate (fixes #1649)

### LINKED ISSUE #1649: _scheduleAutoContinuation 50ms timer fires before all parallel client-tool results arrive → MissingToolResultsError

threepointone:
## Summary

Reworks `@cloudflare/think`'s parallel-tool **auto-continuation barrier** (added in #1649) into a purely **event-driven** mechanism, and adds a **stream-active gate** that fixes #1649's headline `MissingToolResultsError` race.

Fixes #1649. Follow-up tracked as #1650. Related to the think↔ai-chat barrier unification (#1642).

## Background

When the model emits several tool calls in one step, the client answers each independently and sends a `cf_agent_tool_result` (with `autoContinue`) per result. A fast tool's result must **not** trigger inference while a slower sibling is still unanswered — that feeds the provider an incomplete tool-result set (`MissingToolResultsError`) or, via the transcript-repair backstop, errors the in-flight sibling and runs a spurious continuation.

#1649 introduced a barrier, but bounded the wait with a fixed **60s timeout** that *fired through* on expiry. This PR replaces that, and then closes the deeper mid-stream race that the timeout-based fix never addressed.

## Part 1 — Event-driven barrier (replaces the fixed timer)

A timer is the wrong primary mechanism for "wait until the client finishes answering the batch":

- **Human-in-the-loop fire-through.** A no-`execute` tool (`ask_user` / `display_ui`) parked at `input-available` for as long as the user takes to answer was errored after 60s, mid-answer.
- **Orphan isolate hold.** A disconnected client pinned the isolate alive for 60s.

Auto-continuation is *only ever* triggered by a tool-result/approval event, so the barrier now keys off events:

- On the coalesce timer, **drain in-flight applies** (`_drainInteractionApplies` — awaits the apply tail and re-reads it so a sibling enqueued mid-drain is still awaited), re-check, and if a sibling is still unanswered **return without firing and without holding the isolate**, leaving `pending` in place.
- The next sibling's result re-arms the check; the continuation fires exactly once when the final sibling lands.
- Lost to eviction between siblings? The final result re-creates `pending` from the persisted transcript — **self-healing across hibernation**.
- A true orphan never auto-continues and never pins the isolate.

Removes `AUTO_CONTINUATION_PENDING_TOOL_TIMEOUT_MS` (and its `console.warn`).

**Errored-sibling completion:** the client sends `autoContinue: false` for an errored result, so that event no longer schedules a continuation. When a sibling already opted in, `_rearmPendingAutoContinuationForBatch` re-arms the check so the batch still continues exactly once — without ever creating a continuation for a standalone errored tool.

## Part 2 — Stream-active gate (fixes the #1649 headline race)

The event-driven change alone did **not** fix #1649's actual reproduction. Per @abhagsain's debug-log analysis on the issue: the model emits parallel tool calls **sequentially within one step**, so a fast client tool can resolve and round-trip its result **while the model is still streaming the slower siblings**. At that moment the siblings exist **nowhere** — not in `this.messages`, not in the in-flight `_streamingAssistant` accumulator — so no completeness check can see them. The barrier bypassed, the continuation was enqueued, and when it ran (after persist) it repaired the now-materialized-but-still-pending siblings to errored → death spiral.

```
T~250ms: deleteSlides (fast) result arrives → barrier checks
         accumulator = [reasoning(STREAMING), deleteSlides(output-available)]
         → modifySlide x2 NOT EMITTED YET → bypass → continuation enqueued
T~800ms: stream ends, persists all 3 parts (modifySlide x2 input-available)
         → continuation runs → repairs modifySlide x2 to errored
```

The only signal that "more tool calls may still arrive" is that the stream is open:

- **Stream-active gate:** `_fireAutoContinuationWhenStable` returns early while `_streamingAssistant != null` (also re-checked in the drain `.finally`).
- **Stream-finalize re-trigger:** `_onStreamingTurnFinalized` (at the two normal stream-finalize sites) clears the accumulator and re-runs the barrier. Essential for an **all-fast batch** whose every result landed mid-stream — there's no later tool-result event to re-arm it, so without this re-check the held continuation would deadlock. Abort/recovery paths keep a plain clear; recovery re-runs the turn and its own finalize re-triggers the held barrier.

Also corrected the prior NOTE that wrongly claimed turn-queue ordering made the mid-stream bypass safe — the logs disprove it.

## Double-fire safety

`_continuationBarrierActive` absorbs a sibling that re-arms the timer mid-drain; the fire/return decision in the drain's `.finally` is synchronous, and `_fireAutoContinuation` cancels the pending timer.

## Tests

`packages/think/src/tests/client-tools.test.ts` (real Workers runtime via `vitest-pool-workers`):

- **human-in-the-loop**: parked no-`execute` tool — no fire-through, no repair; fires once on answer.
- **errored sibling completes the batch** (`autoContinue: false`): continues once.
- **standalone errored tool**: does not auto-continue.
- **self-heals across eviction** mid-batch.
- **sibling emitted LATER in the same stream** (the #1649 headline repro, new `createMidStreamParallelToolModel`): fast tool resolved mid-stream, no premature continuation, no repair of the slow tool, fires once when it answers.
- **all-fast batch resolved entirely mid-stream**: fires once via the stream-finalize re-check (guards the deadlock).

**Confidence-checked:** the two mid-stream tests **fail** with the gate + finalize re-trigger disabled (the all-fast case fires 2 continuations) and pass with them. Full `client-tools` suite **72/72**; `oxlint` clean; all 91 projects typecheck.

## Scope / follow-ups

- `@cloudflare/ai-chat` keeps the bounded-wait barrier for now (its barrier runs inside the queued continuation turn); event-driven there requires moving the gate before queueing — tracked with #1642.
- Observability: the old `console.warn` is gone; a debug-level (non-firing) diagnostic for long-pending orphans is best added under #1642 rather than reintroducing a timer.

## Changeset

`patch` for `@cloudflare/think`.


threepointone:
### Ported coverage from the abandoned #1663 attempt

Reviewed #1663 (threepointone's earlier attempt at #1649). It tried to fix the bug by making `_hasIncompleteToolBatch` read the streaming accumulator (`_partsAreMidBatch`). That approach never landed — it can't see a sibling the model hasn't streamed yet, which is the actual race (per @abhagsain's debug logs on the issue). Its four **accumulator-scan unit tests** target that implementation, which this PR deliberately does not have (we gate on `_streamingAssistant != null` and never inspect sibling part-states), so they were **not** ported; their behavioral intent is already covered end-to-end by the stream-active-gate tests.

Notably, #1663 shipped a unit test, an integration test, and a full wrangler-dev e2e — all claimed to "fail without the fix" and reproduce the customer signature "deterministically across all 3 retries" — yet it didn't fix the customer's bug. The mock models put both tool calls into the accumulator before the fast result was answered, so the slow sibling was always visible. The lesson: **a wrangler e2e is only as faithful as its mock model's emission ordering.** Our headline test asserts `streamingToolCallState("tc-slow") === undefined` at the moment the fast tool is answered — the exact condition #1663's tests never created.

Ported the two scenarios that carry real signal for this design (commit 68f512ba):

- **Different ordering** — both parallel calls streamed up front and visible, fast answered mid-stream, slow answered only *after* stream end. Guards the gate across stream finalize.
- **Approval-path variant** — a parallel batch whose slow sibling awaits an *approval* response; locks down that `_hasIncompleteToolBatch` treats `approval-requested` as pending so the barrier holds until approval, then fires once.

`client-tools` suite now **74/74**; lint clean; 91/91 typecheck.

PR #1682, opened 2026-06-05
Migrate to pnpm.

cpojer:
This PR is the first one in a series to improve the repo health. pnpm is faster, more secure, and stricter.

There were a few undeclared deps which switching to pnpm revealed.

In a next step, I'll simplify the repo by migrating to Vite+.

(I have no idea about basically anything here, so please tell me what to test manually!)


cpojer:
Rebased + I fixed CI, I think.

Issue #1620, opened 2026-05-29
Surface a live "recovering…" status to chat clients during durable recovery


threepointone:
## Background

Follow-up to the deploy-churn recovery hardening (#1615, #1617, #1618, #1619).

When a chat turn is interrupted (deploy/eviction) and Think recovers it, there is
currently **no client-facing "in progress" signal** while recovery runs. The
client sees the partial transcript with no indication that the turn is being
recovered — it looks idle/stuck until recovery completes (or, now via #1619,
until a terminal error is replayed on connect).

#1619 fixed the **terminal** half (failed/exhausted turns are surfaced, and
replayed on reconnect — option "A", plus keeping benign skips silent, "C"). This
issue tracks the remaining **progress** half ("B"): show clients a
"recovering…" indicator so an interrupted-but-recovering turn reads as *working*
rather than frozen.

## Why it's separate

Unlike #1619 (which reused the existing terminal `MSG_CHAT_RESPONSE` and needed
no client changes), a "recovering" indicator is a **new client-facing
capability** and spans multiple layers:

1. **Protocol:** a new chat message type (e.g. `MSG_CHAT_RECOVERING`) in
   `agents/chat`, broadcast on recovery detect/schedule and cleared on
   resolution.
2. **Hydration:** the status must be durable and replayed on connect (like the
   terminal status in #1619), or a client that connects mid-recovery still sees
   nothing — this is the load-bearing requirement for the deploy/reconnect-storm
   case.
3. **Debounce + reliable clear:** recovery is detected many times under churn,
   so the signal must be idempotent (only on transition) and cleared on **every**
   terminal path (completed/failed/exhausted/skipped) or a spinner spins forever.
4. **Client rendering:** `useAgentChat` (and downstream SPAs) must render the
   new status; needs browser validation.

## Acceptance

- A connected client shows a "recovering…" state for the duration of durable
  chat recovery, and a client that connects mid-recovery learns it on connect.
- The indicator clears on every terminal outcome; benign skips do not produce a
  stuck indicator.
- Backward-compatible: clients that don't understand the new signal ignore it.

## Notes

Investigation + the A/B/C/D design options are written up in
`examples/deploy-churn/INVESTIGATION.md` (added in #1618).

PR #1690, opened 2026-06-06
fix(chat): surface terminal recovery outcome to clients that reconnect after it ended (#1645)

### LINKED ISSUE #1645: ai-chat: chat-recovery exhaustion has no terminal-status hydration (disconnected client misses the terminal banner on reconnect)

threepointone:
## Summary

Fixes #1645. When a durable chat turn **exhausts recovery** (e.g. during a deploy / WebSocket reconnect storm) while **no client is connected**, the terminal error was only broadcast transiently. A client that connected afterward ran the standard resume probe, was told `RESUME_NONE`, and the failed turn stayed **frozen** — it never learned the turn had ended.

This affects both `@cloudflare/ai-chat` and `@cloudflare/think`.

## Root cause

The shared `useAgentChat` client only surfaces a terminal `{ done: true, error: true }` frame as `useChat.error` when that frame arrives on the transport's **stream reader** — i.e. via the resume handshake. A bare broadcast frame received *outside* an active stream read is dropped.

- `@cloudflare/ai-chat` had **no durable terminal record at all**.
- `@cloudflare/think` persisted one, but replayed it only as a **bare on-connect frame** that the client drops.

So both presented the identical frozen-turn symptom.

## Fix

Persist the terminal outcome durably and replay it over the **resume handshake** on the next reconnect:

```
STREAM_RESUMING (id = terminal requestId)
  → client STREAM_RESUME_ACK
    → terminal error frame on the resumed stream
```

This is the only path that becomes `useChat.error` on the real client. The record is **retained across the ACK** so concurrent reconnects (multiple tabs) each learn the outcome.

### Record lifecycle — cleared when superseded

The durable record is dropped on every path that makes a stale exhaustion irrelevant:

- **a new client-initiated chat request**,
- **any later turn that completes successfully** — including turns driven purely server-side via `saveMessages` / auto-continuation with no client request in between (closes a gap where `ai-chat` only cleared on a new client submit),
- **clearing the conversation** — both the WS `chat-clear` path and the programmatic `clearMessages()` API — so a stale exhaustion can't replay onto a freshly-emptied chat.

## Cleanup

- `@cloudflare/think` no longer emits the bare terminal frame in `_buildIdleConnectMessages` — it was dead for the real client and redundant with the handshake. The live **"recovering…"** status (#1620) is still replayed there; it's a plain status frame the client handles on connect.
- **Naming unified** across both packages so the implementations are structurally parallel:
  `_recordChatTerminal` / `_clearChatTerminal` / `_pendingChatTerminal` / `_replayTerminalOnResume` / `_replayTerminalOnAck`, keyed by the shared `CHAT_LAST_TERMINAL_KEY`.

## Tests

Real **Durable Object + real WebSocket** tests driving the exact reconnect handshake the client sends. Every new guard was **verified red without its fix**:

- replays the terminal exhaustion to a client that reconnects after it ended — **ai-chat + think**
- clears the record when a later server-side (`saveMessages`) turn succeeds — **ai-chat**
- drops the record when the conversation is cleared — **ai-chat + think**
- existing think hydration tests updated to assert the **durable record** (which the handshake replays) instead of the removed on-connect frame

### Why not a browser Playwright test?

Neither package has a harness that mounts `useAgentChat` in a browser, and the gap lives at the protocol/transport layer. These real-DO + real-WS tests drive the exact client handshake and are the faithful equivalent; the resumed-stream → `useChat.error` surfacing is covered by existing transport tests. A browser-mount test is the only remaining inch of confidence and can be added as a follow-up.

## Verification

- `@cloudflare/ai-chat`: **563** tests pass
- `@cloudflare/think`: **553** tests pass
- `pnpm run check` green (sherif + export checks + oxfmt + oxlint + typecheck, 92 projects)

## Changeset

`.changeset/chat-terminal-replay-on-reconnect.md` — both packages, `patch`.

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1690" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


threepointone:
## Follow-up: closed a reconnect race in Think (review feedback)

Validated a review comment — it was **correct**. `@cloudflare/ai-chat` eagerly clears the pending terminal record at submit time so a stale exhaustion can't replay once the user has moved on. `@cloudflare/think`'s `_handleChatRequest` was missing that eager clear and relied only on the turn-completion clear.

**The race:** after a user submits a new turn but before it starts streaming (the enqueue / pre-stream window), a reconnecting client hits the terminal-replay branch (no active stream, no pending continuation) and gets served the *previous* failed turn's error over the resume handshake — even though they've moved on. Harmless before this PR (the bare on-connect frame was dropped by the client); user-visible now that the terminal is delivered over the handshake.

**Fix:** `await this._clearChatTerminal()` in `_handleChatRequest` right after the concurrency decision accepts the submit, mirroring ai-chat.

**Test:** drives a real WS submit with the new turn parked in `beforeStep` (`setBeforeStepAsyncDelay`) so the observed clear must be the *eager submit-time* clear, not the turn-completion clear. Verified red without the fix (the stale `root-old` terminal survives the submit).

think now at 554 tests; `pnpm run check` green.

threepointone:
## Follow-up: clear terminal on aborted server-side turns (review feedback)

Validated another review comment — **correct**. The drain loop in `_runExclusiveChatTurn` cleared the terminal record only on `chatResult.status === "completed"`. An **aborted** server-side turn (`saveMessages` / `continueLastTurn` cancelled mid-stream via `options.signal`) left the stale terminal in place, so a post-exhaustion abort would replay the old error to the next reconnecting client over the resume handshake.

Think already clears on any non-error/non-interrupted status (`_recordTerminalChatStatus`), so this was also a cross-package inconsistency. (The client-submit path clears eagerly, so the gap was only reachable for purely server-side turns — exactly as the reviewer noted.)

**Fix:** clear on `"completed"` **or** `"aborted"` — only a fresh *error* should leave a terminal to replay.

**Test:** drive a server-side turn aborted via a pre-aborted signal (the stream loop breaks immediately and pushes an `"aborted"` result through the drain loop) after an exhaustion, then assert the record is cleared. Verified red without the fix (stale terminal survives the abort).

ai-chat now at 564 tests; `pnpm run check` green.

threepointone:
## Follow-up: ordering observation (broadcast vs persist on exhaustion)

The observation is factually correct — ai-chat persists-then-broadcasts, Think broadcasts-then-persists — but the stated risk needs a correction, and the fix is narrower than "flip Think to match ai-chat".

**Ordering can't rescue a write that itself fails.** If the terminal-record write rejects mid-deploy, the record is missing regardless of whether it ran before or after the broadcast — persist-first doesn't save it (the persist *is* the failing op). When storage is healthy, both orderings persist *and* broadcast fine. So flipping Think to persist-first gains **no** reconnect reliability under failure, while **losing** Think's deliberate guarantee that a connected client still sees the terminal banner when storage is rejecting (the broadcast touches no storage).

**The real (marginal) issue is intra-storage:** Think did two storage writes in the order `_markRecoveredSubmissionInterrupted` → `_recordTerminalChatStatus`, leaving the #1645 record (the reconnect-critical one) **last** — so a failure in the unrelated submission-row write would skip it. ai-chat doesn't have that dependency because its terminal write is first.

**Fix (`c4216248`):** keep broadcast-first, but hoist `_recordTerminalChatStatus` ahead of `_markRecoveredSubmissionInterrupted` so the durable record isn't hostage to the bookkeeping write. The two writes are independent (durable terminal key + recovering-status clear vs. the `cf_think_submissions` row), so reordering is safe. The divergence from ai-chat is now documented in-code so nobody "fixes" it later.

No behavior change when storage is healthy; existing exhaustion tests (which assert both the durable record and the interrupted submission row) cover it. think still 554 tests; `pnpm run check` green.

threepointone:
## Follow-up: durably record non-exhaustion errors in ai-chat (review feedback)

Validated — **correct**, and squarely within #1645's intent so I closed it. ai-chat only persisted a terminal record on recovery **exhaustion** (`_exhaustChatRecovery`). A terminal non-exhaustion error (a provider 500 surfaced as a stream `error` part) was broadcast transiently and returned `status: "error"` to the drain loop, which neither cleared nor recorded it — so a client disconnected at that moment got `RESUME_NONE` on reconnect and stayed frozen. That's the same failure #1645 fixes; ai-chat had only closed it for the exhaustion sub-case.

Think already records on `error` (`_fireResponseHook` → `_recordTerminalChatStatus`), so this was also a cross-package parity gap.

**Fix:** in the drain loop, record a durable terminal on `status === "error"` (body = the error message). Recoverable failures (deploy/eviction/stall) don't reach the drain loop as `error` — they arrive as `aborted`, or not at all (isolate gone, with exhaustion recording its own terminal) — so this can't pre-empt recovery.

**Test:** drive a server-side turn that ends in a stream error, then assert the durable record is written **and** replays over the resume handshake on reconnect (terminal error frame, not a bare `RESUME_NONE`). Verified red without the fix (record is `null`). Changeset updated to note the new behavior.

ai-chat now at 565 tests; `pnpm run check` green. The terminal lifecycle is now at full parity across both packages — recorded on error/exhaustion, cleared on completed/aborted/new-submit/conversation-clear.

PR #1697, opened 2026-06-08
Migrate to Vite+.

cpojer:
This PR moves the agents repo to Vite+. Please see https://viteplus.dev/ for how to set up Vite+.

* Package management is done through `vp install`, using pnpm under the hood.
* Oxlint and Oxfmt configs were moved into `vite.config.ts`, and should be used via `vp check --fix`.
* Nx was replaced by Vite Task.
* Use `vite-plus/pack` instead of `tsdown`.
* Use `vite-plus/test` instead of `vitest`.
* Replaced hsuky/lint-staged with Vite+.

In a follow-up I'd like to also add type checking to `vp check` to simplify further.


cpojer:
Updated all GitHub Actions to use pinned actions for security.

pnpm-workspace.yaml
```diff
@@ -19,9 +19,22 @@ allowBuilds:
...
-  nx: true
   sharp: true
   workerd: true
 minimumReleaseAgeExclude:
   - "@cloudflare/*"
   - "@typescript/*"
+catalog:
+  vite: npm:@voidzero-dev/vite-plus-core@latest
+  vitest: npm:@voidzero-dev/vite-plus-test@latest
+  vite-plus: latest
```
fengmk2: I think we should use range version (e.g.: `^0.1.24`) instead of `latest` from now on. And I will also change the `vp migrate` behavior too.

cpojer: Thank you, fixed!

cpojer:
Rebased and resolved conflicts.

fengmk2:
LGTM 👍

.github/workflows/bonk.yml
```diff
@@ -21,10 +21,10 @@ jobs:
       contents: write
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
```
threepointone, line 24: why are we pinning these versions? is that better? won't we miss updates?

.github/changeset-version.ts
```diff
@@ -2,15 +2,15 @@ import { execSync } from "node:child_process";
 
 // This script is used by the `release.yml` workflow to update the version of the packages being released.
 // The standard step is only to run `changeset version` but this does not update the pnpm lockfile.
-// So we also run `pnpm install --lockfile-only`, which does this update.
+// So we also run `vp install --lockfile-only`, which does this update.
 // This is a workaround until this is handled automatically by `changeset version`.
 // See https://github.com/changesets/changesets/issues/421.
-execSync("pnpm exec changeset version", {
+execSync("vp exec changeset version", {
```
threepointone, line 8: is theree a vpx command? could be nice.

examples/assistant/vite.config.ts
```diff
@@ -1,9 +1,25 @@
...
 
 export default defineConfig({
-  plugins: [think(), react(), cloudflare(), tailwindcss()]
+  plugins: lazyPlugins(async () => {
+    const [
+      { cloudflare },
+      { default: tailwindcss },
+      { default: react },
+      { think }
+    ] = await Promise.all([
```
threepointone, line 11: why is this better?

examples/telnyx-voice-agent/package.json
```diff
@@ -4,9 +4,9 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "start": "vite dev",
-    "build": "vite build",
-    "deploy": "vite build && wrangler deploy",
+    "start": "vp dev",
+    "build": "vp run build:vite",
```
threepointone, line 8: we don't have a build:vite command here? why is this not vp build?

threepointone: oh I see it's in the vite config. but then should it be vp run? not vp?

package.json
```diff
@@ -49,40 +37,31 @@
...
   },
-  "lint-staged": {
-    "*.{js,mjs,cjs,jsx,ts,mts,cts,tsx,vue,astro,svelte,css}": [
-      "oxfmt --write"
-    ]
-  },
-  "packageManager": "pnpm@11.5.2"
+  "packageManager": "pnpm@11.5.2",
+  "overrides": {
+    "esbuild": "^0.25.1"
```
threepointone: why this override?

threepointone (commented):
some questions for my understanding

.github/workflows/bonk.yml
cpojer: Security.

We did this for all VoidZero projects to avoid hacked actions from taking over our projects. We should frequently update actions (manually, or through renovate), but keep them pinned.

.github/changeset-version.ts
cpojer: Yes there is!

examples/assistant/vite.config.ts
cpojer: With Vite+, `vite.config.ts` is loaded in many workflows such as `vp lint` or `vp fmt`. Some plugins take a lot of time to load, making all tool invocations slower. We added support for "lazy plugins" so we don't have to wait on the ecosystem to optimize their plugin init times.

cpojer: Ideally, `plugins` would be an async function but the Vite team thought it wasn't worth adding that API surface at this time.

examples/telnyx-voice-agent/package.json
cpojer: It is `vp run`.

package.json
cpojer: Removed. Sorry that was a leftover.

PR #1698, opened 2026-06-08
Allow `callable` to be used as a function.

cpojer:
Decorators are problematic. They have been stuck in standardization limbo for 12 years, there are a lot of versions, and they have recently been bumped back on the path to standardization. Vite 8 does not currently support decorators, which means the `agents` package has to add Babel to pre-process JS which slows down compilation quite a bit. With agents, the value of decorators as syntax sugar may not be so clear anymore.

This PR changes `callable` into a hybrid decorator and function so you can instead define callback functions like this:

```
import { Agent, callable } from "agents";

export class GreetingAgent extends Agent {
  greet = callable(async (name: string) => {
    return `Hello, ${name}!`;
  });
}
```

If you are on board with merging this, I would like to follow it up with an option for the Vite plugin to skip the decorator plugin for faster compilation, and in the future we could choose to flip the default to disable decorators (+ codemod) and possibly even remove it.

### DIFF
--- a/.changeset/function-callable-methods.md
+++ b/.changeset/function-callable-methods.md
@@ -1,0 +1,5 @@
+---
+"agents": minor
+---
+
+Add a function-wrapper form for callable methods, allowing class fields like `greet = callable(fn, metadata)` in addition to the existing `@callable()` decorator.
--- a/docs/agent-class.md
+++ b/docs/agent-class.md
@@ -242,7 +242,7 @@ class MyAgent extends Agent {
 
 ### RPC and Callable Methods
 
-`agents` take Durable Objects RPC one step forward by implementing RPC through WebSockets, so clients can also call methods on the Agent directly. To make a method callable through WS, developers can use the `@callable` decorator. Methods can return a serializable value or a stream (when using `@callable({ stream: true })`).
+`agents` take Durable Objects RPC one step forward by implementing RPC through WebSockets, so clients can also call methods on the Agent directly. To make a method callable through WS, developers can use the `@callable` decorator. Methods can return a serializable value or a stream (when using `@callable({ streaming: true })`).
 
 ```ts
 class MyAgent extends Agent {
--- a/docs/callable-methods.md
+++ b/docs/callable-methods.md
@@ -127,6 +127,30 @@ export class CounterAgent extends Agent<Env, State> {
 }
 ```
 
+You can also wrap a function-valued class field with `callable()` if you do not
+want to use decorators:
+
+```typescript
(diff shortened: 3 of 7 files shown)


Boshen (commented):
Context: [ECMAScript decorators](https://github.com/tc39/proposal-decorators) were moved from Stage 3 to [Stage 2.7](https://tc39.es/process-document/#stage-2.7) in [tc39/proposals@cd2396f](https://github.com/tc39/proposals/commit/cd2396f40e8bcfc76162bcc2d3340024f236bfe0).

At present, I don’t recommend using them, since we only use this syntax in a single place. The lack of tooling support and the confusing `tsconfig` setup mentioned in the docs are what prompted this PR.

I have a few ideas for making ECMAScript decorators generally available in the Vite ecosystem, but that will take a while because it depends on the standard [TC39 process](https://tc39.es/process-document).

threepointone:
I love that this has backward compat. I need to check some stuff with making sure regular rpc works with this and I'll land it (it might not, I think that's the problem with instance props and why we went with decorators here)

ingrave-solucoes:
### Code review

Found 3 critical issues:

1. **Metadata storage inconsistency between decorator and function forms** (packages/agents/src/index.ts:129-131)

The current implementation uses `callableMetadata.set(target, metadata)` which relies on function reference identity. When using `callable(async (name: string) => {...})`, the wrapped function has a different reference than the original, which could break metadata lookup in `_isCallable()`.

https://github.com/cloudflare/agents/blob/b9a45762285a3c0304f5e507e83f60771d83298b/packages/agents/src/index.ts#L129-L131

2. **Type safety compromise in hybrid approach** (packages/agents/src/index.ts:123-135)

The hybrid approach requires complex conditional types that could compromise type safety. The current implementation only supports the decorator pattern cleanly, and adding function wrapping creates type complexity that may lead to runtime errors.

https://github.com/cloudflare/agents/blob/b9a45762285a3c0304f5e507e83f60771d83298b/packages/agents/src/index.ts#L123-L135

3. **Backward compatibility risks with auto-wrapping** (packages/agents/src/index.ts:847-906)

The `_autoWrapCustomMethods()` function automatically wraps methods with agent context. This could interfere with the new function form and potentially break existing functionality that relies on the current decorator behavior.

https://github.com/cloudflare/agents/blob/b9a45762285a3c0304f5e507e83f60771d83298b/packages/agents/src/index.ts#L847-L906

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1720, opened 2026-06-09
feat(memory): single authoritative token accounting for compaction (fire + boundary)

### LINKED ISSUE #1593: Auto-compaction trigger fires, but compaction doesn't run due to internal token counting

mattzcarey:
Fixes #1593

## Design

The Session derives one authoritative context size in model tokens, and that single number drives both decisions: whether to compact, and where to cut. Because both decisions read the same number, they cannot disagree (the #1593 failure was exactly such a disagreement).

This is how [earendil-works/pi](https://github.com/earendil-works/pi) does it (`core/compaction/compaction.ts`), and [hermes-agent](https://github.com/nousresearch/hermes-agent) (`agent/context_compressor.py`) follows the same rule: model-reported usage is a whole-prompt signal, per-message decisions use a cheap proportional heuristic, and the two scales never mix.

### Counting

`Session._knownContextTokens()` resolves in priority order:

1. The `compactAfter()` counter, if configured. It is whole-prompt by signature, so `() => lastUsage.inputTokens` is legal and documented.
2. Usage metadata on assistant messages. `SessionMessage` gained `metadata?: unknown` (an AI SDK `UIMessage` passes straight through), and `metadata.usage` / `metadata.totalUsage` is read the way pi reads it: `calculateContextTokens` returns `totalTokens` when present, else the sum of the components. The context estimate is the last reported usage plus the heuristic for any newer messages. When messages carry usage, configuration is one line:
   ```ts
   session.compactAfter(100_000);
   ```
3. Neither. The chars/4 heuristic applies, as before.

### Compaction

The same total flows to `createCompactFunction` as `CompactContext.contextTokens`. The tail walk is always the plain heuristic walk; when an authoritative total is known, the budget is converted into heuristic units (`tailTokenBudget * heuristic / contextTokens`, equivalent to scaling every per-message estimate up to the model's scale). The walk's distribution is unchanged and the budget means model tokens.

That is the whole #1593 fix. Previously the Session's whole-prompt counter was adapted into per-message calls, so every message appeared to exceed the budget and `tailTokenBudget` silently degraded to `minTailMessages`.

There are no user counters in the boundary walk at all. pi has no per-message counter either; it is the footgun by which #1593 happened, so it is gone rather than documented against. Each remaining lever means one thing:

| Lever | Meaning |
|---|---|
| `compactAfter(threshold, { tokenCounter })` | whole-prompt total (fire + calibration) |
| assistant `metadata.usage` / `totalUsage` | same, zero config |
| `protectHead` / `tailTokenBudget` / `minTailMessages` | tunables, all in model tokens |

Breaking changes in the experimental package: `CompactOptions.tokenCounter`, the `CompactTokenCounter` type, and `CompactContext.tokenCounter` are removed. The changeset is `minor`.

## Docs

`docs/sessions.md` gets a single "Token counting" section (priority order, usage-metadata example, new utils in the helper list), `docs/think/index.md` points at it, and `experimental/session-memory/README.md` documents the priority order under Auto-Compaction.

## Tests

The tests assert the call contracts directly:

- The #1593 end-to-end repro checks that the whole-prompt counter only ever sees the full history, and that the boundary lands at the model-scale budget (`toMessageId: "tool-5"`, not the `minTailMessages` degradation).
- The same `createCompactFunction` no-ops on the raw heuristic and cuts correctly once the context supplies `contextTokens`.
- `compactAfter(1000)` with no counter, and usage attached to the appended assistant message, fires and cuts at the calibrated boundary.
- `calculateContextTokens` handles the totalTokens, input+output, prompt+completion and cached shapes. The newest assistant usage wins, usage on non-assistant messages is ignored, and no usage returns null.
- A non-finite or negative `contextTokens` is ignored and the raw heuristic applies.

Full experimental-memory suite: 193 passed (7 files). oxfmt and oxlint are clean, and the memory module has no typecheck errors.

## Notes for maintainers

- Usage extraction only reads assistant messages and only accepts positive token counts. The component-sum fallback mirrors pi and can over-count caches on providers where cached tokens are a subset of input tokens, which errs toward compacting earlier.
- Calibration assumes the model total distributes across messages roughly in proportion to the heuristic. pi and hermes make the same assumption for trailing estimates, and it is strictly better than the previous behavior, which used an identical value for every message.
- Users who need a custom tokenizer still have a path: put it on `compactAfter()` as a whole-prompt counter. Per-message precision in the walk bought little once the budget is correct at the model scale, and it was the only place the two token scales could be confused.


mattzcarey:
not convinced by this

mattzcarey:
hmmm still sloppy.

PR #1726, opened 2026-06-10
RFC: Rename "Think" to "Assistant" and export from `agents`.

cpojer:
This is an RFC to consider positioning and naming for the `agents` package and "Think" class. This PR is a non-breaking change to re-export "Think" as "Assistant" directly from `agents`.

In recent days I've tried to talk about or write about Think, and it generally fits poorly into sentences and breaks my brain. I was wondering, what is "Think" really, and could it be part of the `agents` package itself? This led me to add it to the `agents` package as an Assistant, which is a layer on top of Agent which is already there. It provides a more natural migration from an agent to an assistant, even though I realize that the mapping of what Think is does not 100% map to Assistant.

No hard feelings if you don't like it, or if it was already considered and decided against, but thought I'll open an RFC to discuss.

### DIFF
--- a/.changeset/assistant-alias.md
+++ b/.changeset/assistant-alias.md
@@ -1,0 +1,7 @@
+---
+"agents": minor
+---
+
+Add `agents/assistant`, an additive entrypoint that exports `Assistant` as an
+alias for `@cloudflare/think`'s `Think` base class while preserving the existing
+`Think` export.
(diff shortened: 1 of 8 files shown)


threepointone:
I could be convinced of this, but I'd hold off until Think is a bit more stable (we have a bunch of api changes incoming).

threepointone (changes requested):
blocking so we can revisit later

mattzcarey:
Hey, thanks for this. You have opened a bunch of discussions on our team about this direction. Would be happy to bring you in but I will close this PR to remove inflight work.

PR #1745, opened 2026-06-12
fix(agents): defer agent teardown to its own alarm invocation so canceled requests can't cut it short

### LINKED ISSUE #1625: McpAgent: ctx.waitUntil(agent.destroy()) on session teardown logs "waitUntil() tasks did not complete" when the SSE client has already disconnected (0.13.3)
### Summary

When an MCP client ends/drops its Streamable-HTTP session, the Worker logs a warning at `warn` level:

> waitUntil() tasks did not complete within the allowed time after invocation end and have been cancelled.

It's emitted against a **canceled** request (the client is already gone), so it's harmless — but it's noisy and shares a Cloudflare `fingerprint` with real `waitUntil` drops, which makes it hard to alert on the ones that actually matter.

### Environment

I'm building an MCP gateway / context augmenter using CF Workers.

- `agents@0.13.3` (latest). Verified the relevant code is **byte-identical in `0.13.2` and `0.13.3`** — see below.
- `McpAgent` via `agents/mcp`, Streamable HTTP transport.
- Cloudflare Workers, `executionModel: "stateless"` (the front Worker, not the DO).
- Mounted through `@cloudflare/workers-oauth-provider` `apiHandlers` (`McpAgent.serve('/mcp')`).

### Observed log

```json
{
  "level": "warn",
  "message": "waitUntil() tasks did not complete within the allowed time after invocation end and have been cancelled.",
  "$workers": {
    "event": { "request": { "method": "GET", "path": "/mcp" } },
    "outcome": "canceled",
    "executionModel": "stateless",
    "eventType": "fetch"
  }
}
```

### Root cause (as far as I can trace it)

Grepping the whole package, there are only three `waitUntil` call sites, and the **only request-scoped (Worker `ctx`) one** is in the Streamable-HTTP handler's `DELETE` branch:

```js
// dist/mcp/index.js:328 (identical in 0.13.2 and 0.13.3)
// src: packages/agents/src/mcp/index.ts — DELETE branch
ctx.waitUntil(agent.destroy().catch(() => {}));
```

`destroy()` itself awaits real I/O:

```js
// dist/index.js — Agent.destroy() (identical body in 0.13.2 and 0.13.3)
// src: packages/agents/src/index.ts
this._dropInternalTablesForDestroy();
await this.ctx.storage.deleteAlarm();
await this.ctx.storage.deleteAll();
this._disposables.dispose();
await this.mcp.dispose();
this._destroyed = true;
setTimeout(() => { this.ctx.abort("destroyed"); }, 0);
```

Session teardown is therefore bound to the **request's** `waitUntil`. When the client has already disconnected, the runtime gives that trailing task little to no post-invocation grace, so `destroy()` (storage deletes + `mcp.dispose()`) gets cancelled and logged. The `.catch(() => {})` doesn't help — the warning comes from the runtime cancelling the `waitUntil`, not from an unhandled rejection.

Note: I consistently see this attributed to a **canceled `GET /mcp`** (the SSE read channel), while the only `ctx.waitUntil` is on `DELETE`. My best guess is that the runtime stamps the orphaned task onto whichever invocation triggered isolate teardown (the canceled SSE GET), and/or clients send `DELETE` concurrently with closing the SSE stream. Either way the underlying task is the `destroy()` `waitUntil` — happy to be corrected if there's another teardown path.

### Why it's worth fixing

Teardown work bound to a *canceled* request's `waitUntil` is inherently unreliable, and:

- If `deleteAll()` is cut mid-way, the DO can be left partially torn down.
- It produces a steady stream of `warn` logs that share a fingerprint with genuine `waitUntil` failures, defeating alerting.

### Suggested fix

Run session-teardown cleanup **independently of the initiating request** rather than via `ctx.waitUntil` — e.g. self-schedule it on a DO alarm so it executes in its own invocation with a clean budget (the alarm infra is already there), or make `destroy()` idempotent/resumable so a cut-short teardown is finished on next wake. At minimum, document that this warning is expected on MCP session end and benign.

### Repro

1. Connect an MCP client over Streamable HTTP to an `McpAgent`.
2. End the session / drop the SSE connection (client disconnect, not a clean request/response).
3. Observe the `warn` in Workers Logs against the canceled `/mcp` request.

cjol:
This PR makes MCP session teardown durable by deferring `destroy()` to the agent's own alarm invocation rather than running it on the initiating request's `ctx.waitUntil`. Fixes #1625 (but not the excess warning on disconnection, which, as discussed in the issue comments is caused by a lower-level workerd trigger).

## Why

- `McpAgent`'s Streamable-HTTP session-DELETE handler ran `agent.destroy()` via the front Worker's `ctx.waitUntil`. By the time a DELETE lands, the client is usually already gone, and the runtime gives a canceled request's trailing work little to no post-invocation grace.
- `destroy()` is a multi-step I/O sequence: drop internal tables, `deleteAlarm`, `deleteAll` storage, dispose connections, abort. With a short budget, it was routinely cut short.
- The `_destroyed` flag was only written at the very end of the sequence, so an interrupted teardown was invisible: the DO's constructor (via `CREATE IF NOT EXISTS`) silently recreated the internal tables on next wake and resumed as a live agent. Half-deleted session DOs accumulated indefinitely.
- The `waitUntil() tasks did not complete` warning surfaced in the issue originates inside workerd's WebSocket handling and is emitted on any canceled MCP request, independent of userland `waitUntil` calls. It cannot be silenced from this package. This change fixes the real defect the report surfaced: teardown reliability. The warning itself is unchanged.

## Code Changes

Three cooperating pieces in `packages/agents/src/index.ts`:

- `Agent._cf_scheduleDestroy()` (new, internal): writes a durable "condemned" marker (`cf_agents_destroy_pending`, a KV record that survives `_dropInternalTablesForDestroy`) and arms an immediate alarm. Two fast storage writes. The MCP DELETE handler now awaits them before responding 204 instead of riding `waitUntil`. On a facet it falls through to `destroy()` directly, since facet teardown is a single root-coordinated `ctx.facets.delete`.
- `alarm()` preamble: checks for the pending-destroy marker before any other alarm work, including `super.alarm()`'s `onStart`. This prevents user `onStart` from running on a condemned agent against half-dropped tables (the pre-fix symptom: `no such table: cf_agents_schedules`). If the marker is present, it runs `destroy()` with its own full execution budget and returns.
- Convergence guards: `destroy()` writes the marker first, so a direct destroy that gets interrupted is finished by the next alarm wake. The final `deleteAll()` removes the marker, making "marker absent" the durable definition of "teardown complete". `_scheduleNextAlarm()` treats a pending destroy as scheduled work so it does not delete the destroy alarm as "no work pending".

`packages/agents/src/mcp/utils.ts`: the DELETE handler now calls `_cf_scheduleDestroy()` and awaits it before sending 204, instead of `ctx.waitUntil(agent.destroy())`.

`packages/agents/src/tests/deferred-destroy.test.ts` (new, 5 tests): scheduled teardown completes via alarm; a half-torn agent with marker present and some tables already dropped converges on next alarm instead of resuming schedules; `_scheduleNextAlarm` keeps the destroy alarm armed; direct destroy leaves no marker; end-to-end MCP Streamable-HTTP DELETE responds 204 and the session DO is wiped. The first three tests fail against the previous code, reproducing the zombie-resurrection symptom.

## Design tradeoffs

- Alarm vs marker-only: a marker alone converges only if the DO wakes again for some other reason. An abandoned MCP session (unique session id, never addressed again after DELETE) would keep its half-deleted storage forever. The alarm guarantees the wake. Cost: teardown spans two invocations and touches the alarm-scheduling path.
- The alarm check runs before `super.alarm()`/`onStart` on purpose: running user `onStart` on a condemned agent wastes work and risks failing against half-dropped tables.
- `destroy()` keeps its existing step order (`deleteAlarm` before `deleteAll`). The reverse would risk a storage-leaking zombie: an alarm firing after `deleteAll` would recreate empty tables with no marker left to condemn them. The residual unguarded window between `deleteAlarm` and `deleteAll` (two adjacent awaits inside a fresh alarm invocation) is accepted. Alarm retries cover handler failures before that point, and the original failure mode, a canceled request's budget, no longer applies to any part of the sequence.
- `_hasPendingDestroy()` reads the storage key directly rather than gating on `_isFacet`, because the in-memory facet flag is not yet hydrated at the `alarm()` preamble. Facets never write the marker, so the read is safe pre-initialization.

## Public API Surface

- New internal method: `Agent._cf_scheduleDestroy(): Promise<void>`. The `_cf_` prefix is consistent with the existing internal RPC surface. Unlike `destroy()`, it does not abort the isolate, so callers do not need to swallow an abort error.
- New internal storage record: `cf_agents_destroy_pending` (KV). Treated as reserved namespace, consistent with all other `cf_agents_*` records.
- No public API or type-level changes. `destroy()`'s observable contract is unchanged.

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1745" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/deferred-destroy-alarm-teardown.md
+++ b/.changeset/deferred-destroy-alarm-teardown.md
@@ -1,0 +1,11 @@
+---
+"agents": patch
+---
+
+Make agent teardown reliable when the initiating request is already canceled (#1625).
+
+The MCP Streamable-HTTP session-DELETE handler ran `agent.destroy()` via the request's `ctx.waitUntil`. By the time the DELETE lands the client is usually gone, the runtime gives a canceled request's trailing work little to no grace, and the multi-step teardown (drop tables, delete alarm, delete all storage, dispose connections) was routinely cut short — leaving half-deleted session DOs whose tables the constructor silently recreated on the next wake. (The associated `waitUntil() tasks did not complete` log warning itself originates inside workerd's WebSocket handling and is unaffected by this change.)
+
+Teardown is now deferred to the agent's own alarm invocation. The DELETE handler awaits two fast storage writes — a durable "condemned" marker plus an immediate alarm — and responds 204; the alarm then runs the real `destroy()` with a fresh execution budget. The marker is removed by the final `deleteAll()`, so it survives any interruption: `alarm()` checks it before any other work (including `onStart`) and finishes the teardown instead of resuming normal operation on a condemned agent, and `_scheduleNextAlarm()` keeps the destroy alarm armed rather than deleting it as "no work pending". `destroy()` itself now writes the marker first, so a direct destroy that gets interrupted converges the same way.
+
+New internal API: `Agent._cf_scheduleDestroy()` (used by the MCP handler; unlike `destroy()` it does not abort the isolate, so callers don't need to swallow an abort error). No public API or storage-schema changes; the marker is a single internal KV record (`cf_agents_destroy_pending`).
--- a/packages/agents/src/index.ts
+++ b/packages/agents/src/index.ts
@@ -938,7 +938,25 @@ const DEFAULT_AGENT_TOOL_RECOVERY_TOTAL_TIMEOUT_MS = 5_000;
 // with its own execution budget, #1625). The final `deleteAll()` in `destroy()`
 // removes it, so "marker present" always means an unfinished teardown that the
 // next wake must complete instead of resuming normal work.
+//
+// Scope: the marker is only consulted on alarm-driven paths (`alarm()` and
+// `_scheduleNextAlarm()`). It deliberately does NOT gate request entrypoints
+// (`onRequest`/`onMessage`/RPC) — a request that lands between scheduling and
+// the teardown alarm runs normally and `_ensureSchema()` recreates tables. For
+// the MCP session-DELETE use case this is benign: the session id is unique and
+// is never addressed again after DELETE, so no further request reaches a
+// condemned session DO before its teardown alarm fires.
 const DESTROY_PENDING_KEY = "cf_agents_destroy_pending";
+// Delay before the deferred-teardown alarm fires (#1625). `_cf_scheduleDestroy`
+// is awaited by an HTTP handler (the MCP session-DELETE) that then returns its
+// response. The teardown alarm runs `destroy()`, which ends in
+// `ctx.abort("destroyed")` — and an immediate (`Date.now()`) alarm fires and
+// aborts the isolate fast enough to race the still-in-flight RPC response,
+// surfacing to the caller as a 500 instead of the intended 204 (observed
+// against a real deployment; the local test runtime does not exhibit it). A
+// small delay lets the response flush before the abort. Teardown latency of a
+// second is irrelevant for an already-abandoned session.
+const DESTROY_ALARM_DELAY_MS = 1_000;
 // Ceiling for the exponential backoff applied to the runFiber-recovery
 // follow-up alarm. A scan that makes NO forward progress (every pending orphan
 // row's recovery hook threw) but still has work pending backs off so a poison
@@ -9513,15 +9531,29 @@ export class Agent<
    * callers don't need to swallow an abort error.
    */
   async _cf_scheduleDestroy(): Promise<void> {
+    // Hydrate facet state before deciding. `_isFacet` (and the `_parentPath`
+    // /`selfPath` the facet teardown path needs) is only populated by `onStart`
+    // /facet bootstrap, and `destroy()` below branches on the in-memory
+    // `_isFacet`. Without this, an RPC landing before init would see it as
+    // `false`, fall through to `destroy()`'s top-level path, and write the
+    // destroy marker on a facet — which the `alarm()`/`_scheduleNextAlarm()`
+    // guards forbid (only top-level agents write it; facet teardown is
+    // root-coordinated via `ctx.facets.delete`). Mirrors the other internal
+    // RPC entrypoints (`_workflow_*`). We must NOT push this into `destroy()`
+    // itself: the `alarm()` preamble calls `destroy()` precisely to avoid
+    // running `onStart` on a condemned agent.
+    await this.__unsafe_ensureInitialized();
     if (this._isFacet) {
-      // Facet teardown is coordinated by the root (`ctx.facets.delete`
-      // wipes the facet's storage in one step), not by a multi-step local
-      // sequence, so there is nothing to defer.
+      // Facet teardown is coordinated by the root (`ctx.facets.delete` wipes
+      // the facet's storage in one step), so there is nothing to defer.
       await this.destroy();
       return;
     }
     await this.ctx.storage.put(DESTROY_PENDING_KEY, true);
-    await this.ctx.storage.setAlarm(Date.now());
+    // Future, not immediate: see DESTROY_ALARM_DELAY_MS — an immediate alarm
+    // aborts the isolate fast enough to race this RPC's response back to the
+    // DELETE handler, turning the intended 204 into a 500.
+    await this.ctx.storage.setAlarm(Date.now() + DESTROY_ALARM_DELAY_MS);
   }
 
   /**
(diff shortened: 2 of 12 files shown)


threepointone:
Rebased on `main` and pushed a follow-up commit (`9077917e`) after reviewing the change and validating it against a **real Cloudflare deployment** — the local `vitest-pool-workers` runtime never cancels `waitUntil` or aborts the isolate under request timing, so the production behavior here isn't covered by unit tests.

### The important find: DELETE returned 500, not 204

`_cf_scheduleDestroy` armed the teardown alarm at `Date.now()`. In production that alarm fires and runs `destroy()` → `ctx.abort("destroyed")` fast enough to race the still-in-flight `_cf_scheduleDestroy` RPC response back to the awaiting DELETE handler, so the MCP Streamable-HTTP session-DELETE returned **`500` on every request** instead of the intended `204`. Teardown still converged, but the HTTP contract was broken (and this is a behavior change vs. the old `waitUntil` path, which returned 204 immediately).

Fix: defer the alarm by `DESTROY_ALARM_DELAY_MS` (1s) so the response flushes before the isolate aborts. Confirmed live against a deployed worker:

- before: `DELETE=500` ×6
- after: `DELETE=204` ×6, session DO still fully wiped in ~2s (counter reset to `initialState`, marker gone, no alarm, not initialized)

### Other hardening

- **Facet status resolved durably** in `_cf_scheduleDestroy` (new `_isFacetDurable()` helper) instead of the in-memory `_isFacet` flag, which is only hydrated in `onStart`. An RPC landing pre-init would otherwise read it as `false` and wrongly write the destroy marker on a facet — breaking the invariant the `alarm()` / `_scheduleNextAlarm()` guards depend on (only top-level agents write the marker; facet teardown is root-coordinated via `ctx.facets.delete`).
- **Marker scope documented**: it only gates alarm-driven paths, not request entrypoints. Benign for the session-DELETE case (unique session id, never re-addressed after DELETE).
- **New test**: `_scheduleNextAlarm` keeps the destroy alarm immediate even while a keepAlive lease is held (the branch most likely to push the teardown alarm out a full heartbeat). `deferred-destroy` suite is now 6/6.

### Verification harness

Added `wip/issue-1625-live/` — a deploy-based repro (real `McpAgent` + `scripts/repro.ts`) that initializes a Streamable-HTTP session, seeds a sentinel counter, DELETEs it (with optional `--abort` to drop the client mid-request, the disconnected-client condition behind #1625), and polls until the session DO is provably wiped vs. a zombie. This is what surfaced the 204/500 regression and confirms convergence on real infra. It's `wip/`, not part of CI.

`pnpm run check` (104 projects) and the agents MCP test files (26 files / 490 tests) pass. The disposers were also reviewed — `DisposableStore.dispose()` swallows per-item errors, so the only thing that can throw during teardown is a socket close in `mcp.dispose()`, which now converges via alarm retry rather than a lost `waitUntil`.

threepointone:
Follow-up (amended into `270886d8`): tightened the facet guard in `_cf_scheduleDestroy`.

The previous revision resolved facet status from durable storage (`_isFacetDurable()`) but then delegated to `destroy()`, which branches on the **in-memory** `_isFacet`. In the exact pre-`onStart` window that helper existed to protect, the two disagreed — the durable read said "facet" while `destroy()` saw `_isFacet === false` and took the top-level path, writing the `cf_agents_destroy_pending` marker on a facet (which the `alarm()`/`_scheduleNextAlarm()` guards forbid). It was also worse than just the flag: `destroy()`'s facet path needs hydrated `_parentPath`/`selfPath`, which durable detection alone wouldn't provide.

Fix: `_cf_scheduleDestroy` now calls `await this.__unsafe_ensureInitialized()` first (matching the other internal `_workflow_*` RPC entrypoints), then uses the same in-memory `_isFacet` as `destroy()`. The misleading `_isFacetDurable()` helper is removed. Deliberately **not** pushed into `destroy()` — the `alarm()` preamble calls `destroy()` precisely to avoid running `onStart` on a condemned agent.

This path isn't reachable today (`_cf_scheduleDestroy` is only called from the MCP DELETE handler on non-facet `streamable-http:*` session DOs), so it's a correctness-of-intent fix rather than a live bug. The top-level session-DELETE behavior is unchanged (the added init call is idempotent there) — 204 + teardown convergence still verified. `pnpm run check` (104 projects) and the `deferred-destroy` suite (6/6) pass.

PR #1747, opened 2026-06-12
fix(voice): gapless audio chunk playback

### LINKED ISSUE #1736: [voice] audible click at every audio chunk seam (w/ working fix)
VoiceClient starts each chunk at currentTime, then awaits its ended event before scheduling the next. That round-trip leaves a few ms of silence at every chunk boundary, which produces a periodic click during agent speech. Measured 5.33ms per seam streaming 320ms pcm16 chunks through a fake transport.

Expected: chunks scheduled back-to-back on the audio clock, no gaps.

Repro: stream any sustained pcm16 response and listen for clicks at chunk rate. A continuous tone makes it unmissable.

Version: 0.2.0

[Working fix](https://github.com/cloudflare/agents/commit/babf5bab38bdbb9d5b84b2072ef3c5da497af9dd) with cursor scheduling, interrupt stops the whole scheduled tail, tests + changeset, feel free to use.

cjol:
This PR fixes audible clicks at audio chunk boundaries during agent speech in `VoiceClient`. Closes #1736.

## Why

- `VoiceClient` played each response chunk by starting it at `currentTime` and waiting for its `ended` event before scheduling the next. The round-trip latency (event-loop delay plus the next chunk's setup) left a few milliseconds of silence at every seam, audible as a periodic click.
- We could have reduced the latency by overlapping decode and schedule, but the fundamental problem is that `ended` callbacks run on the main thread, not the audio clock, so there will always be some gap.
- Instead, we now schedule chunks back-to-back on the audio clock itself using a playback cursor. Consecutive chunks butt together sample-tight because each starts exactly when the previous ends, regardless of main-thread latency.

## Code Changes

- `packages/voice/src/voice-client.ts`
  - Replaced the single `#activeSource` with a `#scheduledSources` set and a `#playbackCursor`. Each incoming chunk is scheduled at `Math.max(currentTime, cursor)`, and the cursor is advanced by the chunk duration.
  - Playback now counts as active until the last scheduled source finishes, so barge-in detection (user transcript interrupt) keeps working through the scheduled tail.
  - `#stopPlayback` stops every scheduled source, not just the current one, so an interrupt or end-call cleanly cancels the whole pipeline.
  - Reset the cursor on end-call so a subsequent call starts fresh instead of picking up a stale offset.

- `packages/voice/src/tests/voice-client.test.ts`
  - Added gapless playback tests: consecutive chunk scheduling, catch-up when the cursor falls behind `currentTime`, interrupt stops all scheduled chunks, barge-in works through the scheduled tail, and cursor reset on end-call.

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1747" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


threepointone:
/bonk do a deep review, look for edge cases and things we might have missed, how's the ux/dx on this?

PR #1777, opened 2026-06-18
feat(think): Add retries to Think step.prompt()

thomasgauvin:
When I run step.prompt() as part of an agentic workflow, I want to have an option for the step.prompt() to be retried (in case the agent fails due to capacity constraints, etc).

I have kept this 'dumb', which will retry the whole step.prompt(). I think the Think harness itself should have retries or 'continues' to ensure the task is pushed to completion before stopping. 

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1777" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


thomasgauvin:
Remove from maxModelRetries from Think.submitMessages() because you can do it in beforeTurn()/getModel() (this sets the retry count for the whole agent)

thomasgauvin:
This cancel's Think built-in recovery, and we should make it so step.prompt() if it needs to recover, should resume Think and let it do its own chat recovery

thomasgauvin:
@threepointone would love to get your thoughts on this one. I've been working with workflows and step.prompt() would regularly die on deployments due to the underlying DOs dying. 

I've been testing this PR in my own setups for the past few days iterating to make it more reliable. It hooks into the Agent's built in chat recovery to avoid replaying previous turns (you can see that in previous iterations, I kept the retries 'dumb' but that was very expensive and my workflows would take forever to complete)

PR #1791, opened 2026-06-22
fix(codemode): keep optional AI peer out of root entry

### LINKED ISSUE #1779: [codemode] 0.4.x main entry statically imports the optional `ai` peer, breaking bundlers when `ai` is absent
**Describe the bug**

`@cloudflare/codemode@0.4.x`'s main entry (`dist/index.js`) has a **static, top-level** `import { tool } from "ai"`, while `ai` is declared as an **optional** peer dependency (`peerDependenciesMeta.ai.optional = true`). Because `DynamicWorkerExecutor` and `normalizeCode` are exported *only* from that same entry module, importing the core executor pulls in the static `ai` import. When `ai` is not installed, bundlers that honor the optional-peer contract (Vite + `@cloudflare/vite-plugin`, in our case) replace `ai` with an empty stub and the named import `tool` fails to resolve — breaking the build of code that never touches the AI-SDK surface. The executor is functionally independent of the Vercel AI SDK, so this looks like a packaging issue (a static import of an optional peer in the shared entry) rather than an intended hard requirement.

**To Reproduce**

1. In a Cloudflare Workers project built with Vite + `@cloudflare/vite-plugin`, do **not** install `ai`.
2. Import only the core executor: `import { DynamicWorkerExecutor, normalizeCode } from "@cloudflare/codemode";`
3. Run the build (`vite build`).
4. See the error:
   ```
   [MISSING_EXPORT] "tool" is not exported by "__vite-optional-peer-dep:ai:@cloudflare/codemode". 
   node_modules/@cloudflare/codemode/dist/index.js:6:10
   ```

**Expected behavior**

Importing `DynamicWorkerExecutor` / `normalizeCode` (the non-AI-SDK core) should not require `ai` to be installed, consistent with `ai` being declared an optional peer.

**Screenshots**

N/A — build-time bundler error, full text included under "To Reproduce".

**Version:**

`@cloudflare/codemode@0.4.1` (also reproduces on 0.4.0). Bundler: `vite@8` + `@cloudflare/vite-plugin@1.42` (Cloudflare Workers build).

**Additional context**

Root cause, in `dist/index.js`:

- line 6: `import { tool } from "ai";` (static, top-level)
- `tool` is only used in the connector/runtime helpers (`tool(...)` at ~L1712 and a `tool(options) { ... }` method at ~L1845), i.e. `createCodemodeRuntime` / `CodemodeConnector` — not in `DynamicWorkerExecutor` (defined ~L185) or `normalizeCode`.
- All of these are exported from the same `index.js` (~L1927), so the executor exports can't be reached without the `ai`-importing module being evaluated/linked.

Tree-shaking doesn't help: named-binding resolution against the optional-peer stub happens at link time, before unused-export elimination.

Suggested fix — make the `ai` dependency genuinely lazy so the optional peer is optional for bundlers:

- Use a dynamic `await import("ai")` inside the connector/runtime helpers that call `tool(...)`, **or**
- Move those AI-SDK-coupled helpers entirely behind the existing `./ai` subpath (which already imports `ai`) and keep `DynamicWorkerExecutor` / `normalizeCode` on a main entry that has no `ai` import.

Either keeps `import { DynamicWorkerExecutor } from "@cloudflare/codemode"` working without `ai` installed.

mattzcarey:
## Summary

- remove the root `@cloudflare/codemode` entry's runtime imports of the optional `ai` and `zod` peers
- return the runtime tool as the same plain AI SDK-compatible object that `tool()` already returned, with a dependency-free Standard Schema input validator
- fail the codemode build if the root JS or declaration entrypoint regains a framework-peer import
- cover `runtime.tool()` end to end through AI SDK `streamText`, including schema conversion, input validation, and execution
- add a patch changeset

Closes #1779.

## Testing

- packed `@cloudflare/codemode` and built an isolated Vite + `@cloudflare/vite-plugin` Worker fixture without `ai` installed
- `pnpm --filter @cloudflare/codemode run build`
- `pnpm --dir packages/codemode test` (297 unit/Workers, 38 runtime, 33 browser tests)
- `pnpm run build` (24 projects)
- `pnpm run check` (113 TypeScript projects)

### DIFF
--- a/packages/codemode/scripts/build.ts
+++ b/packages/codemode/scripts/build.ts
@@ -1,5 +1,27 @@
+import { readFileSync } from "node:fs";
 import { build } from "tsdown";
 import { formatDeclarationFiles } from "../../../scripts/format-declarations";
+
+function assertFrameworkFreeRootEntry() {
+  const optionalFrameworkPeers = ["ai", "zod", "@tanstack/ai"];
+
+  for (const file of ["dist/index.js", "dist/index.d.ts"]) {
+    const rootEntry = readFileSync(file, "utf8");
+    const frameworkImport = optionalFrameworkPeers.find(
+      (peer) =>
+        rootEntry.includes(`from "${peer}"`) ||
+        rootEntry.includes(`from '${peer}'`) ||
+        rootEntry.includes(`import("${peer}")`) ||
+        rootEntry.includes(`import('${peer}')`)
+    );
+
+    if (frameworkImport) {
+      throw new Error(
+        `${file} must not import optional framework peer "${frameworkImport}"`
+      );
+    }
+  }
+}
 
 async function main() {
   await build({
@@ -21,6 +43,8 @@ async function main() {
     sourcemap: true,
     fixedExtension: false
   });
+
+  assertFrameworkFreeRootEntry();
 
   // then run oxfmt on the generated .d.ts files
   formatDeclarationFiles();
--- a/packages/codemode/src/proxy-tool.ts
+++ b/packages/codemode/src/proxy-tool.ts
@@ -19,8 +19,6 @@
  * concurrently without clobbering one another.
  */
 import { RpcTarget } from "cloudflare:workers";
-import { tool, type Tool } from "ai";
-import { z } from "zod";
 import type { Executor, ResolvedProvider, ConnectorBinding } from "./executor";
 import { runCode } from "./run-code";
 import { normalizeCode } from "./normalize";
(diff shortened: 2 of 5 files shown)


packages/codemode/src/proxy-tool.ts
```diff
@@ -680,9 +735,16 @@ async function applyTransform(
...
-  options: CreateProxyToolOptions
-): Tool<ProxyToolInput, ProxyToolOutput> {
+export type CodemodeTool = {
+  description: string;
+  inputSchema: ProxyToolInputSchema;
+  execute: (
+    input: ProxyToolInput,
+    options: unknown
+  ) => Promise<ProxyToolOutput>;
+};
```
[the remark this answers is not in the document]
mattzcarey: Added an end-to-end `streamText` regression test using AI SDK’s `MockLanguageModelV3`. It emits a codemode tool call and asserts the plain runtime tool passes schema conversion/input validation, executes, and returns the expected tool result. Codemode tests and the full repository check pass.

packages/codemode/scripts/build.ts
```diff
@@ -1,6 +1,28 @@
+import { readFileSync } from "node:fs";
 import { build } from "tsdown";
 import { formatDeclarationFiles } from "../../../scripts/format-declarations";
 
+function assertFrameworkFreeRootEntry() {
```
threepointone, line 5: this seems like overkill

threepointone (approved):
approving with a note

packages/codemode/scripts/build.ts
mattzcarey: yeah fair

mattzcarey: it was just the 3rd time ive caused this bug haha

PR #1796, opened 2026-06-22
fix(agents): make AI SDK peer optional

### LINKED ISSUE #751: Deploy to Cloudflare buttons are failing on demos using v0.3
ai is now a peer dep of agents. Some of these demos are linked in docs which have a Deploy to Cloudflare button which now errors

```bash
wrangler 4.56.0

✘ [ERROR] Build failed with 1 error:
  ✘ [ERROR] Could not resolve "ai"
  
      node_modules/agents/dist/client-QZa2Rq0l.js:945:39:
        945 │       const { jsonSchema } = await import("ai");

    You can mark the path "ai" as external to exclude it from the bundle, which will remove this error and leave the unresolved path in the bundle. You can also add ".catch()" here to handle this failure at run-time instead of bundle-time.
```

Raised in https://github.com/cloudflare/ai/pull/345#issuecomment-3688190316

**Version:**
0.3.0

mattzcarey:
## Summary

- mark the `ai` peer optional again
- remove AI SDK types from the emitted `agents` root declaration graph while preserving structural `ToolSet` compatibility for MCP tools
- make agent-tool streamed parts generic so AI SDK callers can opt into `UIMessage["parts"][number]` without forcing that type on every `agents` consumer

## Why

`ai` became required in `agents@0.3.1` to fix #751. At the time, every root `Agent` constructed `MCPClientManager`, whose module contained a literal `import("ai")` for `jsonSchema()`. Bundlers resolved that dynamic import even when applications never called `getAITools()`.

That runtime dependency is gone: MCP schema conversion now uses Zod, and the old chat implementation moved to `@cloudflare/ai-chat`. The remaining root coupling was type-only: `MCPClientManager.getAITools()` returned AI SDK `ToolSet`, and `AgentToolRunState.parts` referenced `UIMessage["parts"]`.

This removes those final declaration references and restores the optional peer. AI-specific entry points still require `ai` when imported.

## Compatibility

`MCPAIToolSet` is structurally assignable to AI SDK `ToolSet`. `AgentToolRunState` and `useAgentToolEvents` now accept an optional part generic; AI SDK consumers can use `UIMessage["parts"][number]` for the previous exact union.

## Validation

- `pnpm run build`: 24 projects pass
- `pnpm run check`: format, lint, exports, and all 113 TypeScript projects pass
- `pnpm --filter agents test:chat`: 477 tests pass
- focused MCP client manager: 149 tests pass
- focused agent-tool reducer: 16 tests pass
- full Workers suite: 1,456 pass / 7 skip; five unrelated Python skill tests timed out fetching the external Pyodide bundle

### DIFF
--- a/packages/agents/scripts/build.ts
+++ b/packages/agents/scripts/build.ts
@@ -1,7 +1,6 @@
 import { build } from "tsdown";
 import { globSync } from "glob";
 import { existsSync, readFileSync, writeFileSync } from "node:fs";
-import { dirname, resolve } from "node:path";
 import { formatDeclarationFiles } from "../../../scripts/format-declarations";
 
 const entries = [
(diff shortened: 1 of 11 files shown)


packages/agents/scripts/build.ts
```diff
@@ -54,6 +55,66 @@ function injectSkillsTypeReference(): void {
   }
 }
 
+function moduleSpecifiers(source: string): string[] {
```
threepointone: this is all overkill, we don't need this here

mattzcarey: removed

packages/agents/src/tests-d/optional-ai-peer.test-d.ts
```diff
@@ -0,0 +1,21 @@
+import type { ToolSet, UIMessage } from "ai";
```
threepointone: use `satisfies` to verify all our new types match ai sdk's types

threepointone (approved):
approving but please add the type tests

packages/agents/src/tests-d/optional-ai-peer.test-d.ts
mattzcarey: Added in 97a160b7: the optional-ai-peer declaration test now uses `satisfies` for `MCPAITool -> Tool`, `MCPAIToolSet -> ToolSet`, and the `UIMessage`-specialized `AgentToolRunState` / `useAgentToolEvents` parts. Verified with `pnpm exec tsx scripts/typecheck.ts packages/agents` and full `pnpm run check`.

PR #1800, opened 2026-06-23
RFC: Agent Auth authority plane

mattzcarey:
## Summary

Proposes a minimal authority plane for Agents covering three concerns:

- durable Agent-local OAuth grants and API keys
- mutable policy and requestable denials shared by MCP, local AI SDK tools, codemode, and domain libraries
- a minimal Agent/user principal envelope for attribution and policy evaluation

The RFC introduces a native `Agent.auth` manager, a canonical `AuthorityOperation` / `authorize()` contract, durable idempotent authorization requests, typed auth descriptors for locally authored tools, and a framework-neutral domain gate for libraries such as Workspace.

## Key design choices

- Grants belong to one Agent instance; there is no shared user credential vault.
- Remote MCP tools remain server-defined and keep standard MCP/OAuth protocol ownership.
- Local `authenticatedTool()` values remain ordinary AI SDK tools.
- MCP annotations provide a common risk vocabulary but do not impose SDK policy defaults.
- An `ask` policy is implemented as a requestable denial; approval triggers fresh policy evaluation rather than becoming a standing grant.
- Codemode keeps durable replay ownership while authorization requests and policy move to the shared kernel.
- Optional history-aware policy tracks conservative execution/thread labels and prior-operation lineage so known private sources can block or step up known public sinks. This is not claimed as complete DLP or field-level taint tracking.
- Intent-derived policy is positioned as an optional app/PDP layer: compile user task intent into structured transient constraints, then enforce them through the same AuthorityOperation checks.
- Portable Agent identity is deferred: a future opt-in profile may bind an Agent Ed25519 key to RFC 9421 HTTP Message Signatures, with key registration, replay defense, rotation, and revocation.
- Domain libraries normalize their own actions/resources and gate the last controllable effect boundary without depending on Agents.
- Typed auth descriptors replace fragile string-keyed auth profiles and normalize into semantic IDs usable by optional central credential and mandatory policy providers.
- OAuth follows RFC 9728, RFC 8414, PKCE, RFC 8707, and RFC 7591 where supported.

## Migration

The RFC includes an ownership table and phased migration for:

1. introducing the kernel without behavior changes
2. local AI SDK tools
3. MCP credential-storage adaptation and call-time policy
4. codemode replay-safe authorization
5. shared durable approval presentation
6. domain libraries, starting with Workspace
7. history-aware information context
8. audit hardening
9. opt-in Agent identity and centralized providers

MCP discovery, transports, OAuth challenges, callback routing, and SDK interfaces stay in MCP. Codemode replay, result logging, pause/resume, and rollback stay in codemode. The shared layer owns Agent-local grants, policy, authorization requests, and audit.

## Validation

- `npx oxfmt --check design/rfc-agent-auth.md`
- `git diff --check`

Design-only change; no runtime code changed.


JosephDoUrden:
really like the direction here, one authorize() over a single AuthorityOperation is the right shape and the requestable-denial thing (ask isnt a grant) is nice. two things i hit reading it against whats on main, both from the mcp client side.

first one, x402 isnt mentioned anywhere in the rfc and i think it dents the "does not introduce a second safeguard field" line (L547). the payment path in packages/agents/src/mcp/client/x402.ts already wraps callTool (it defineProperty's over it) and it fires on a server-controlled field, res._meta has an x402/error with accepts and it triggers the payment branch. so a server can raise it on a call the policy allowed as readOnly. and it fails open, if no confirmationCallback is configured the payment just proceeds, the only bound left is maxPaymentValue which defaults to 0.10 usdc and lives in client config not auth.policy. then it retries the call with _meta["x402/payment"] set, so one model-visible tool call turns into two dispatches and a signed payment. and none of that is behind MCPClientManager.callTool, withX402Client wraps the raw client and the manager has no x402 in it, so the boundary you call canonical in L987 isnt in that path at all. might be worth saying where _meta sits in the digest too, L484 lists parameters and policy-relevant context, L153 calls those the validated normalized parameters and L981 binds the effect after arguments validate, so on the plain reading the paid retry digests the same as the unpaid one and rides its approval. either answer looks bad so probably worth pinning down in the mcp adapter normalization. feels like payment wants to be a first-class op, something like domain payment / action x402.settle, authorized before the payload is signed, with the max value as a policy rule instead of a client config option.

second, AgentGrant (L224) has no principal on it. AuthorityEvent does carry user issuer+subject (L1139) right next to grantId (L1153), so the audit can name a user that the grant it used was never bound to. agents get addressed by an arbitrary name in the route and onBeforeConnect is opt-in, so nothing in the platform forces one agent per user. take a shared support or workspace agent, user A does the oauth, grant lands in that DO, user B connects to the same DO and calls the same tool, resolution finds A's grant by resource + agent identity and dispatches with A's token. thats the same shape as state handle hijacking in the mcp security best practices, where the fix is binding stored state to the authenticated user and rejecting it when another principal presents it. and i dont think its the thing you deferred to rfc 9421, that ones about proving agent identity externally, this is just binding a locally held grant to the locally verified user, which you already have the primitive for. maybe AgentGrant carries a principal and resolution refuses a grant whose principal != operation.principal.user, or grants get keyed by (agentId, userPrincipal).

havent run any of this, theres no impl in the tree yet so both are from reading the rfc against main, not a fixture. can dig into the resources/read + prompts/get paths and the always_allow selector if thats useful, didnt want to dump everything at once.

PR #1820, opened 2026-06-27
test: cover recovery after forced Durable Object eviction

mattzcarey:
## Summary

Test-only coverage for recovery after a running Durable Object is explicitly evicted with `evictDurableObject()` / `evictAllDurableObjects()`.

- verifies volatile instance state is dropped while SQLite and Durable Object storage are retained
- covers Agent state, schedules and fibers; Session active-leaf reconstruction; MCP startup/session restoration; AIChat history, streams and WebSockets; Codemode durable replay; Workspace; Think startup/recovery; and Voice history
- keeps a live AIChat WebSocket usable across forced eviction and separately covers `{ webSockets: "close" }`
- adds no runtime code or changeset

## Scope

This proves selected reconstruction paths after **forced eviction**. It does not prove natural idle hibernation, timer absence, hibernation eligibility, or duration-billing suspension.

Related to #1956; does not fix it. The test helper bypasses the eligibility gate, so a passive workerd/vitest-pool API is still needed to detect timer-pinned actors without waking or forcibly evicting them.

## Verification

- `pnpm run build`
- `pnpm run check`
- `CI=true pnpm exec nx affected -t test --base=origin/main --parallel=3`

### DIFF
--- a/packages/agents/src/tests/experimental/memory/session/provider.test.ts
+++ b/packages/agents/src/tests/experimental/memory/session/provider.test.ts
@@ -1,5 +1,6 @@
 import type { UIMessage } from "ai";
 import { env } from "cloudflare:workers";
+import { evictDurableObject } from "cloudflare:test";
 import { describe, expect, it, beforeEach } from "vitest";
 import { getAgentByName } from "../../../..";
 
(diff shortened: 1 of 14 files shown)


packages/agents/src/tests/experimental/memory/session/provider.test.ts
```diff
@@ -388,6 +389,42 @@ describe("AgentSessionProvider — tree-structured messages", () => {
...
+    });
+    await agent.appendMessage({
+      id: "m2",
+      role: "assistant",
+      parts: [{ type: "text", text: "Hi" }]
+    });
+
+    await evictDurableObject(agent as unknown as DurableObjectStub);
+
+    const history = await agent.getHistory();
```
threepointone: hmm if we did an eviction, the stub shouldn't work after that I think? have to confirm with workerd team

mattzcarey: weird that these tests pass then

mattzcarey: ## Why the test passes — the stub *is* supposed to work after eviction

`evictDurableObject(stub)` → `workerdUnsafe.evict(stub)` → `channel->evictForTest()` in workerd. The source says it outright:

**`workerd/src/workerd/api/unsafe.h:118-122`**

> *"gracefully evict the Durable Object referred to by `stub` from its isolate, simulating the runtime tearing it down when it goes idle. Durable storage is left intact, so the DO rebuilds (rerunning its constructor) on its next request."*

**`workerd/src/workerd/io/io-channels.h:267-274`**

> *"forcibly evict the target … this destroys the actor instance while durable storage survives, so the next request rebuilds it. … Throws if the target Durable Object is not currently running."*

So threepointone's intuition is incorrect (and this is exactly the workerd-team confirmation he asked for). A `DurableObjectStub` is an ID-addressed routing handle, not a pointer to the live instance. Eviction destroys the in-isolate actor but keeps storage; the next call through the same stub reruns the constructor and rehydrates from SQLite. The stub continuing to work is the production hibernation lifecycle — that's the whole point of the helper.

For contrast, the three sibling APIs differ precisely here:

- `evict()` / `evictAllDurableObjects()` — graceful: storage intact, stub keeps working, hibernates/closes WebSockets.
- `abortAllDurableObjects()` — throws an exception into the actor (this is the one that breaks in-flight work).
- `deleteAllDurableObjects()` — abort and wipes storage so the DO restarts clean.

One nice corollary: `evict()` rejects if the DO isn't currently running. In `provider.test.ts` the agent was just written to (`appendMessage`), so it's live → the eviction definitely happened, otherwise `evict()` itself would have thrown and failed the test.

PR #1862, opened 2026-07-02
fix: addMcpServer returns READY for a restored connection stuck AUTHENTICATING

### LINKED ISSUE #1855: addMcpServer returns READY for a restored connection stuck AUTHENTICATING (in-memory authUrl never rehydrated on wake)
## Summary

`Agent.addMcpServer()` returns `{ state: "ready" }` for an existing connection that is actually stuck in `AUTHENTICATING` — whenever that connection has no **in-memory** `authProvider.authUrl`, which is exactly the state after any Durable Object wake (hibernation / eviction / redeploy). A caller that drives OAuth off the return value therefore believes the server is connected and never re-surfaces a sign-in link, so the flow wedges silently.

Observed in `agents@0.17.1` (please confirm against `main`).

## Mechanism

The existing-connection early-return in `addMcpServer` only reports `authenticating` when the provider still holds a live `authUrl`:

```js
if (existingServer && this.mcp.mcpConnections[existingServer.id]) {
  const conn = this.mcp.mcpConnections[existingServer.id];
  if (conn.connectionState === MCPConnectionState.AUTHENTICATING && conn.options.transport.authProvider?.authUrl)
    return { id, state: MCPConnectionState.AUTHENTICATING, authUrl: conn.options.transport.authProvider.authUrl };
  if (conn.connectionState === MCPConnectionState.FAILED) throw ...;
  return { id, state: MCPConnectionState.READY };   // ← AUTHENTICATING-without-authUrl falls through to here
}
```

The provider's `_authUrl_` is assigned in exactly one place — `DurableObjectOAuthClientProvider.redirectToAuthorization()` (i.e. only during a live authorize dance). It is **never rehydrated** when connections are restored from storage on wake. So a connection that was left `AUTHENTICATING` (user minted a link but hasn't completed consent) comes back after a wake with `connectionState === AUTHENTICATING` but `authProvider.authUrl === undefined`, and the early-return reports it `READY`.

## Impact

Any harness that decides "do I need to prompt the user to sign in?" from `addMcpServer`'s return value gets a false READY on every wake after an unconsumed sign-in link, and stops prompting. In our case the per-user OAuth nudge concluded "already connected" and never re-sent the link; the connection never actually reached ready.

## Repro sketch

1. `addMcpServer(id, url, …)` on an OAuth server → `{ state: "authenticating", authUrl }`. Don't complete consent.
2. Evict / wake the DO (or just let it hibernate), so connections restore from storage.
3. `addMcpServer(id, url, …)` again → returns `{ state: "ready" }`, but `getMcpServers().servers[id].state` is still `authenticating`.

`getMcpServers()` (which reads persisted state) and `addMcpServer()`'s return disagree for the same connection in the same tick.

## Suggested fix

On the existing-connection path, treat `AUTHENTICATING` as authenticating regardless of whether `authUrl` is in memory — re-derive/re-mint the `authUrl` (or return a state that tells the caller to re-authorize) rather than falling through to `READY`. At minimum, don't report `READY` for a connection whose `connectionState` is `AUTHENTICATING`.

## Workaround

Trust `getMcpServers().servers[id].state` over the return value; on disagreement, `removeMcpServer(id)` + re-add to mint a fresh flow.

Related but distinct: #942 (restore bypasses the `createMcpOAuthProvider` override) — same "restore path doesn't reconstruct provider state" family.

mattzcarey:
Fixes #1855

## Problem

`addMcpServer()` returns `{ state: "ready" }` for an existing connection that is actually stuck in `AUTHENTICATING`, whenever that connection has no in-memory `authProvider.authUrl` — which is exactly the state after any Durable Object wake (hibernation / eviction / redeploy). A caller that drives OAuth off the return value believes the server is connected and never re-surfaces the sign-in link, so the flow wedges silently. Meanwhile `getMcpServers()` (which reads persisted state) reports `authenticating` for the same server in the same tick.

## Root cause

1. When a connection enters `AUTHENTICATING`, `connectToServer()` persists `auth_url` to the `cf_agents_mcp_servers` table.
2. On wake, `restoreConnectionsFromStorage()` sees the stored `auth_url`, recreates the connection, and sets `connectionState = AUTHENTICATING` — but the provider's in-memory `authUrl` is only ever assigned inside `redirectToAuthorization()` during a live authorize dance, so it comes back `undefined`.
3. `addMcpServer`'s existing-connection early return required `connectionState === AUTHENTICATING && authProvider?.authUrl` to report `authenticating`; the restored connection failed the second condition and fell through to `READY`.

## Fix

- For an existing connection in `AUTHENTICATING`, fall back to the persisted `auth_url` (already in scope from `listServers()`) when the in-memory URL is missing, so `addMcpServer`'s return value agrees with `getMcpServers()`.
- If no auth URL exists in memory **or** storage, fall through to re-run the connect flow and mint a fresh one — never report `READY` for a connection whose state is `AUTHENTICATING`. (`MCPClientConnection.init()` is explicitly re-enterable after a 401 → OAuth cycle, and if stored tokens turn out valid this self-heals to `READY` through the normal discover path.)
- Supporting change for that fall-through: the HTTP path now reuses the existing server's id (`requestedId ?? existingServer?.id ?? nanoid(8)`), matching what the RPC path already did, so re-adding a known server never orphans its storage row under a freshly generated id.

## Tests (TDD)

- `returns authenticating with the persisted authUrl for a restored connection awaiting OAuth` — reproduces the issue's exact scenario (persisted `auth_url` row → wake → duplicate `addMcpServer`); failed with `"ready"` before the fix. Also pins the invariant that `addMcpServer`'s return state agrees with `getMcpServers()` in the same tick.
- `prefers the live in-memory authUrl during an in-flight OAuth flow` — pins the existing fast path (in-memory URL wins over persisted state).

`packages/agents` workers suite: 1510 tests pass; repo `npm run check` (sherif, exports, oxfmt, oxlint, typecheck) green.

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1862" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/spotty-pandas-refuse.md
+++ b/.changeset/spotty-pandas-refuse.md
@@ -1,0 +1,9 @@
+---
+"agents": patch
+---
+
+fix: `addMcpServer` no longer reports `ready` for a restored connection that is still awaiting OAuth.
+
+A connection restored after a Durable Object wake (hibernation, eviction, redeploy) keeps its persisted `AUTHENTICATING` state, but the OAuth provider's in-memory `authUrl` was never rehydrated. The existing-connection early return in `addMcpServer` required that in-memory URL to report `authenticating`, so it fell through to `ready` — while `getMcpServers()` correctly reported `authenticating` for the same server. Callers driving OAuth off the return value stopped re-surfacing the sign-in link and the flow wedged silently.
+
+`addMcpServer` now falls back to the persisted `auth_url` for authenticating connections, so its return value matches `getMcpServers()`. If no auth URL exists in memory or storage, it re-runs the connect flow to mint a fresh one instead of reporting `ready`. Re-adding a known server also reuses its existing id on the HTTP path (as the RPC path already did) instead of generating a new one.
--- a/packages/agents/src/index.ts
+++ b/packages/agents/src/index.ts
@@ -12259,22 +12259,30 @@ export class Agent<
 
     if (existingServer && this.mcp.mcpConnections[existingServer.id]) {
       const conn = this.mcp.mcpConnections[existingServer.id];
-      if (
-        conn.connectionState === MCPConnectionState.AUTHENTICATING &&
-        conn.options.transport.authProvider?.authUrl
-      ) {
-        return {
-          id: existingServer.id,
-          state: MCPConnectionState.AUTHENTICATING,
-          authUrl: conn.options.transport.authProvider.authUrl
-        };
-      }
-      if (conn.connectionState === MCPConnectionState.FAILED) {
+      if (conn.connectionState === MCPConnectionState.AUTHENTICATING) {
+        // Prefer the in-memory authUrl from a live OAuth flow; fall back to
+        // the persisted auth_url for a connection restored after hibernation,
+        // where the provider's in-memory URL is lost.
+        const authUrl =
+          conn.options.transport.authProvider?.authUrl ??
+          existingServer.auth_url;
+        if (authUrl) {
+          return {
+            id: existingServer.id,
+            state: MCPConnectionState.AUTHENTICATING,
+            authUrl
+          };
+        }
+        // No auth URL in memory or storage: fall through to re-run the
+        // connect flow and mint a fresh one, rather than reporting READY
+        // for an unauthenticated server.
+      } else if (conn.connectionState === MCPConnectionState.FAILED) {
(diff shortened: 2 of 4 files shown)


mattzcarey:
closing this. want another shot

PR #1881, opened 2026-07-06
Minimal python package support

abstractedfox:
This implements the beginning of package support for dynamic python workers. At present, it allows defining a flat list of pure Python packages as dependencies, which it installs in the dynamic worker without dependency resolution


packages/worker-bundler/src/index.ts
```diff
@@ -143,6 +143,29 @@ export async function createWorker(
     );
   }
 
+  if (entryPoint.endsWith(".py")) {
+    const modules: Modules = {};
+    for (const path of fileSystem.list()) {
+      if (path !== "pyproject.toml") {
```
dom96: Nit:
```suggestion
      if (path === "pyproject.toml") { continue; }
```

packages/worker-bundler/src/index.ts
```diff
@@ -143,6 +143,29 @@ export async function createWorker(
...
+        const content = fileSystem.read(path);
+        if (content !== null) {
+          modules[path] = content;
+        }
+      }
+    }
```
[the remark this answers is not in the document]
dom96: These are just files specified by the user in the createWorker call, right? So this AI review comment seems wrong.

packages/worker-bundler/src/installer.ts
```diff
@@ -106,48 +118,100 @@ export async function installDependencies(
...
+    const depsToInstall: Record<string, string> = {
+      ...packageJson.dependencies,
+      ...(dev ? packageJson.devDependencies : {})
+    };
 
-  if (Object.keys(depsToInstall).length === 0) {
-    return result; // No dependencies to install
-  }
+    if (Object.keys(depsToInstall).length === 0) {
+      return result; // No dependencies to install
```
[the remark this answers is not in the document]
dom96: This seems to be fine. Python Workers shouldn't specify package.json.

packages/worker-bundler/src/installer.ts
```diff
@@ -106,48 +118,100 @@ export async function installDependencies(
...
+  const pyprojectTomlContent = fileSystem.read("pyproject.toml");
 
-  let packageJson: PackageJson;
-  try {
-    packageJson = JSON.parse(packageJsonContent) as PackageJson;
-  } catch {
-    result.warnings.push("Failed to parse package.json");
-    return result;
-  }
+  if (packageJsonContent) {
```
dom96, line 128: Code style suggestion: instead of pushing this into its own if branch, create a new `installDependenciesPython` function and move the Python logic there, then call this function when pyprojectTomlContent is non-null.

This will make it easier to review this too.

packages/worker-bundler/src/installer.ts
```diff
@@ -106,48 +118,100 @@ export async function installDependencies(
...
+      if (!trimmed) continue;
+
+      //once base functionality is working, test this against more elaborate version strings
+      const match = trimmed.match(/^([A-Za-z0-9._-]+)(.*)$/);
+      if (!match) {
+        result.warnings.push(
+          `Skipping invalid pyproject.toml dependency: ${dep}`
+        );
+        continue;
+      }
```
dom96: I wouldn't bother parsing these. For this PR you should only support dependencies without version requirements being specified. In the future this is something that the package resolver would handle the parsing for anyway, so no need to worry about it here.

So remove this code and just accept the dep names as they are.

packages/worker-bundler/src/installer.ts
```diff
@@ -106,48 +118,100 @@ export async function installDependencies(
...
+    await Promise.all(
+      Object.entries(depsToInstall).map(([depName, depVersion]) =>
+        installPythonPackage(
+          depName,
+          depVersion,
+          result,
+          fileSystem,
+          installedPackages,
+          inProgress,
+          PYPI_JSON_API //hardcoding this for now to keep the implementation light
```
dom96: Nit: space after `//` (same for rest of comments)

packages/worker-bundler/src/installer.ts
```diff
@@ -250,6 +314,148 @@ async function installPackage(
...
+/**
+ * Install a single Python package from PyPI.
+ *
+ * This is a minimal implementation: it downloads the latest version of the
+ * package as a source distribution and adds it to python_modules/. It does not
+ * resolve version ranges or install transitive dependencies.
+ */
+async function installPythonPackage(
+  name: string,
+  _versionRange: string,
```
dom96:
```suggestion
```

packages/worker-bundler/src/installer.ts
```diff
@@ -250,6 +314,148 @@ async function installPackage(
...
+  inProgress: Map<string, Promise<void>>,
+  registry: string
+): Promise<void> {
+  // Skip if already installed in this run
+  if (installedPackages.has(name)) {
+    return;
+  }
+
+  // Skip if the package already exists in the filesystem
+  if (fileSystem.read(`python_modules/${name}/__init__.py`) !== null) {
```
dom96: could we rely on just this check and remove the `installedPackages` tracking?

packages/worker-bundler/src/installer.ts
```diff
@@ -250,6 +314,148 @@ async function installPackage(
...
+  // If installation is already in progress, wait for it
+  const existing = inProgress.get(name);
+  if (existing) {
+    return existing;
+  }
+
+  const installPromise = (async () => {
+    try {
+      // The JS impl. does the metadata part in another function, consider moving it if this gets too long
+      // Fetch package metadata from PyPI JSON API
```
dom96: Yeah, might be worth splitting this up

packages/worker-bundler/src/installer.ts
```diff
@@ -250,6 +314,148 @@ async function installPackage(
...
+      );
+
+      // Mark as installed before writing to prevent cycles
+      installedPackages.set(name, version);
+      result.installed.push(`${name}@${version}`);
+
+      // Add files to python_modules
+      for (const [filePath, content] of Object.entries(packageFilesWheel)) {
+        fileSystem.write(`python_modules/${name}/${filePath}`, content);
+      }
```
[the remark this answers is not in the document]
dom96: Yeah, that's a fair point. Does the metadata help here to determine where to install this?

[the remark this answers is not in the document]
dom96: Not a big deal in this case

packages/worker-bundler/src/tests/e2e.test.ts
```diff
@@ -984,3 +984,116 @@ describe("hasNodejsCompat", () => {
...
+    });
+    const id = "test-worker-" + testId++;
+
+    const worker = env.LOADER.get(id, () => ({
+      mainModule: dynamic_worker.mainModule,
+      modules: dynamic_worker.modules,
+      compatibilityDate:
+        dynamic_worker.wranglerConfig?.compatibilityDate ?? "2026-01-01",
+      compatibilityFlags: dynamic_worker.wranglerConfig?.compatibilityFlags ?? [
+        "python_workers"
```
dom96: Should we expect createWorker to set this and therefore assert that these are non-null?

packages/worker-bundler/src/tests/e2e.test.ts
```diff
@@ -984,3 +984,116 @@ describe("hasNodejsCompat", () => {
...
+        ].join("\n"),
+        "pyproject.toml": [
+          "[project]",
+          'name = "dummy"',
+          'version = "0.0.0"',
+          // typing_extensions has zero dependencies. typing_inspection depends on typing_extensions
+          'dependencies = ["typing_extensions", "typing_inspection"]'
+        ].join("\n")
+      }
+    });
```
dom96, line 1087: Nice! Exactly what I was looking for :)

dom96 (commented):
Overall approach looks good! Left a few comments with suggestions.

CC @hoodmane @ryanking13 @threepointone for reviews too

packages/worker-bundler/src/index.ts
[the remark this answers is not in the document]
abstractedfox: At that point it should be those files + all the packages that were just added, so my assumption is anything that's in there would've been put there deliberately

packages/worker-bundler/src/installer.ts
abstractedfox: I borrowed this pattern from the JS implementation ~~where I assumed it was done to prevent races, my assumption is that since you can't predict when `package.json/__init__.py` will get put into the file system you could feasibly end up with it happening twice if two packages with the same dependency got fetched around the same time (or maybe that anticipates more concurrency from JS than is possible?).~~ If it bares removing maybe I should change it in the JS one too?
(upon actually reviewing it again, I've realized the comment also just says it's intended to save fetches if called on a pre-warmed filesystem that already has that dep, which makes more sense)

packages/worker-bundler/src/tests/e2e.test.ts
abstractedfox: That's probably a good idea

packages/worker-bundler/src/tests/wrangler.jsonc
```diff
@@ -1,5 +1,5 @@
 {
-  "compatibility_date": "2026-06-11",
+  "compatibility_date": "2026-05-15",
```
ryanking13: Why is this changed? (no need rollback, just curious).

packages/worker-bundler/src/index.ts
```diff
@@ -143,6 +143,30 @@ export async function createWorker(
...
+      if (content !== null) {
+        modules[path] = content;
+      }
+    }
+
+    return {
+      mainModule: entryPoint,
+      modules,
+      wranglerConfig: {
+        compatibilityDate: wranglerConfig?.compatibilityDate ?? "2026-01-01",
```
ryanking13: How is this default compat date selected? How does the JS worker handle this?

packages/worker-bundler/src/index.ts
```diff
@@ -143,6 +143,30 @@ export async function createWorker(
...
+        compatibilityFlags: wranglerConfig?.compatibilityFlags ?? [
```
ryanking13: We will always need to add `python_workers` flag to make the worker work. So maybe we need to inject one if this flag does not exist?

packages/worker-bundler/src/installer.ts
```diff
@@ -6,9 +6,12 @@
  */
 
 import * as semver from "semver";
+import { unzipSync } from "fflate";
```
ryanking13, line 9: We do have CompressionStream support in CF worker. [fflate](https://github.com/101arrowz/fflate#what-about-compressionstream) claims that the performance of CompressionSTream is bad in the browser, but that might not be true.

you don't need to use CompressionStream in this PR as it is an implementation detail that we can change anytime and the performance is not our concern at this point, but it would be nice to add a comment in the code why you decided to use this package and things that need to be checked (performance compared to the CompressionStream) later.

packages/worker-bundler/src/installer.ts
```diff
@@ -106,26 +118,81 @@ export async function installDependencies(
...
+          installedPackages,
+          inProgress,
+          registry
+        )
+      )
+    );
   }
 
-  let packageJson: PackageJson;
+  if (pyprojectTomlContent) {
```
ryanking13: nitpick: maybe use `else if` to make sure we don't install dependencies for js and python both in some weird case?

packages/worker-bundler/src/installer.ts
```diff
@@ -106,26 +118,81 @@ export async function installDependencies(
...
+  for (const dep of pyprojectToml.project?.dependencies ?? []) {
+    const name = dep.trim();
+    if (!name) continue;
+
+    depsToInstall[name] = "*"; // in the future this should be a version specifier, if one was set
+  }
 
   if (Object.keys(depsToInstall).length === 0) {
-    return result; // No dependencies to install
+    return; // No dependencies to install
```
ryanking13: We always install `workers-runtime-sdk` into the python worker ([code pointer](https://github.com/cloudflare/workers-py/blob/63e2eb2e71bf0e29968e0e85708cc7304b91fce8/packages/cli/src/pywrangler/resolve.py#L121))

packages/worker-bundler/src/installer.ts
```diff
@@ -135,20 +202,17 @@ export async function installDependencies(
...
-        versionRange,
+    Object.entries(depsToInstall).map(([depName]) =>
+      installPythonPackage(
+        depName,
         result,
         fileSystem,
         installedPackages,
         inProgress,
-        registry
+        PYPI_JSON_API // hardcoding this for now to keep the implementation light
```
ryanking13, line 216: you'll also need to pass [`index.pyodide.org/<version>`](https://github.com/cloudflare/workers-py/blob/63e2eb2e71bf0e29968e0e85708cc7304b91fce8/packages/cli/src/pywrangler/utils.py#L420) to install packages that are not available in the PyPI yet.

packages/worker-bundler/src/installer.ts
```diff
@@ -250,6 +314,116 @@ async function installPackage(
...
+  inProgress: Map<string, Promise<void>>,
+  registry: string
+): Promise<void> {
+  // Skip if already installed in this run
+  if (installedPackages.has(name)) {
+    return;
+  }
+
+  // Skip if the package already exists in the filesystem
+  if (fileSystem.read(`python_modules/${name}/__init__.py`) !== null) {
```
ryanking13: When will the package be already installed in the filesystem? We should be installing each dependency sets in a fresh, clean temporary directory and I don't quite get when the package will be already installed in the system.

Also, this is very fragile. There is no guarantee that the package will have either the `<package_name>` directory or the `__init__.py` file in it.

packages/worker-bundler/src/installer.ts
```diff
@@ -250,6 +314,116 @@ async function installPackage(
...
+      const version = metadata.info.version;
+      const wheel = metadata.urls.find(
+        (url) => url.packagetype === "bdist_wheel"
+      );
+      if (!wheel) {
+        throw new Error(
+          `No wheel distribution found for ${name}@${version} on PyPI`
+        );
+      }
+      const wheelUrl = wheel.url;
```
ryanking13, line 365: A single package version will have multiple wheels for different platforms. So just checking the `bdist_wheel` is not enough and you need to check the tags to make sure the wheel is installable for Python workers.

packages/worker-bundler/src/installer.ts
```diff
@@ -250,6 +314,116 @@ async function installPackage(
...
+
+      const response = await fetchWithTimeout(
+        wheelUrl,
+        {},
+        DEFAULT_TIMEOUT_MS * 2
```
ryanking13, line 370: Just curious, why not double the DEFAULT_TIMEOUT_MS itself?

packages/worker-bundler/src/installer.ts
```diff
@@ -250,6 +314,116 @@ async function installPackage(
...
+ */
+function stripWheelToPackage(
+  files: Record<string, string>
+): Record<string, string> {
+  const result: Record<string, string> = {};
+  for (const [path, content] of Object.entries(files)) {
+    // Skip wheel metadata and data directories
+    if (path.includes(".dist-info/") || path.includes(".data/")) {
+      continue;
+    }
```
ryanking13, line 422: Why? These are important files used in runtime as well.

packages/worker-bundler/src/installer.ts
```diff
@@ -286,6 +460,30 @@ async function fetchPackageMetadata(
   return (await response.json()) as NpmPackageMetadata;
 }
 
+async function fetchPythonPackageMetadata(name: string, registry: string) {
+  // Fetch package metadata from PyPI JSON API
+  const metadataResponse = await fetchWithTimeout(`${registry}/${name}/json`);
```
ryanking13, line 468: This is a legacy API endpoint that only PyPI uses. There is a [PEP 691](https://peps.python.org/pep-0691/) which is a standard now, so I would recommend to use that instead.

packages/worker-bundler/src/installer.ts
```diff
@@ -348,6 +546,32 @@ export async function fetchPackageFiles(
...
+function extractWheel(
+  data: Uint8Array,
+  result: InstallResult
+): Record<string, string> {
+  const unzipped = unzipSync(data);
+  const files: Record<string, string> = {};
+  const textDecoder = new TextDecoder();
+
+  for (const [path, content] of Object.entries(unzipped)) {
+    if (!isTextFile(path)) {
```
ryanking13, line 568: a python package can contain non-python files, so I don't think this filtering is needed.

packages/worker-bundler/src/installer.ts
```diff
@@ -250,6 +314,116 @@ async function installPackage(
...
 }
 
+/**
+ * Install a single Python package from PyPI.
+ *
+ * This is a minimal implementation: it downloads the latest version of the
+ * package as a source distribution and adds it to python_modules/. It does not
+ * resolve version ranges or install transitive dependencies.
+ */
+async function installPythonPackage(
```
ryanking13, line 329: How are the nested dependencies handled? It look like only the top-level dependencies are installed and the nested one are not if I understadn correctly.

ryanking13 (commented):
I left a few comments, but it looks like a good starting point.

Not all of my reviews need to be addressed in this PR. Some of my comments are for FYI, as you'll eventually handle them. If you consider some of my comments as out-of-scope of this PR and would want to hand it in the follow up PRs (which is a good practice), then please add a comment about it.

packages/worker-bundler/src/tests/wrangler.jsonc
abstractedfox: I actually think this was just a git fumble, I don't remember changing it on purpose. I'd originally started this on an older fork by mistake and reconstituted the work on the current one instead of changing the remote url, so I think it was due to that

packages/worker-bundler/src/index.ts
abstractedfox: I recall just choosing it as an arbitrary date that I thought would work, I can put a newer one. The JS worker as in the worker that creates the dynamic worker?

packages/worker-bundler/src/index.ts
abstractedfox: As in to the user-defined compatibility flags, in the event the user passed flags that didn't include `python_workers`? I can add that

packages/worker-bundler/src/installer.ts
abstractedfox: I actually only did it this way because I dryly assumed I'd need a lib for it, I'd much rather save a dependency so I'm glad you told me that

packages/worker-bundler/src/installer.ts
abstractedfox: ahh thanks, that actually answers a question that was on my mind this morning. I may want to do that in an upcoming pr if that's alright with you

packages/worker-bundler/src/index.ts
dom96: Actually, that should no longer be needed for Python dynamic workers. You can run python dynamic workers without that flag.

packages/worker-bundler/src/installer.ts
dom96: That's true. Not necessary for this PR, though you should add a TODO comment.

packages/worker-bundler/src/installer.ts
abstractedfox: You're seeing it correctly, the advice from Dominik was to PR something that worked without nested dependencies and then add that next, I actually do have that working in another branch so that'll be next

packages/worker-bundler/src/installer.ts
abstractedfox: ah, thanks for pointing that out, I realized distribution packages could have differently named packages inside them while working in another branch but I didn't change this there either (though I didn't know packages could not have `__init__.py`). I more or less did what the JS implementation was doing here, a snap assumption was that it was to prevent races in some way but the comment for the JS one says it's done in case installDependencies is called on a 'pre-warmed filesystem', though I don't know how often this actually happens

packages/worker-bundler/src/installer.ts
abstractedfox: Mostly because it's not clear to me whether that would have undesirable effects elsewhere, there are other places where this function is called with the default duration. I ended up just staying with what the JS one was doing

packages/worker-bundler/src/installer.ts
abstractedfox: Ah thank you, I was thinking earlier about what I'd need to do about platform specific wheels

packages/worker-bundler/src/installer.ts
abstractedfox: this one was mostly driven by me wanting packages with compiled extensions to fail gracefully since I haven't tested them yet and don't expect them to work right away

abstractedfox (commented):
Will have some more commits in for changes requested by @ryanking13, though I may want to save some of them for a future PR if that's alright

packages/worker-bundler/src/installer.ts
dom96: yeah, this PR is specifically not handling nested dependencies for now

packages/worker-bundler/src/installer.ts
dom96: That's a good way to do things, but be sure to add a TODO comment

packages/worker-bundler/src/installer.ts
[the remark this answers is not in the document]
abstractedfox: Following up on this: no, but it seems that the non-metadata subfolders simply _are_ the installable packages, so I changed it to just install each one in python_modules/<subfolder>

packages/worker-bundler/src/index.ts
abstractedfox: @dom96 Should I remove it?

packages/worker-bundler/src/installer.ts
abstractedfox: I did end up adding it, though with a comment wondering out loud if it's best to have it always use the latest version (which is what it does right now)

packages/worker-bundler/src/tests/e2e.test.ts
```diff
@@ -984,3 +984,116 @@ describe("hasNodejsCompat", () => {
...
+      ]
+    }));
+
+    let response = await worker
+      .getEntrypoint()
+      .fetch(new Request("http://worker/"));
+    expect(response.status).toBe(200);
+    expect(await response.text()).toBe("hi");
+  });
+}, 20000);
```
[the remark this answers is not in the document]
abstractedfox: Didn't review the code but passing it some impractically small value makes it fail in that many ms so I'm gonna resolve this

packages/worker-bundler/src/installer.ts
```diff
@@ -348,6 +554,32 @@ export async function fetchPackageFiles(
...
+      result.warnings.push(
+        `Could not install file ${path}, extension must match an approved text format type. This may corrupt this dependency.`
+      );
+      continue;
+    }
+    files[path] = textDecoder.decode(content);
+  }
+
+  return files;
+}
```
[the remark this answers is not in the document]
abstractedfox: Closing this as this is a temporary blockage that will be removed later

packages/worker-bundler/src/tests/e2e.test.ts
```diff
@@ -984,3 +984,123 @@ describe("hasNodejsCompat", () => {
...
+          '    return f"hello, {name}"'
+        ].join("\n")
+      }
+    });
+    const worker = env.LOADER.get(id, () => ({
+      mainModule: createWorkerResult.mainModule,
+      modules: createWorkerResult.modules,
+      compatibilityDate:
+        createWorkerResult.wranglerConfig?.compatibilityDate ?? "2026-01-01",
+      compatibilityFlags: createWorkerResult.wranglerConfig?.compatibilityFlags
```
dom96: nit: Should this also be an assert like the above?

packages/worker-bundler/src/tests/e2e.test.ts
```diff
@@ -984,3 +984,123 @@ describe("hasNodejsCompat", () => {
...
+          'dependencies = ["typing_extensions", "typing_inspection", "attrs"]'
+        ].join("\n")
+      }
+    });
+    const worker = env.LOADER.get(id, () => ({
+      mainModule: createWorkerResult.mainModule,
+      modules: createWorkerResult.modules,
+      compatibilityDate:
+        createWorkerResult.wranglerConfig?.compatibilityDate ?? "2026-01-01",
+      compatibilityFlags: createWorkerResult.wranglerConfig?.compatibilityFlags
```
dom96: nit: assert same as above

packages/worker-bundler/src/index.ts
```diff
@@ -143,6 +143,40 @@ export async function createWorker(
...
+    }
+
+    // Add the python_workers flag if user-supplied compatibilityFlags exists but doesn't include it
+    if (
+      wranglerConfig &&
+      wranglerConfig.compatibilityFlags &&
+      !wranglerConfig.compatibilityFlags.includes("python_workers")
+    ) {
+      wranglerConfig.compatibilityFlags.push("python_workers");
+    }
```
dom96: You can delete this. Python dynamic workers no longer need this flag being specified explicitly.

packages/worker-bundler/src/index.ts
dom96: Yep

packages/worker-bundler/src/index.ts
```diff
@@ -143,6 +143,40 @@ export async function createWorker(
     );
   }
 
+  if (entryPoint.endsWith(".py")) {
+    const modules: Modules = {};
+    for (const path of fileSystem.list()) {
+      if (path === "pyproject.toml") {
+        continue;
+      }
```
dom96, line 151: Is this necessary? No one is likely to be specifying pyproject.toml in the `createWorker` call

packages/worker-bundler/src/installer.ts
```diff
@@ -106,26 +118,76 @@ export async function installDependencies(
...
+    return;
   }
 
   // Collect dependencies to install
-  const depsToInstall: Record<string, string> = {
-    ...packageJson.dependencies,
-    ...(dev ? packageJson.devDependencies : {})
-  };
+  const depsToInstall: Record<string, string> = {};
+  depsToInstall["workers-runtime-sdk"] = "*"; // Should this always take the latest?
```
dom96: Be sure to mark these as "TODO" so we can find these by grepping

packages/worker-bundler/src/installer.ts
```diff
@@ -106,26 +118,76 @@ export async function installDependencies(
 
   // Read package.json
   const packageJsonContent = fileSystem.read("package.json");
-  if (!packageJsonContent) {
-    return result; // No package.json, nothing to install
+  const pyprojectTomlContent = fileSystem.read("pyproject.toml");
```
dom96, line 121: Thanks for moving the Python logic into a separate function. The diff can still be simplified by doing an early return here and keeping the rest of `installDependencies` unchanged.

Also while you're doing this you should early return with a warning if both are specified.

```suggestion
  const pyprojectTomlContent = fileSystem.read("pyproject.toml");
  if (packageJsonContent && pyprojectTomlContent) {
    result.warnings.push("Cannot have both package.json and pyproject.toml");
    return result;
  }
  if (pyprojectTomlContent) {
    return await installDependenciesPython(pyprojectTomlContent, fileSystem);
  }
```

dom96: Note how `installDependenciesPython` also returns an InstallResult in my example above. You should define the `result` in `installDependenciesPython` in the same way that it is done in this function.

packages/worker-bundler/src/installer.ts
dom96: With @ryanking13's comment below, perhaps the best thing to do here is to just delete this and rely on the `installedPackages` check instead. You can add a TODO to consider doing this in the future because the JS is doing it.

packages/worker-bundler/src/installer.ts
dom96: https://github.com/cloudflare/agents/pull/1881/changes#r3581110896

packages/worker-bundler/src/installer.ts
dom96: I think this kind of thing can be pushed to a TODO comment and done in separate PRs. Let's iterate via follow up PRs rather than ensuring everything is 100% in this PR, especially when it comes to Python packaging details.

packages/worker-bundler/src/installer.ts
dom96: Is it easy to move to the new endpoints? Like is it just a URL change? If not we can do this in a follow up (so a TODO comment is fine here IMO)

dom96 (approved):
A few more comments and things that I think should be fixed before merge, but overall looks good to me so approving.

packages/worker-bundler/src/index.ts
abstractedfox: @dom96 This is the pyproject.toml that contains all the deps for the worker, so as far as I understand we do need to have it coming into createWorker

packages/worker-bundler/src/installer.ts
abstractedfox: I did end up starting this one on a separate branch, will add a todo

abstractedfox (commented):
One last question in the comments but I think this is otherwise just about there!

packages/worker-bundler/src/installer.ts
abstractedfox: Took care of this across two commits (wanted to mention since it doesn't show as outdated or anything in the gh ui)

packages/worker-bundler/src/index.ts
dom96: Ahh, my mistake, for some reason I thought pyproject.toml wasn't in the "files" dict, but one level above (same for packages.json, but I just checked and it's also in the "files" dict).

All good to leave this then.

packages/worker-bundler/src/index.ts
ryanking13: Yeah, following whatever the JS worker is doing would be a good idea.

packages/worker-bundler/src/installer.ts
ryanking13: sure sounds good (please add a comment in the code for the reviewers)

packages/worker-bundler/src/installer.ts
ryanking13: Yeah, note that without this, only pure python packages will be successfully installed, and all other platform-dependent wheels are not going to pick the correct one. So you can test only with a pure python package then start testing cross-compiled packages in a follow up PR.

packages/worker-bundler/src/installer.ts
ryanking13: A little bit of URL change + json handling. You can start with this legacy api and switch to the PEP 691 one in a follow up PR (note that the Pyodide index (index.pyodide.org) does not support this endpoint)

ryanking13:
cc: @threepointone The python workers team doesn't seem to be possible to approve and merge this. Can you help with this?

xcaptain:
So if this PR got merged, dynamic worker can be a full featured python code runner, no need to use docker based sandbox right? I don't like the docker sandbox because it's imcompatitable with [vitest-pool-workers](https://github.com/cloudflare/workers-sdk/issues/10408) which make writing tests very hard (need to start real server to connect to docker).

ryanking13:
@xcaptain This is still experimental so the answer is yes and no. I would say you should not rely on this yet (or you can volunteer to help us debugging things :) ). 

> no need to use docker based sandbox right? I don't like the docker sandbox because it's imcompatitable with https://github.com/cloudflare/workers-sdk/issues/10408 which make writing tests very hard (need to start real server to connect to docker).

Also, yes and no. It is true that you can use dynamic workers with Python, but it you need to run a workload that is computation / memory heavy, then it is not a good fit for worker. So it depends on your workload

mattzcarey:
this looks like a lot of work and thought has been put into this but I would love to know what the aim is and why we want this? can someone DM me to chat. We dont have any python support really in this repo yet. 

cc. @aron-cf @ghostwriternr for viz

abstractedfox:
Closing in favor of #2043

PR #1887, opened 2026-07-07
refactor(agents): extract state subsystem into agent/state.ts

cjol:
This PR pulls the Agent state machinery out of the monolithic Agent class in `packages/agents/src/index.ts` and into a dedicated `packages/agents/src/agent/state.ts` module (`AgentState`). It is a surface-level extraction: behaviour and the public API are unchanged.

## Why

- `index.ts` is ~12k lines and the state logic (in-memory cache, SQLite persistence, protocol broadcast, connection readonly/no-protocol flags, state-change hook dispatch) was interleaved with everything else, making it hard for humans and agents to reason about in isolation.
- Grouping that logic in one documented module shrinks `index.ts` substantially and gives the state subsystem a single place to read and edit.
- The extraction uses a narrow `AgentStateHost<State>` callback interface rather than handing `AgentState` the whole Agent, so the boundary is explicit instead of "everything can reach everything". Passing the full Agent would have been less churn but would have preserved the coupling this change is meant to expose.
- This is deliberately not a functional or lower-layer extraction. `AgentState` still depends on Agent-only facilities, so it is not yet usable on a bare partyserver `Server`; see the couplings listed below.

## Architectural Changes

Before: state cache, persistence, broadcast, connection flags, and hook dispatch lived as methods and fields directly on `Agent`.

After: `Agent` holds a private `_state: AgentState<State>` and delegates to it. `AgentState` calls back into `Agent` only through the `AgentStateHost<State>` interface. `AgentStateApi<State>` describes the curated user-facing subset that `AgentState` implements, so a future major could expose the subsystem without leaking internal plumbing.

Still Agent-only (why it is not yet reusable on a lower layer). partyserver already provides `ctx`, `broadcast`, `getConnections`, `onError`, connections, and hibernation, so those are not blockers. The remaining couplings are:

- Ambient agent context: `state.ts` imports the `agentContext` AsyncLocalStorage directly and uses it for readonly-connection gating and to run the hook so `getCurrentAgent()` works. This must become host-provided context.
- Schema ownership: the `cf_agents_state` table is created by the Agent constructor (not by `AgentState`), the `sql` tagged-template helper is an Agent method, and the `cf_schema_version` migration framework co-inhabits the same table. `AgentState` is currently a tenant of a table the Agent layer owns.
- State-sync protocol/client: the live half depends on the `CF_AGENT_STATE` message type and a client that understands it (`useAgent` / `AgentClient`).
- Typed hook surface: `initialState`, `validateStateChange`, `onStateChanged` / `onStateUpdate` are Agent's user-facing override points.
- Observability: `_emit("state:update")` is Agent's event bus.
- Connection-flag semantics: the wrapping mechanism is pure partyserver, but the meanings (readonly = block writes, no-protocol = suppress `CF_AGENT_*` frames) are Agent-layer policy.

## Code Changes

- New `AgentState<State>` in `agent/state.ts` owns the state cache, persistence, broadcast, connection-flag machinery, and hook dispatch, talking back to Agent via `AgentStateHost<State>`.
- `Agent` delegates its state methods to `_state`. Public API (`state`, `setState`, `setConnectionReadonly`, `isConnectionReadonly`, `isConnectionProtocolEnabled`) is unchanged and still declared on `Agent.prototype`; `agent.state` still returns the state value.
- Removed five private pass-through wrappers on Agent (`_broadcastProtocol`, `_setStateInternal`, `_ensureConnectionWrapped`, `_setConnectionNoProtocol`, `_cf_getRawConnectionState`) and inlined their call sites to `_state`.
- Tombstoned `_unsafe_getConnectionFlag` / `_unsafe_setConnectionFlag` as `@deprecated` forwarders with no remaining in-repo callers.
- Documented the module: a top-level overview of what "state" is, and an expanded explanation of the persistence-hook dispatch mode.
- Test helpers that force the in-memory cache now go through the renamed private `_cachedState` accessor.

No changeset is included: `AgentStateApi` is exported from the module but not from the package's public entry points, and no published behaviour changes.

## Compatibility

Nothing user-facing changes. `_unsafe_getConnectionFlag` / `_unsafe_setConnectionFlag` remain on the prototype but are now `@deprecated` internal forwarders slated for removal in a future major; they were never part of the documented API.

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1887" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/agents/src/agent/state.ts
+++ b/packages/agents/src/agent/state.ts
@@ -1,0 +1,760 @@
+/**
+ * Agent state subsystem.
+ *
+ * "State" is a single JSON-serializable value owned by one Agent instance
+ * (one Durable Object). It is read with `agent.state` and replaced wholesale
+ * with `agent.setState(next)` — there are no partial updates.
+ *
+ * What makes it more than a plain field is that it is durable and live:
+ * - Durable: persisted as one row in the `cf_agents_state` SQLite table and
+ *   lazily rehydrated into an in-memory cache on the first read of a cold
+ *   isolate, so it survives hibernation and reconnects.
+ * - Live: `setState` (from the server or from a connected client) persists the
+ *   value and broadcasts it over WebSocket so the server and every connected
+ *   client stay in agreement, with the browser `useAgent` hook re-rendering on
+ *   each change.
+ *
+ * Use it for the small, canonical snapshot clients should watch update in real
+ * time (session/task status, progress, collaborative counters). For large or
+ * relational data (message history, logs) use the agent's raw SQL instead;
+ * `state` is loaded and rewritten whole and broadcast on every change, so keep
+ * it small.
+ *
+ * This module owns that behaviour (`AgentState`), the connection-state flag
+ * machinery (readonly / no-protocol), and the state-change hook dispatch. The
+ * `Agent` class holds an `AgentState` instance and delegates to it; see
+ * `AgentStateApi` for the curated, public-ready subset.
+ */
+import type { AgentEmail } from "../internal_context";
+import { __DO_NOT_USE_WILL_BREAK__agentContext as agentContext } from "../internal_context";
+import { MessageType } from "../types";
+import type { Connection } from "partyserver";
+
+// Row id for the agent's user-visible state in cf_agents_state.
+// A separate SCHEMA_VERSION_ROW_ID row in the same table tracks migrations.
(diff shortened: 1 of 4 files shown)


cjol:
Not sure this is useful enough in isolation.

PR #1888, opened 2026-07-07
fix(think): preserve system prompt when skills are enabled

### LINKED ISSUE #1871: getSystemPrompt() is silently ignored once getSkills() registers a context block
### Summary

When an agent overrides both `getSystemPrompt()` and `getSkills()`, the agent's system prompt is silently dropped from the model request. The model only receives the skills catalog context block + the generic "You are running inside a Think agent." capability block — the entire `getSystemPrompt()` output (persona, instructions, output contract) never reaches the model, and `getSystemPrompt()` is never even called.

### Environment

- `@cloudflare/think@0.11.1`
- Agent extends `Think`, running on Cloudflare Workers (Durable Object), streaming via the AI SDK.

### Expected

Skills should **augment** the agent's system prompt. `getSystemPrompt()` should still reach the model when `getSkills()` returns sources.

### Actual

As soon as `getSkills()` returns a non-empty source, `getSystemPrompt()` is ignored and its content is absent from the model's system message.

### Root cause (traced in the bundled `dist/think.js` of 0.11.1)

1. `getSystemPrompt()` is documented as a fallback only:
   > "Used as fallback when no context blocks are configured via `configureSession`."

2. Skills initialization registers the catalog as a **context block**:
   ```js
   // _initializeSkills()
   await this.session.addContext(registry.contextLabel, {
     description: "Think skills: available skill catalog",
     provider: { get: () => registry.systemPrompt() },
   });
   ```

3. The per-turn assembly is mutually exclusive — context block OR fallback, never both:
   ```js
   const rawBaseSystem = await this.session.freezeSystemPrompt() || this.getSystemPrompt();
   ```
   Since skills registered a context block, `freezeSystemPrompt()` is truthy, so `getSystemPrompt()` is never called.

Result: the model system = skills catalog + `_buildThinkCapabilityBlock()` ("You are running inside a Think agent."). The agent's own system prompt is gone.

### Minimal repro

```ts
class MyAgent extends Think<Env> {
  getModel() { return myModel }
  getSystemPrompt() { return "You are Robbie, a pirate. Always answer in pirate speak." }
  getSkills() { return [mySkillSource] } // any non-empty source
}
```

Send any message → the model's system message contains the skills catalog + "You are running inside a Think agent.", but NOT "You are Robbie, a pirate."

We confirmed by capturing the outbound provider request: the system prompt was ~4.2k chars of skills catalog + capability block, with none of `getSystemPrompt()`'s content. Removing `getSkills()` makes `getSystemPrompt()` work again.

### Impact

Any agent that uses skills AND relies on `getSystemPrompt()` for its persona/instructions/output format ships to the model without them — silently, with no error or warning. Easy to miss because the agent still responds and uses tools; it just runs without its own instructions.

### Workaround

Register the base system prompt as its own context block in `configureSession`:

```ts
configureSession(session: Session) {
  return session
    .withContext("base-system", { provider: { get: async () => this.getSystemPrompt() } })
    .withCachedPrompt()
}
```

### Suggested fix / question

- Should the skills context block be **additive** to `getSystemPrompt()` instead of replacing it (assemble `getSystemPrompt()` + context blocks, rather than `freezeSystemPrompt() || getSystemPrompt()`)?
- If "with context blocks, `getSystemPrompt()` is ignored" is intended, please (a) document that explicitly on `getSystemPrompt()` / `getSkills()`, and (b) emit a dev warning when both an overridden `getSystemPrompt()` and skills are present.

mattzcarey:
## Summary

- keep `getSystemPrompt()` in the assembled prompt when `getSkills()` registers the Think skills catalog
- preserve the existing replacement behavior for ordinary Session context blocks when skills are not enabled
- document the composition rule and add a patch changeset

Closes #1871.

## Testing

- `pnpm --filter @cloudflare/think test`
  - 42 Workers test files / 896 tests
  - generated-entry: 10 tests
  - Vite: 10 tests
  - CLI: 55 tests
  - React: 5 tests
- `tsc --noEmit -p packages/think/tsconfig.json`
- `tsc --noEmit -p packages/think/src/tests/tsconfig.json`
- repo-wide export, formatting, and lint checks pass

`pnpm run check` reaches the repo-wide typecheck, where three existing voice example projects fail against the current voice types (`examples/telnyx-voice-agent`, `examples/voice-agent`, and `examples/playground`). The affected Think package and test TypeScript projects pass.

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1888" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/tidy-skills-keep-prompts.md
+++ b/.changeset/tidy-skills-keep-prompts.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/think": patch
+---
+
+Keep `getSystemPrompt()` in the assembled model prompt when `getSkills()` registers its catalog context.
--- a/packages/think/README.md
+++ b/packages/think/README.md
@@ -209,8 +209,8 @@ export class MyAgent extends Think<Env> {
 
 Think supports the [Agent Skills](https://agentskills.io/) directory format as
 a first-class API. Return one or more `SkillSource` objects from `getSkills()`;
-Think adds the skill catalog to the prompt and exposes `activate_skill` and
-`read_skill_resource` tools when skills are available.
+Think appends the skill catalog to `getSystemPrompt()` and exposes
+`activate_skill` and `read_skill_resource` tools when skills are available.
 
 ```ts
 import { Think, skills } from "@cloudflare/think";
@@ -268,9 +268,10 @@ can read `{ name, path }` or a qualified path such as
 from other skills.
 
 Skills are on-demand instructions, not always-on system prompt text. The model
-sees the catalog first, then calls `activate_skill` when a user task matches a
-skill description. Use `getSystemPrompt()` or a Session context block for
-behavior that should apply to every turn.
+sees their catalog alongside the base prompt, then calls `activate_skill` when a
+user task matches a skill description. Enabling skills does not replace
+`getSystemPrompt()`. Use that method or a Session context block for behavior that
+should apply to every turn.
 
 Script execution is opt-in and **experimental**. `getSkillScriptRunner()`
 enables `run_skill_script`, which can run JavaScript, TypeScript, Python, and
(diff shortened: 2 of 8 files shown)


mattzcarey:
actually idk what to do here.

mattzcarey:
will look more tomorrow

arthurfiorette:
Hey @mattzcarey, any updates here? just faced this problem as well.

mattzcarey:
This is annoying sorry. Very bad signposting from us. The best way to get around this is to use the context blocks api in sessions to build the system prompt. 

This getSystemPrompt is actually legacy now and used as a fallback.

mattzcarey:
Point your agent at agent-think in this repo to show how this should be done. 

@arthurfiorette

PR #1895, opened 2026-07-08
feat(mcp): make the client manager a Durable Object capability

mattzcarey:
## Summary

Make `MCPClientManager` a reusable capability for the public Durable Object lifecycle shipped in #2133.

A plain Durable Object keeps the manager as `this.mcp` and installs that same object directly:

```ts
import { DurableObject } from "cloudflare:workers";
import { Lifecycle } from "agents/lifecycle";
import { MCPClientManager } from "agents/mcp/client";

export class MyObject extends DurableObject<Env> {
  readonly mcp = new MCPClientManager("my-object", "1.0.0", {
    storage: this.ctx.storage
  });

  readonly lifecycle = Lifecycle.install(this).use(this.mcp);

  onRequest() {
    return Response.json({ tools: this.mcp.listTools() });
  }
}
```

`MCPClientManager` owns its durable MCP-client work:

- SQLite schema initialization
- persisted HTTP and RPC connection restoration
- default OAuth-provider construction for fresh registrations
- exact OAuth callback URL interception and response handling
- state-change publication after startup and callbacks

Lifecycle invokes these hooks automatically. Native Durable Object RPC bypasses Lifecycle handlers, so RPC methods that use capabilities call `await this.lifecycle.start()` first.

## Lifecycle host context

`agents/lifecycle` exports `LifecycleObject`: a Cloudflare `DurableObject` with an installed `Lifecycle` and the semantic host hooks Lifecycle dispatches. This is a host interface, not the `Agent` class exported from `agents`.

The canonical context accessor is exported from `agents/lifecycle`:

```ts
import { getCurrentAgent } from "agents/lifecycle";
```

The root `agents` package exports the same function for existing `Agent` applications.

Lifecycle establishes context only while invoking host hooks:

- startup and alarm: host object
- HTTP request: host object and request
- WebSocket connect: host object, connection, and upgrade request
- WebSocket message, close, and error: host object and connection

`getConnectionTags(connection, { request })` remains argument-driven because both values are already explicit.

Capability hooks deliberately run outside host context. They use their own `this`, hook arguments, and explicitly supplied dependencies. Lifecycle exits any inherited current-Agent context before capability startup, request, and alarm hooks, so capability behavior does not depend on the entrypoint that triggered it.

There is no public capability-phase callback, phase context type, Lifecycle options object, or Agent-specific MCP adapter.

The `Agent` class remains the `Agent` class. Its existing public-method and tracing wrappers retain responsibility for native Durable Object RPC, cross-Agent entry, callables, email, schedules, fibers, chat turns, and detached work. Native-RPC initialization remains separate in #2156; tracing consolidation is not part of this PR.

## Agent compatibility

`Agent.this.mcp` remains the public `MCPClientManager`. Existing `addMcpServer()`, `removeMcpServer()`, `getMcpServers()`, custom OAuth-provider, protocol broadcast, and observability behavior remains available.

Agent installs the manager directly:

```ts
this.lifecycle.use(this.mcp);
```

The reusable constructor surface is intentionally small:

```ts
new MCPClientManager(name, version, {
  env, // optional for HTTP-only catalogs; required to restore rpc:// rows
  storage,
  createAuthProvider // optional
});
```

There is no separate OAuth storage-name or RPC resolver option. `MCPClientManager` owns RPC restoration through explicit `env`: it maps persisted Wrangler binding names to live `DurableObjectNamespace` values and reconnects through its ordinary RPC transport. If an HTTP-only manager omits `env` while persisted RPC rows exist, startup warns for each affected server rather than silently dropping them.

Agent restores facet identity and virtual connection bridges in its own startup. MCP protocol publication waits until that routing state is ready.

## Routing and recovery

The manager does not impose an OAuth route shape. It persists the exact callback URL supplied during registration and intercepts only requests whose origin and pathname match it. The outer Worker only needs to forward that URL to the same named Durable Object through `routeAgentRequest()`.

Recovery tests use real `evictDurableObject()` calls. They cover plain-Durable-Object HTTP restoration, Agent RPC restoration, connection readiness, exact callback interception, and MCP restoration before existing fiber recovery.

This PR remains MCP-focused. It does not extract schedules or fibers as capabilities.

## Example

`examples/next/mcp-client` is a Vite + React plain-Durable-Object example. It accepts any Streamable HTTP MCP endpoint, completes OAuth in separate popups, displays discovered tools, accepts JSON arguments, and renders tool results.

Live demo: https://next-mcp-client-capability.mattzcarey.workers.dev

Each browser receives a stable random Durable Object name stored locally, isolating catalogs, OAuth state, credentials, and tool calls between visitors. The former shared `demo` object remains empty.

## Verification

On commit `1255c408`:

- Lifecycle Workers: **12/12**, covering capability context isolation (including inherited foreign context), host startup/request/alarm context, root accessor aliasing, WebSocket connect/message/close context, real eviction, and fresh OAuth registration
- Agent Workers: **1,846/1,846**
- AI Chat: **737/737**
- Think Workers: **886/886**
- Think React: **2/2**
- Channels: **161/161**
- focused MCP OAuth: **30/30**
- focused sub-agent and connection routing: **80/80**
- full repository build: **19/19**
- package, isolated Lifecycle, and both example TypeScript checks
- exports, formatting, and lint are green
- repository typecheck: **115/117** projects; the two `agent-think` TS2589 failures reproduce on clean `origin/main`

Fresh repository CI is running for this commit.


docs/agents/mcp-client.md
```diff
@@ -40,6 +40,28 @@ export class MyAgent extends Agent {
...
+  override mcp = new MCPClientManager(this, {
+    name: "my-agent",
+    version: "1.0.0"
+  });
+}
+```
+
+The manager is resolved when Agent startup begins, after subclass field initializers run. Replacing `mcp` therefore replaces the default component rather than registering both.
+
+`MCPClientManager` no longer has a standalone storage-only constructor. It must be attached to an `Agent` through the host-first form above so startup, OAuth callbacks, turn tools, protocol updates, observability, and destruction all use the same lifecycle.
```
cjol: I want to think about this for a little bit. Maybe it's the right move, but it makes the interface contract between MCPClientManager and host agent implicit and invites tighter coupling than I'd like.

packages/agents/src/index.ts
```diff
@@ -2249,30 +2147,43 @@ export class Agent<
 
     this._ensureSchema();
 
-    // Initialize MCPClientManager AFTER tables are created
-    this.mcp = new MCPClientManager(this._ParentClass.name, "0.0.1", {
+    const mcpHost: MCPClientManagerHost = {
```
cjol: (same point as my other comment, really) This is effectively the object that we're requiring non-Agent hosts to implement in order to use our MCP client, which is quite extensive and also non-specific (e.g. what bindings need to be returned by `getEnv`?)

Equally I know from experience trying it with Channels (and then State, where I ended up with a [similar situation](https://github.com/cloudflare/agents/pull/1887/changes#diff-2b5a0a580637779323f8b95ddd735d874ec6cd71d6827350c32787339cf69642R162-R179)), that this is the bit where extraction from the current spaghetti gets very hairy, and it's what made me want to go in the opposite direction with Channels -- build a neat, isolated interface for the module, and then put all the adapter code inside Agent rather than bending the module to suit the current Agent implementation.

cjol (commented):
I think the lifecycle hooks are great and work well here. I'm not entirely sure what each represents yet (what's a request vs a turn? Doesn't matter right now, and I know we'll obviously document them prominently eventually) but it feels a good shape.

I'm not so sure about the interface of MCPClientManager but -- a bit like my state PR from yesterday, maybe it's still progress even if it's not the best shape yet.

mattzcarey:
CI diagnosis for the red `think (ai@6)` lane:

- the compatibility job performs an unfrozen install of `ai@^6.0.0`;
- the earlier green run resolved `ai@6.0.259`;
- both current attempts resolve newly published `ai@6.0.260` and fail the same 16 tests;
- the exact lane reproduces with the same failures on current `origin/main`;
- in an isolated checkout of this PR, pinning only `ai` back to `6.0.259` makes the two failing files pass **97/97**.

All MCP/conformance lanes and the main `ci` job pass. This is an upstream AI SDK patch regression, so I have not mixed an unrelated compatibility pin into this MCP PR.

mattzcarey:
Follow-up CI fix: #2141 caps the unfrozen AI SDK v6 compatibility resolver at the last known-good patch (`>=6.0.0 <6.0.260`). This is separate from the MCP diff. After testing the registry updates: `6.0.260`, `.261`, and `.263` all reproduce the same 16 failures; `6.0.259` passes the full Think Workers suite (**886/886**).

docs/agents/mcp-client.md
mattzcarey: Resolved by the lifecycle redesign. There is no host interface or hidden host registry now: a plain Durable Object constructs `MCPClientManager` with its own storage and installs that same object with `Lifecycle.install(this).use(this.mcp)`. The only optional dependency is OAuth-provider construction.

packages/agents/src/index.ts
mattzcarey: Agreed. The final shape now follows this direction without an Agent MCP adapter. `MCPClientManager` is the isolated capability and both plain Durable Objects and Agent install that same object directly. Lifecycle supplies one generic capability-phase execution boundary for Agent context, facet hydration, tracing, and errors. MCP receives `env` explicitly for persisted RPC binding reconstruction. Plain hosts implement no broad host contract or service locator.

examples/next/mcp-client/tsconfig.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": "agents/tsconfig",
+  "include": ["src/**/*.ts", "env.d.ts"]
```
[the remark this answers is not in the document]
mattzcarey: Fixed in d1afc691. The example TSConfig now includes `src/**/*.tsx`; `pnpm --filter @cloudflare/agents-next-mcp-client-example run typecheck` passes with the React client in the program.

mattzcarey:
Final review update:

- removed the broad host contract and public binding resolver; Agent and plain Durable Objects now install the same `MCPClientManager` object directly;
- added a generic Lifecycle execution boundary so Agent invocation context, facet hydration, tracing, and errors apply without an MCP adapter;
- moved RPC restoration into `MCPClientManager` with explicit `{ env }` and warnings when RPC rows cannot be restored;
- added regression tests for OAuth callback/provider context, lifecycle phase ordering, RPC restoration, and missing `env`;
- fixed the React example TSConfig and isolated deployed demo sessions per browser.

mattzcarey:
Architecture follow-up in `7522ad05`: `Agent` now installs `MCPClientManager` directly with `this.lifecycle.use(this.mcp)`. `_agentMcpLifecycleAdapter()` has been deleted, and Agent no longer calls `mcp.onStart()` or `mcp.onRequest()` manually. Agent-owned RPC binding reconstruction remains in its normal startup sequence.

mattzcarey:
Follow-up: the two direct-install regressions found here are now fixed at the Lifecycle seam rather than by restoring the adapter. `354fa7e0` adds a generic capability-phase execution boundary; `fa17cee6` installs `this.mcp` directly and gives the manager explicit `env` ownership for RPC restoration. The callback/provider context probes and real-eviction tests pass on this architecture.

mattzcarey:
Architecture resolved in `354fa7e0` + `fa17cee6`: Lifecycle now owns a generic capability-phase execution boundary, Agent installs `this.mcp` directly, and `MCPClientManager` receives explicit `env` to restore persisted RPC bindings itself. The private MCP adapter and Agent-managed RPC restore method are gone. Missing `env` with persisted RPC rows warns per server. Full local regression matrix is green; fresh CI is running.

mattzcarey:
Thermo-nuclear quality pass complete on `4cb8baa5`:

- extracted RPC restoration from the 2.4k-line manager into focused `client-rpc.ts`;
- centralized MCP table DDL in `client-storage.ts`;
- removed the redundant public lifecycle phase alias and moved host-policy types to the Lifecycle module;
- removed the demo's opener/postMessage authority and redeployed version `058f0ea6-40ee-4062-a584-e7b06e650276`;
- moved lifecycle Workers tests to `src/tests/lifecycle/` while retaining their isolated Wrangler/Vitest project;
- merged current `main` (Channels #2129).

No file crosses 1k lines because of this PR; both existing giant implementation files shrink. Fresh CI is running.

mattzcarey:
Thermo-nuclear review final result: **approved on code quality, pending human review**.

The ambitious cleanup is included in `4cb8baa5`: direct capability composition, generic Lifecycle execution boundary, focused RPC restoration module, canonical MCP schema owner, no new 1k-line crossing, hardened OAuth popup, and lifecycle tests colocated under `src/tests/lifecycle/`. Current-main merge is included. All 14 checks are green and all review threads are resolved. @cjol, this head is ready for your final architecture review.

design/durable-object-lifecycle.md
```diff
@@ -0,0 +1,83 @@
+# Durable Object lifecycle
+
+`Lifecycle` is the composition root that turns a Cloudflare Durable Object into
+an Agent. It owns runtime handlers, ordered capability phases, current Agent
+context, and hibernating WebSocket connections.
+
+The `ComposableAgent` interface exported from `agents/lifecycle` extends
```
cjol: Why is composableagent required over DurableObject?

design/durable-object-lifecycle.md
```diff
@@ -0,0 +1,83 @@
...
+For an HTTP request, capability `onRequest` hooks run in registration order. The
+first `Response` wins; otherwise the Agent's `onRequest` handles the request.
+For an alarm, capability `onAlarm` hooks run in registration order before the
+Agent's `onAlarm`.
+
+Failures stop the phase and propagate. Failed startup remains retryable. Native
+RPC methods explicitly call `lifecycle.start()` because native Durable Object
+RPC bypasses Lifecycle handlers.
+
+## Current Agent context
```
cjol: To what extend should we be talking about "Agent" context? Lifecycle is usable outside of the Agent class, right?

design/durable-object-lifecycle.md
```diff
@@ -0,0 +1,83 @@
...
+explicitly at construction.
+
+## Additional entry surfaces
+
+The batteries-included Agent has entry surfaces beyond Lifecycle, including
+native Durable Object RPC, WebSocket callables, email, schedules, fibers, chat
+turns, and detached work. Its existing invocation wrappers continue to own
+those surfaces. In particular, the automatic public-method wrapper is not made
+redundant by Lifecycle because native Durable Object RPC does not pass through a
+Lifecycle handler.
```
cjol: See #2005

docs/agents/get-current-agent.md
```diff
@@ -1,204 +1,127 @@
...
 
-## Real-World Example
+## What is an Agent?
 
-```typescript
-import { AIChatAgent } from "@cloudflare/ai-chat";
-import { getCurrentAgent } from "agents";
-import { generateText } from "ai";
-import { openai } from "@ai-sdk/openai";
+An Agent is any Cloudflare Durable Object with an installed `Lifecycle`:
```
cjol: Maybe not the time to discuss this, but I hate this definition of "Agent". Especially since we then have to distinguish with "the batteries-included Agent class" further below, I think this is asking for confusion

PR #1906, opened 2026-07-09
fix: allow user-defined interfaces as agent Props

### LINKED ISSUE #1886: Allow proper types for agent props
**Describe the bug**
The generic types for `Agent` enforces `Record<string, unknown>` as bound for `Props`. This Record type requires subtypes to define an index signature for type `string`, which prevents using interfaces with well-defined properties.

**To Reproduce**

**Expected behavior**
While the type could still *default* to `Record<string, unknown>`, it should allow using proper interfaces that have no index signature.

**Screenshots**

This is the error reported by Typescript on a `getAgentsByName` call passing `{ props: config }` as third argument:

Image

**Version:**
0.17.3

**Additional context**

mattzcarey:
## Summary

The `Props` generic on `Agent` (and every API that carries props) was bounded by `Record<string, unknown>`. Interfaces do not get implicit index signatures in TypeScript, so a well-typed interface fails that bound and users hit:

```
Type 'MyProps' is not assignable to type 'Record<string, unknown>'.
  Index signature for type 'string' is missing in type 'MyProps'.
```

exactly as reported on a `getAgentByName(ns, name, { props: config })` call.

## Fix — `object` bounds + derived props generics (no `any`)

- **`Props extends object = Record<string, unknown>`** on `Agent`, `McpAgent`, `getAgentByName`, and `AgentGetOptions`. Interfaces satisfy `object` natively; defaults stay `Record<string, unknown>` so untyped usage still reads props values as `unknown`.
- **partyserver seam**: partyserver still bounds its `Props` at `Record<string, unknown>` (`Server`, `getServerByName`), which we can't change from this repo. `Agent` bridges it with `extends Server<Env, Props & Record<string, unknown>>` — the intersection is erased at runtime (props is a JSON bag) — and `getAgentByName` widens the props value with one commented cast when handing off to `getServerByName`.
- **`addMcpServer` RPC props are now *derived*, not loosened**: the overload takes `AddRpcMcpServerOptions<McpAgentProps<T>>` where `McpAgentProps<T> = NonNullable<T["props"]>` reads the Props type off the target `McpAgent`. Props passed over RPC are type-checked against the agent actually being connected to — a wrong-shape props object is now a compile error (previously any record was accepted). The old bare-`McpAgent` constraint also rejected typed subclasses through Durable Object stub variance; the constraint is now `T extends McpAgent<Cloudflare.Env, unknown, object>`.

## Repro / regression tests

New compile-time test `src/tests-d/agent-props.test-d.ts`, following the existing `tests-d/` pattern (runs under the package typecheck, which CI enforces). Written first and confirmed red on main with 6 errors matching the issue:
- `getAgentByName` with inferred and explicit interface `Props`
- `Agent` / `McpAgent` class declarations with interface `Props`, and `this.props` keeping the interface type
- `addMcpServer(name, namespace, { props })` with an interface-Props `McpAgent` namespace, plus `McpAgentProps<T>` extraction
- negative pins: primitives are still rejected as `Props`, and wrong-shape RPC props are rejected against the target agent's declared Props

## Verification

- `npm run check` clean (sherif, exports, oxfmt, oxlint, all 116 typecheck projects)
- Full `packages/agents` runtime suite passing (types-only change)

## Out of scope / follow-up

`routeAgentRequest`'s `AgentOptions` pins props at the partyserver default and would need an upstream partyserver bound change (`object` instead of `Record<string, unknown>` on `Server`/`getServerByName`/`routePartykitRequest`) — that would also remove the intersection bridge and cast here.

Fixes #1886

### DIFF
--- a/.changeset/shiny-mirrors-behave.md
+++ b/.changeset/shiny-mirrors-behave.md
@@ -1,0 +1,7 @@
+---
+"agents": patch
+---
+
+Allow user-defined interfaces as agent `Props`.
+
+Interfaces do not get implicit index signatures in TypeScript, so the previous `Record<string, unknown>` bound rejected them with "Index signature for type 'string' is missing". A shared `AgentProps` bound (exported from `agents`) is now used everywhere props are typed — the `Agent`, `McpAgent`, `getAgentByName`, and `AgentGetOptions` generics, the `addMcpServer` RPC constraint and its `props` option, and the RPC client transport options — so plain interfaces work as props. Generic defaults stay `Record<string, unknown>`, so untyped usage still reads props values as `unknown`.
--- a/.changeset/shiny-mirrors-behave.md
+++ b/.changeset/shiny-mirrors-behave.md
@@ -4,4 +4,6 @@
 
 Allow user-defined interfaces as agent `Props`.
 
-Interfaces do not get implicit index signatures in TypeScript, so the previous `Record<string, unknown>` bound rejected them with "Index signature for type 'string' is missing". A shared `AgentProps` bound (exported from `agents`) is now used everywhere props are typed — the `Agent`, `McpAgent`, `getAgentByName`, and `AgentGetOptions` generics, the `addMcpServer` RPC constraint and its `props` option, and the RPC client transport options — so plain interfaces work as props. Generic defaults stay `Record<string, unknown>`, so untyped usage still reads props values as `unknown`.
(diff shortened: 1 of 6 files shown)


mattzcarey:
Closing in favor of fixing the Props bound upstream in partyserver (cloudflare/partykit), where `Server`/`getServerByName` own the `Record<string, unknown>` constraint — that removes the need for the intersection bridge and cast here. Issue #1886 stays open until the upstream fix lands and agents bumps the dependency.

stof:
This PR changing the bound in `Agent` will still be necessary even when the partyserver PR is merged (the partyserver PR will allow removing the cast in the *implement* of the function, but will not automatically change the public API in this repo)

PR #1911, opened 2026-07-09
fix(mcp): configure client elicitation handler

mattzcarey:
Follow-up to #1903/#1910 (none of this has shipped — the pending Version Packages release absorbs all of it, so no compatibility path is kept).

## What changed

Elicitation handling moves off the Agent class onto the MCP client manager, matching the `configureOAuthCallback` pattern. `Agent.onElicitRequest` is removed.

## How to use

Register handlers in `onStart()`. The handler keys are the capability declaration: connections advertise exactly the modes with configured handlers at the `initialize` handshake — `form` only, `url` only, or both. No handlers → no elicitation capability advertised, so servers use their non-elicitation fallbacks.

```ts
class MyAgent extends Agent<Env> {
  onStart() {
    this.mcp.configureElicitationHandler({
      url: async (request, serverId) => {
        // deliver request.params.url to the user out-of-band
        return { action: "accept", content: {} };
      },
      form: async (request, serverId) => {
        // collect fields matching request.params.requestedSchema
        return { action: "accept", content: { /* fields */ } };
      },
    });
  }
}
```

An explicit `client.capabilities.elicitation` on `addMcpServer` still wins wholesale (persisted, survives hibernation), e.g. to declare `form: { applyDefaults: true }`.

## Surviving hibernation: persisted capabilities

Handlers are functions and cannot be persisted, but the capabilities they advertise can. Whenever handlers are configured (and whenever a server is registered), the handler-derived client capabilities are recorded on each stored server row (`server_options.capabilities`, a general `ClientCapabilities` object — only `elicitation` is populated today, but future handler-driven capabilities such as sampling reuse the same field with no storage migration).

On wake, restored connections re-advertise the same modes at the `initialize` handshake before `onStart()` has run. Elicitation dispatch looks handlers up at request time, so when `onStart()` calls `configureElicitationHandler`, the handlers attach to the already-live connections — no reconnect needed. The seeding is self-contained in the manager: `createConnection` consumes the stamp from its own stored row, so every path that recreates a known server (storage restore, the Agent RPC restore, `addMcpServer` re-adds) gets it uniformly with no parameter threading.

The stamp is valid for **one restore**: consuming it clears it, and any `configureElicitationHandler` call re-stamps every row. Apps that configure handlers each wake see no difference; a deploy that deletes the configure call stops advertising the stale modes after a single wake instead of forever (adversarial review caught the forever-stale variant). Configuring handlers also clears the in-memory seed (handlers become the source of truth). Explicit `client.capabilities` declarations win over the seed, and reconnects rebuild the SDK client so handler changes on a live connection take effect at the next handshake.

Precedence at client build: `client.capabilities.elicitation` (declared) → handler-derived → persisted seed.

## Lifecycle

Because restored connections no longer depend on handlers existing at connect time, automatic MCP restore keeps its original position: **before** fiber/chat recovery and user `onStart()`. Recovered turns (`waitForMcpConnections`) see the pending connections and run with MCP tools — an earlier revision of this PR moved restore after `onStart()` and traded that away; the persisted-capability seed removes the tradeoff. An end-to-end test replays the wake sequence and pins the ordering (mutation-tested: reordering restore after recovery fails it).

`addMcpServer` reuses the stored server id when re-adding a known server (`requestedId ?? existingServer?.id ?? nanoid(8)`), keeping registration idempotent during the onStart window and preserving OAuth token storage keyed by server id across hibernation.

Two windows remain by construction: an elicitation request arriving after a restored connection connects but before `onStart()` configures handlers (including during a recovered turn) is answered with a JSON-RPC error — servers handle elicitation failure per spec, and this equals main's behavior for form mode. And the first wake after handlers are first introduced has no stored record yet, so restored connections advertise nothing for that boot; the record is written during that wake and every later wake is correct.

## Implementation notes

- `MCPClientConnection.configureElicitationHandler` rebuilds the SDK client for not-yet-connected connections because SDK `registerCapabilities` is merge-only — a rebuild is the only way to un-advertise a mode when handlers are cleared before connecting (evaluated and rejected the `registerCapabilities` approach for exactly this reason). `init()` re-entry (mid-session 401 → OAuth → reconnect) also rebuilds, so reconnects genuinely refresh advertised capabilities.
- `migrateServerId` rescopes handlers to the new server id; when the manager has no handlers it leaves the connection untouched so an id migration cannot wipe a restored capability seed (mutation-tested).
- RPC elicitation tests use the connection-level `elicitationHandlers` option; conformance host and the mcp-client example configure via `onStart()`.

## Tests

Full agents suite (114 files / 2,267 tests), full MCP suite, `npm run check`, and all conformance jobs green. New coverage: seed precedence (seed vs handlers vs explicit declaration), capability persistence on stored rows, restore-then-configure across a simulated hibernation (the fiber-recovery window), a live RPC connection seeded at connect with handlers attached afterwards (full elicitation round-trip), and the wake-ordering test above. Verified end-to-end against the running `mcp-client` + `mcp-elicitation` examples: both form- and url-mode round-trips complete through `configureElicitationHandler`.

### DIFF
--- a/docs/agents/mcp-client.md
+++ b/docs/agents/mcp-client.md
@@ -113,20 +113,24 @@ These options are persisted and used when reconnecting after hibernation or afte
 
 ### Elicitation
 
-MCP servers can request input from the client during a tool call (`elicitation/create`). To respond, override `onElicitRequest` on your agent:
+MCP servers can request input from the client during a tool call (`elicitation/create`). To respond, configure an elicitation handler before MCP connections are registered or restored:
 
 ```typescript
+import { Agent } from "agents";
+
 class MyAgent extends Agent<Env> {
-  async onElicitRequest(request: ElicitRequest, serverId: string) {
-    // e.g. deliver a url-mode elicitation link out-of-band
-    return { action: "accept" as const, content: {} };
+  onStart() {
+    this.mcp.configureElicitationHandler(async (request, serverId) => {
+      // e.g. deliver a url-mode elicitation link out-of-band
+      return { action: "accept" as const, content: {} };
+    });
   }
 }
 ```
 
-Because it is a class method, the handler survives Durable Object hibernation and applies to connections restored from storage.
+The Agent lifecycle runs `onStart()` before automatic MCP connection restore, so handlers configured there apply to connections restored from storage. Configuring a handler after an MCP connection is already active updates the in-memory handler, but the server only sees new advertised elicitation modes after that connection reconnects.
 
-Overriding `onElicitRequest` is all you need: when a handler is present, connections automatically advertise both form- and url-mode elicitation (MCP spec 2025-11-25 — url-mode is used for sensitive flows like OAuth URLs) at the `initialize` handshake. Without a handler, connections advertise no elicitation capability, so spec-compliant servers use their non-elicitation fallbacks instead of sending requests the agent cannot answer.
+When a handler is present, connections automatically advertise both form- and url-mode elicitation (MCP spec 2025-11-25 — url-mode is used for sensitive flows like OAuth URLs) at the `initialize` handshake. Without a handler, connections advertise no elicitation capability, so spec-compliant servers use their non-elicitation fallbacks instead of sending requests the agent cannot answer.
 
 To narrow the advertised modes, declare them explicitly — an explicit declaration always wins and is persisted with the server options, surviving hibernation:
 
--- a/docs/agents/mcp-client.md
+++ b/docs/agents/mcp-client.md
@@ -134,7 +134,7 @@ class MyAgent extends Agent<Env> {
 }
 ```
 
-The Agent lifecycle runs `onStart()` before automatic MCP connection restore, so handlers configured there apply to connections restored from storage. Configuring a handler after an MCP connection is already active updates the in-memory handler, but the server only sees new advertised elicitation modes after that connection reconnects.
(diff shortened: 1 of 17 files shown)


docs/agents/mcp-client.md
```diff
@@ -113,20 +113,24 @@ These options are persisted and used when reconnecting after hibernation or afte
...
-  async onElicitRequest(request: ElicitRequest, serverId: string) {
-    // e.g. deliver a url-mode elicitation link out-of-band
-    return { action: "accept" as const, content: {} };
+  onStart() {
+    this.mcp.configureElicitationHandler(async (request, serverId) => {
+      // e.g. deliver a url-mode elicitation link out-of-band
+      return { action: "accept" as const, content: {} };
+    });
   }
 }
```
mattzcarey: Could we have different handlers for url and form would that make sense. Like grouped tho id like configure callback has custom something. I only want one function tho configureElicitationHandler

examples/mcp-client/src/server.ts
```diff
@@ -27,6 +27,10 @@ export class MyAgent extends Agent {
   >();
 
   onStart() {
+    this.mcp.configureElicitationHandler((request, serverId) =>
+      this.handleElicitationRequest(request, serverId)
```
mattzcarey: Why are we still using this? Mont we want to deprecate this?

PR #1935, opened 2026-07-13
Bound Agent Workflow tracking so that local state is automatically pruned

ericclemmons:
**tl;dr – Automatically prune terminal Agent Workflow tracking rows from local SQLite.**

Agent Workflow tracking currently grows without bound even after the Workflow platform expires instance state. This persists normalized success/error retention, derives an indexed `expires_at` after the terminal outcome is known, and maintains one earliest-expiry cleanup alarm per Agent or facet owner.

Cleanup is armed when `runWorkflow()` first writes tracking state, reconciles missed terminal callbacks after eviction or crashes, deletes expired local rows, and stops when no cleanup remains. Omitted retention uses a bounded 30-day local fallback because the SDK cannot detect Workers Free versus Paid; this improves the existing unbounded behavior.

This deletes only Agent-local SQLite tracking. It never deletes Workflow platform instances, state, or logs.

---

References: [retention passthrough PR](https://github.com/cloudflare/agents/pull/1948), [Agents cleanup strategy](https://developers.cloudflare.com/agents/runtime/execution/run-workflows/#cleanup-strategy)

### DIFF
--- a/.changeset/workflow-tracking-retention.md
+++ b/.changeset/workflow-tracking-retention.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
+Automatically bound Agent Workflow tracking rows with terminal-outcome retention and alarm-driven cleanup that survives idle Agent eviction.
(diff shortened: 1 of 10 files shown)


ericclemmons:
Closing to redo this using https://github.com/cloudflare/agents/pull/1897

PR #1947, opened 2026-07-14
feat(think): make platform tools native to code

### LINKED ISSUE #1938: Think repeatedly compiles large MCP tool catalogs into Zod schemas and exceeds Durable Object memory limits
**Describe the bug**

When running a Think instance with a Code Mode MCP executor connected through an environment binding to another `McpAgent` Worker, the agent times out and can use more than 300 MB of memory when the MCP server exposes more than 300 tools with complex Zod/JSON schemas.

The MCP connection is intended to be consumed through Code Mode. The individual MCP tools are not intended to be exposed directly to the language model. However, merely keeping the MCP server registered with the Think instance is enough to trigger the high-memory behavior when a message starts a turn.

CPU and heap profiles show that the process spends most of the active turn in garbage collection while Zod schemas are recursively created and cloned. In the captured reproduction, the Worker starts from a relatively stable idle heap, then allocates rapidly as soon as a message is submitted. The process eventually reaches the heap limit before the model can complete the turn.

This is reproducible with a Durable Object MCP binding and does not require executing an MCP tool. Starting a normal chat turn is sufficient.

**To Reproduce**

1. Create a Think-based chat agent.
2. Register an `McpAgent` server through a Durable Object environment binding.
3. Have the MCP server expose more than 300 tools with complex nested input and output schemas.
4. Create a Code Mode tool that uses the live MCP connection.
5. Keep direct MCP tools unavailable to the model, for example by filtering them from `activeTools`.
6. Send a normal user message.
7. Observe memory, garbage collection, and turn completion.

**Expected behavior**

- The Code Mode MCP connection should remain usable with a large, valid tool catalog.
- A normal chat turn should complete within the Durable Object memory limit.
- Memory should return to a stable level after the turn completes or fails.
- A failed turn should recover without entering repeated memory-limit failures.

**Screenshots**

Image

**Version:**

- `agents`: `0.17.3`
- `@cloudflare/think`: `0.12.1`
- `@cloudflare/ai-chat`: `0.9.3`
- `@cloudflare/codemode`: `0.4.2`
- MCP server: Durable Object binding to an `McpAgent`
- MCP catalog: more than 300 tools with complex nested schemas

**Additional context**

| Measurement | Observed value |
| --- | ---: |
| Idle V8 heap before sending | 89.91 MiB |
| MCP tools | 313 |
| Cached MCP output validators | 311 |
| MCP manager retained heap before the turn | 4.02 MiB |
| Active CPU samples spent in GC | 83.85% to 88.37% |
| Non-GC JavaScript dominated by `getAITools` / `fromJSONSchema` | approximately 91% |
| Local heap at fatal OOM | approximately 1.07 GB |
| Deployed Durable Object memory after trigger | approximately 275 MB |

mattzcarey:
## Summary

- make Think's built-in `code` tool a durable Code Mode runtime while keeping `read`, `write`, `edit`, and `code` as the four built-in model tools
- expose Think platform capabilities inside `code` as `workspace.*`, `context.*`, `skills.*`, `extensions.*`, `fetch.*`, and one namespace per connected MCP server
- separate MCP connection registration from direct model-tool exposure: Think no longer calls `mcp.getAITools()` while assembling a turn
- cache explicit `MCPClientManager.getAITools()` conversion by live connection, server id, and catalog revision, without scattered invalidation branches
- make Code Mode runtime ownership explicit: applications using custom `bash`, custom `code`, or `createExecuteTool(this)` set `readonly codeTool = false`
- add Worker Loader defaults, migration docs, an updated coding starter, and a focused `examples/think-codemode-mcp` app

## Why

Issue #1938 reproduces with a 313-tool MCP server. A normal Think turn currently calls `getAITools()` before `activeTools` filtering, recursively converting 626 input/output JSON Schemas into Zod graphs even when the application intends to use MCP only through Code Mode.

The regression registers 313 complex MCP tools, replaces `getAITools()` with a throwing/counting trap, and completes this flow through the real Durable Object transport and Dynamic Worker runtime:

1. `codemode.search()` finds `catalog.echo`
2. `codemode.describe()` generates its types on demand
3. `catalog.echo()` executes over MCP RPC
4. the model still receives only four built-in schemas
5. `getAITools()` is called zero times

A separate MCP manager regression verifies that filtered and unfiltered `getAITools()` calls share converted schemas. Replacing the connection, changing its id, or replacing its catalog causes conversion exactly once for the new revision.

## Compatibility boundary

This is a pre-1.0 minor change for `@cloudflare/think`:

- the built-in tool is named `code` and accepts `{ code: string }`; it does not accept POSIX shell syntax or the legacy `{ script, cwd }` payload
- `readonly codeTool = false` opts out of the built-in runtime and restores the direct workspace, context, skills, extensions, and fetch layout
- custom `bash`, custom `code`, and `createExecuteTool(this)` do not opt out implicitly; Think rejects competing runtime owners with an actionable error
- legacy `workspaceBash` configuration moves to `codeTool` or an explicit `createWorkspaceTools()` tool set
- application `getTools()`, Think actions, and client tools remain direct because they have separate lifecycle and approval semantics
- the built-in runtime requires `worker_loaders: [{ binding: "LOADER" }]`; generated Think config includes it, and custom Worker entries export `CodemodeRuntime` through `@cloudflare/think/server-entry`
- the standalone `createWorkspaceTools().bash` remains a real `just-bash` POSIX shell

## Structural audit

- replaced the MCP manager's server-id cache and reconnect/rename/close invalidation branches with a `WeakMap` keyed by the live connection
- removed runtime-handle identity inference from `Think`; `codeTool` is the sole ownership policy
- moved the Workspace filesystem capability guard into `@cloudflare/shell` and deleted Think's duplicate adapter module
- moved namespaced StateConnector type rendering into Shell's prompt module
- replaced Think's string rewrite of the Agent Skills prompt with an explicit activation-guidance seam
- rebuilt the native runtime after hibernation using the same MCP wait policy as inference

## Verification

- `pnpm run build`: all 26 targets passed
- `pnpm run check`: exports, formatting, lint, and all 118 TypeScript projects passed
- `pnpm exec nx affected -t test --base=origin/main`: all 19 affected projects passed
- Agents: 2,280 tests passed
- AI Chat: 728 tests passed
- Code Mode: 373 tests passed
- Shell: 251 passed, 1 skipped
- Think: 923 Worker, 10 generated-entry, 10 Vite, 55 CLI, and 5 React tests passed
- `agent-think`: 37 tests passed
- focused MCP example production build passed

Fixes #1938


mattzcarey:
I hate this

PR #1949, opened 2026-07-15
fix(think): warn when skills hide system prompt fallback

cjol:
This PR documents `getSystemPrompt()` as a fallback-only API and adds a warning for the confusing `getSystemPrompt()` plus `getSkills()` combination. Addresses #1871.

## Why

- `getSkills()` registers the skills catalog as a Session context block, so Think correctly treats `getSystemPrompt()` as fallback-only once skills are present.
- That behavior is easy to miss because an agent can override both APIs and still run, but its `getSystemPrompt()` instructions will not reach the model.
- We could merge `getSystemPrompt()` with context blocks, but that changes behavior for existing agents that use Session context as the system prompt source.
- The safer path is to keep the current prompt assembly semantics, document them clearly, and warn when a likely footgun is detected.

## Code Changes

- Updates `getSystemPrompt()` JSDoc and the Think README to describe it as a legacy fallback used only when no Session context blocks are configured.
- Emits a deduped `[think]` warning when `getSkills()` returns sources and the agent overrides `getSystemPrompt()`.
- Adds a regression test agent that reproduces the `getSkills()` plus overridden `getSystemPrompt()` setup and asserts the warning.
- Adds a patch changeset for `@cloudflare/think`.

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/1949" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


mattzcarey (approved):

mattzcarey:
looks good but I want to fix this properly

PR #1993, opened 2026-07-29
fix(observability): bound top-level approval spans

Ankcorn:
Follow-up to #1982.

AI SDK v7 top-level `toolApproval` policies were not receiving the invocation lifetime policy, so asynchronously-decided approval spans could be emitted as complete after their invocation had already ended. Pass the existing boundary through and cover the escaped-invocation path with the real v7 SDK.


mattzcarey:
close as superseeded

PR #2002, opened 2026-07-30
feat(think): add experimental Computer workspace

cjol:
This PR adds an opt-in `@cloudflare/think/experimental/computer` entrypoint for using Cloudflare Computer as a Think workspace. Existing Think agents continue to use `@cloudflare/shell` and `just-bash` unless they opt in. It consumes the shared client compatibility added by cloudflare/computer#26.

## Why

- Computer provides a durable filesystem plus native shell execution, while the current Think workspace runs commands through the in-process `just-bash` fallback.
- Replacing the default workspace would change execution behavior and hide existing workspace data. There is no migration between the legacy `@cloudflare/shell` tables and Computer `vfs_*` tables.
- The integration therefore lives behind an experimental export and optional peer dependency, so each Think class can opt in independently.
- Local and remote Computer clients now share the upstream `useThink` compatibility surface. Keeping a second filesystem adapter in Think would duplicate path, stat, and missing-file translations and allow the two implementations to drift.

## Public API Surface

All additions are under the new experimental export.

| Symbol | Kind | Notes |
| --- | --- | --- |
| `createComputerWorkspace` | Function | Constructs a locally owned Computer with `useThink: true` |
| `adaptComputer` | Function | Validates and narrows a compatible `getWorkspace()` client without wrapping it |
| `ComputerWorkspace` | Type | A Computer client intersected with Think `WorkspaceLike` |
| `LocalComputerWorkspace` | Type | A local Computer intersected with Think `WorkspaceLike` |
| `Computer` | Class re-export | Alias for the current Computer `Workspace` class |
| `getWorkspace` | Function re-export | Resolves local and remote Computer clients |
| `withWorkspace` | Mixin re-export | Adds Computer ownership to another Durable Object class |
| `WorkspaceServiceProxy` | Worker entrypoint re-export | Connects the Worker shell backend to the owning Think instance |
| `WorkerShellBackend` | Class re-export | Worker Loader shell backend |
| `ComputerOptions` | Type | Options accepted by `createComputerWorkspace` |
| `WorkspaceClient` | Type | Local or remote Computer client type |
| `WorkerShellBackendOptions` | Type | Worker shell backend configuration |
| `WorkerShellFetcher` | Type | Fetcher contract used by the Worker backend |

## Architectural Changes

```text
Think.workspace
├── default: @cloudflare/shell Workspace
│   └── built-in bash tool uses just-bash
└── opt-in override: @cloudflare/computer
    ├── owner enables useThink once
    ├── local and remote clients expose the same Think methods
    ├── native fs, runtime, git, assets, and artifacts stay available
    └── built-in bash tool uses Computer shell exec
```

## Code Changes

- Adds the experimental Computer entrypoint and preserves the native Computer object for both local ownership and remote clients.
- Makes `adaptComputer()` a runtime assertion because the owner-side `useThink` option cannot be inferred across RPC. It returns the original client rather than a compatibility wrapper.
- Teaches `createWorkspaceTools()` to use a workspace native shell when available while retaining the existing `just-bash` fallback.
- Adds the host bridge that lets `WorkspaceServiceProxy` obtain the stub for a Computer owned by the Think Durable Object.
- Adds the optional Computer peer, build entrypoint, package export, changeset, and user documentation.

## Compatibility

- This is an additive minor change. Think default workspace, direct workspace methods, stored data, and `just-bash` execution remain unchanged.
- Opting in starts with a separate Computer workspace. Existing legacy files remain untouched but are not visible through Computer, and Computer files are not visible after switching back.
- Consumers importing the experimental entrypoint need `@cloudflare/computer@>=0.1.1 <0.2.0`.
- The development dependency uses the published `@cloudflare/computer@^0.1.1` release, which includes the shared Think compatibility surface from cloudflare/computer#26.


cjol:
@aron-cf now that [Computer #26](https://github.com/cloudflare/computer/pull/26) and [workspace #24](https://github.com/cloudflare/workspace/pull/24) have merged, could you publish "@cloudflare/computer 0.1.0-alpha.2" from main? Once that's on npm I can swap the PR preview dep, regenerate lockfile, and undraft this PR.

PR #2005, opened 2026-07-31
fix(agents): initialize cold native RPC calls

### LINKED ISSUE #1990: Agent/Think: an RPC-triggered cold start serves calls against uninitialised state, causing errors and incorrect responses
On a cold Durable Object instance, `submitMessages()` works and `saveMessages()` throws. Both are public RPC methods on the same object.

The difference is that `submitMessages` persists its work and defers execution to an alarm that runs `onStart()`, while `saveMessages` assumes that `onStart()` has already run. Native DO RPC never triggers `onStart`. partyserver runs `onStart` from `fetch`, `alarm`, the `webSocket*` handlers, and its own `setName` / `__unsafe_ensureInitialized` RPCs, but not from anything else. So any public or user-defined RPC method that touches state that `onStart` establishes fails whenever it is the call that wakes the instance, which is the normal case for an agent that goes idle between messages, due to DO eviction policies.

For Think the missing state is `session`, `workspace`, the hydrated message view, protocol handlers and recovery. `chat`, `runTurn`, `getMessages` and the history mutators all depend on this state. `Agent.onStart` also restores MCP connections and drives fiber and agent-tool recovery, and a subclass's own `onStart` hasn't run either.

## Reproduction

**package.json**

```json
{
  "name": "think-cold-rpc-mre",
  "private": true,
  "type": "module",
  "dependencies": {
    "@cloudflare/think": "0.15.0",
    "agents": "0.20.0",
    "ai": "^6.0.0"
  },
  "devDependencies": {
    "wrangler": "^4.114.0"
  }
}
```

**wrangler.jsonc**

```jsonc
{
  "name": "think-cold-rpc-mre",
  "main": "src/index.ts",
  "compatibility_date": "2026-06-11",
  "compatibility_flags": ["nodejs_compat"],
  "durable_objects": {
    "bindings": [{ "class_name": "ColdAgent", "name": "ColdAgent" }]
  },
  "migrations": [{ "tag": "v1", "new_sqlite_classes": ["ColdAgent"] }]
}
```

**src/index.ts**

```ts
import { Think } from "@cloudflare/think";
import type { UIMessage } from "ai";

type Env = {
  ColdAgent: DurableObjectNamespace<ColdAgent>;
};

export class ColdAgent extends Think<Env> {}

export default {
  async fetch(_req: Request, env: Env): Promise<Response> {
    const stub = env.ColdAgent.get(env.ColdAgent.idFromName("mre"));

    // This instance has never run, so the RPC below cold-starts it and
    // this.session is still undefined when saveMessages reaches it.
    const message: UIMessage = {
      id: "m1",
      role: "user",
      parts: [{ type: "text", text: "hi" }]
    };
    await stub.saveMessages([message]);
    return new Response("ok");
  }
} satisfies ExportedHandler<Env>;
```

```sh
npm install
npx wrangler dev --port 8899 &
curl http://localhost:8899/
```

**Observed**:

```
TypeError: Cannot read properties of undefined (reading 'appendMessage')
```

**Expected**: the call behaves as it would on an initialized instance.

The MRE uses a never-woken name for brevity, but the same failure reproduces on an established agent by evicting it and calling any of those entrypoints on the existing stub.

## Other failure modes

The missing state doesn't always result in a throw. Where the missing state is a cache rather than an object, the call succeeds with a wrong answer and no error:

- `getMessages()` returns `[]` — an empty transcript for a conversation that has one.
- `getMcpServers()` reports every registered server as `not-connected`, with an empty tool list.

## Why the caller can't close this

`getAgentByName` does close it, by making one extra RPC (`setName`) that synchronises `onStart`. But that guarantee belongs to the instance it initialised, and stubs outlive instances. Any hibernation, eviction or deployment between acquisition and use can cause the DO belonging to the stub to go cold.

Callers who can't reach the helper have it worse. `getAgentByName` needs a parameterised namespace, so cross-script bindings and stubs received over RPC are left hand-rolling the handshake as `stub.setName(name)`, which is deprecated, even though `getServerByName` relies on it.

## What I'd like

Ideally I'd like a solution that requires no per-method boilerplate. Initialisation on cold start is a framework responsibility, and every entrypoint the framework owns already upholds it. `agents` opens its own internal RPC entrypoints with `__unsafe_ensureInitialized()`. Application-defined entrypoints are the gap.

I can live with a solution that requires some level of boilerplate, but that would erode some of the benefit that I expected from Think. I decided to trial Think mostly because I didn't want to have to handle eviction and cold starts in my application code.

I'd be happy to work on a PR for this, but I'd like to know what type of solution the maintainers would be willing to accept before I start, as I suspect that fixing this properly may require some breaking changes, and may require enforcing some restrictions on the public API surface (e.g. everything must be async).

cjol:
This PR runs lifecycle startup before a native Durable Object RPC method executes on a cold `Agent` or `Think` instance. Fixes #1990. Rebased onto the vendored lifecycle from #2133.

## Why

- Native RPC dispatches straight onto the instance. The lifecycle starts from the five handlers it installs (`fetch`, `alarm`, `webSocketMessage`, `webSocketClose`, `webSocketError`) and from an explicit `lifecycle.start()`. An RPC method is none of those, so `onStart` never runs when the RPC is the call that wakes a cold instance. For an agent that idles between messages, that is the normal case.
- Two symptoms follow. An empty `class ColdAgent extends Think {}` with a single `stub.saveMessages([message])` throws `TypeError: Cannot read properties of undefined (reading 'appendMessage')`, because `session` is declared `session!: Session` and assigned during startup. The quieter symptom is worse: `getMessages()` returns `[]` for a conversation that has one, and `getMcpServers()` reports every registered server as `not-connected`, with no error to distinguish those from a genuinely empty result.
- `Agent` already treats initialization as an entry-point responsibility, calling `__unsafe_ensureInitialized()` from `_cf_invokeAgentPath`, `_cf_initAsFacet`, `_cf_scheduleDestroy` and the three `_workflow_*` methods. All six are framework-internal. The surface callers actually dial is the gap.
- `docs/agents/lifecycle.md` prescribes `await this.lifecycle.start()` for RPC methods that need startup, and that works. It cannot reach the reported bug, because the repro is an empty subclass calling a framework method on the stub and there is no user code in the path. The only workaround is overriding framework methods purely to inject the line, on signatures the subclass does not own.
- Calling `lifecycle.start()` from every public method was the other option. `Agent` has 82 public methods and `Think` has 60; around 40 are synchronous and would have to become async, breaking every subclass that overrides `getModel()` or `getTools()`. It is also opt-in, so a public method added later silently misses the call. Widening the wrapping `Agent` already performs preserves those signatures and fails closed instead.
- Full rationale, including the alternatives rejected and the costs accepted, is in `design/rfc-cold-rpc-initialization.md`.

## Public API Surface

| Symbol | Kind | Notes |
| --- | --- | --- |
| `Agent.__unsafe_ensureInitialized(props?)` | Changed behaviour | Returns early unless the instance is constructed but unstarted, so startup cannot re-enter itself. Signature unchanged. |
| `Agent._rpcInitializationState` | Added, protected | Native RPC lifecycle state. Reaches emitted types because it is `protected`. Not intended for subclass use. |
| `LifecycleHostCallback` | Added, internal | Exported from `lifecycle/durable-object-lifecycle.ts` to derive the exclusion boundary. Not re-exported from `agents/lifecycle`, so not public API. |

## Architectural Changes

The initialization boundary moves from the framework's own RPC entry points to the whole public RPC surface.

```
before                                  after
  fetch / alarm / webSocket*              fetch / alarm / webSocket*
    -> lifecycle startup                    -> lifecycle startup
  _cf_* / _workflow_*                     _cf_* / _workflow_*
    -> lifecycle startup                    -> lifecycle startup
  public + user-defined methods           public + user-defined methods
    -> no startup                           -> lifecycle startup when cold
```

`Agent` already replaces prototype methods at construction to establish `getCurrentAgent()` context, guarded by a module-level set so it runs once per class. That traversal previously excluded everything on `Agent.prototype`, so it only wrapped user-defined methods. Removing that exclusion brings the framework's own API into scope, and the wrapper gains one branch:

```ts
if (this._rpcInitializationState !== "pending") {
  return method.apply(this, args);
}
return this.__unsafe_ensureInitialized().then(() => method.apply(this, args));
```

Only a constructed-but-unstarted instance defers. A warm call is unchanged.

### `packages/agents`

- A four-state field (`constructing`, `pending`, `starting`, `started`) tracks the boundary. The constructor arms `pending` in a microtask under `blockConcurrencyWhile`, so a subclass calling its own methods during construction stays synchronous, and the runtime cannot dispatch an external event before the callback settles.
- `onStart` sets `starting` then `started`, restoring the previous value on failure so a failed startup can be retried. The `starting` state stops startup's own calls into public methods from recursing.
- The exclusion boundary is assembled in three parts, since it can no longer be derived from a base prototype now that `Server` is gone. Platform methods come from `DurableObject.prototype`. Lifecycle callbacks come from `LifecycleHostCallback` via `satisfies Record<LifecycleHostCallback, true>`, so adding a callback to the lifecycle without excluding it fails the build. Runtime entry points, lifecycle accessors, connection policy hooks, state hooks, `destroy()` and underscore-prefixed names are listed explicitly.
- `destroy()` is excluded so a cold alarm can finish deleting a condemned Agent without startup recreating the state that teardown then erases.
- Connection policy hooks (`shouldConnectionBeReadonly`, `shouldSendProtocolMessages`, `isConnectionProtocolEnabled`) are excluded because they make synchronous decisions inside framework entry points and must not return a Promise.
- Traversal resolves to the nearest overriding descriptor with no depth limit, so a subclass override wins over an inherited definition and deep hierarchies are not silently truncated. Getters, setters and non-function properties are skipped by descriptor, `callableMetadata` is copied onto the wrapper so `@callable` survives, and a `WeakSet` prevents double-wrapping when two classes in a hierarchy are both constructed.
- The explicit `_cf_*` cross-DO RPC boundaries initialize as well. Cloudflare's facet documentation states that a facet may hibernate and be resumed by `ctx.facets.get()` while the supervisor is handling a request, so facets do not share the host's in-memory lifetime.

### `packages/think`

- Think's `onStart` assignment drives the same state transitions, preserving the existing startup recursion guard so `onStart` runs exactly once.
- The host bridge methods (`_hostReadFile`, `_hostGetMessages` and the rest) call `__unsafe_ensureInitialized()` directly. They are underscore-prefixed, so wrapping deliberately skips them, and `HostBridgeLoopback` reconstructs the Agent stub from its string id.

### Fixtures

- Two MCP OAuth tests moved from `newUniqueId()` to `idFromName()`. Those instances were never addressable by name, so their first RPC now fails startup rather than running uninitialized. This is the supported setup rather than a concealed regression, and it is the same change consumers on raw ids need to make.

## Compatibility

- An Agent addressed with `newUniqueId()` or `idFromString()` now fails its first native RPC with the lifecycle's addressing error, where it previously served the call against uninitialized state. Address Agents by name with `getAgentByName()`, `getByName()` or `idFromName()`.
- A declared-synchronous public method returns a Promise when its call is the one that wakes a cold instance. Over RPC this is invisible, since every call is awaited. An in-process caller on that path sees a Promise while TypeScript still shows the synchronous signature.


threepointone:
I'll review this one

threepointone:
I checked this against current `main`. The bug is still there: a native RPC call (`stub.method()`) on a cold instance runs without `onStart`.
- `withAgentContext` sets up context but does not start the agent.
- Startup runs only from the lifecycle's `fetch`, `alarm` and `webSocket*` handlers, or from the few internal callers of `__unsafe_ensureInitialized`.
- Think's `saveMessages` and `getMessages` still do not initialize.

The PR conflicts heavily with `main`: it is 79 commits behind, with 22 conflict hunks in `index.ts`. State, WebSockets, sessions, facets and the queue have all moved into capabilities since, so many of the `_cf_*` insertions and the method exclusion list target code that has moved. I would rework it rather than rebase it:

1. **Use the lifecycle's own started state.** Check a small public "is started" method on the lifecycle instead of adding a second four-state machine armed from a `blockConcurrencyWhile` microtask in the constructor.
2. **Start inside `withAgentContext`.** Start the lifecycle there only when the call comes from outside an agent context. That covers every wrapped method without wrapping `Agent.prototype` behind a hand-kept exclusion list.
3. **Keep synchronous methods synchronous.** Today, the first call that wakes the instance makes a sync method such as `getMcpServers()` return a Promise. Devin flagged this; widening the return type avoids the type error but does not fix the behavior. With (2), sync base methods can stay unwrapped.
4. **Ship the unnamed-Agent change as a minor.** An RPC on a `newUniqueId()` / `idFromString()` Agent now throws, which needs to be called out in the changeset.

Your tests are the right acceptance suite and should carry over: exactly-once startup, sync policy hooks, `destroy` on a cold instance, and Think hydration after eviction.

PR #2006, opened 2026-07-31
fix: reuse sub-agent broadcast bridges

Ankcorn:
A streamed sub-agent chat produced 1,019 `_cf_broadcastToSubAgent` and 882 `setName` JSRPC calls because broadcasts that outlived their WebSocket handler resolved the root Agent for every chunk.

Retain the connection bridge with `dup()` and reuse that capability after the handler returns. This keeps broadcasts on the existing root/facet RPC session instead of creating root DO subrequests, while preserving root resolution as the cold-wake fallback. Retained bridges are replaced on reconnect and disposed on replacement or close.

### DIFF
--- a/.changeset/quiet-agents-broadcast.md
+++ b/.changeset/quiet-agents-broadcast.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
+Reuse retained sub-agent connection bridges for broadcasts that continue after a WebSocket handler returns, avoiding root Agent resolution and Durable Object RPC requests for every streamed chunk.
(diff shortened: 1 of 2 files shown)


Ankcorn:
Local workerd + streaming-tail reproduction found an important constraint: retained reverse-RPC bridge calls avoid `durable_object_subrequest` spans, but the originating invocation remains open until the duplicated bridge is disposed. Retaining it for the WebSocket lifetime could prevent invocation completion/hibernation, so this draft needs reworking before review.

Ankcorn:
Closing this approach after validating it with a local workerd + streaming-tail reproduction. Retaining the reverse-RPC bridge removes DO subrequests, but keeps the originating invocation open until disposal and could interfere with hibernation/recovery. We need a broader, lifecycle-bounded design.

PR #2016, opened 2026-08-03
fix(worker-bundler): normalize npm tarball roots

### LINKED ISSUE #2008: @cloudflare/worker-bundler: installDependencies writes DefinitelyTyped packages one directory too deep, silently — tarball root is assumed to be `package/`
**Describe the bug**

`installDependencies` strips the root directory of an npm tarball only when that directory is literally named `package/`:

https://github.com/cloudflare/agents/blob/0efd545a58c9075885977627e5d853b6e98f6d54/packages/worker-bundler/src/installer.ts#L440-L444

`package/` is a convention of `npm pack`, not a rule of the registry. DefinitelyTyped roots its tarballs at the package's bare name instead:

```console
$ curl -sL "$(curl -s https://registry.npmjs.org/@types/react-dom \
    | python3 -c 'import json,sys;d=json.load(sys.stdin);v=d["dist-tags"]["latest"];print(d["versions"][v]["dist"]["tarball"])')" \
    -o rd.tgz && tar tzf rd.tgz | head -5
react-dom/
react-dom/LICENSE
react-dom/README.md
react-dom/canary.d.ts
react-dom/client.d.ts
```

So for every `@types/*` package the root is never stripped and each file is written one directory too deep:

```
node_modules/@types/react-dom/react-dom/index.d.ts   ← actual
node_modules/@types/react-dom/index.d.ts             ← expected
```

TypeScript then resolves nothing from the package.

**The failure is silent.** `installDependencies` returns the package in `installed` and `warnings` is empty:

```
installed: ["@types/react-dom@19.2.4","@types/react@19.2.18","react@19.2.8",
            "csstype@3.2.3","react-dom@19.2.8","scheduler@0.27.0"]
warnings:  []
fs.read("node_modules/@types/react/package.json")  →  null
```

Nothing reports a problem. The only symptom appears later, from `createTypescriptLanguageService`:

```
TS7016: Could not find a declaration file for module 'react'.
        '/node_modules/react/index.js' implicitly has an 'any' type.
        Try `npm i --save-dev @types/react` if it exists or add a new declaration (.d.ts)
        file containing `declare module 'react';`
```

That message reads as though the package genuinely ships no types, and it invites you to send a pull request to DefinitelyTyped. `@types/react` *is* installed and `installed` says so, so the message points away from the actual cause. It cost us the better part of a day before we ran `tar tzf` on the tarball.

This matters because it breaks the composition the README documents — [“### With `installDependencies`”](https://github.com/cloudflare/agents/blob/main/packages/worker-bundler/README.md), installing type packages into a filesystem so the language service has something to check against. `@types/*` is exactly what you install for that, and it is exactly the family of packages that does not use a `package/` root. The two headline features of the package do not compose for any DefinitelyTyped package.

**To Reproduce**

Inside workerd (`@cloudflare/vitest-pool-workers`), against the real registry:

```ts
import { InMemoryFileSystem, installDependencies } from "@cloudflare/worker-bundler";

const fs = new InMemoryFileSystem({
  "package.json": JSON.stringify({
    devDependencies: { "@types/react": "^19" }
  })
});

const result = await installDependencies(fs, { dev: true });

console.log(result.installed);                                  // ["@types/react@19.2.18"]
console.log(result.warnings);                                   // []
console.log(fs.read("node_modules/@types/react/package.json")); // null  ← installed, unusable
console.log(fs.list("node_modules/@types/react/").slice(0, 2));
// [ "node_modules/@types/react/react/LICENSE",
//   "node_modules/@types/react/react/README.md" ]   ← one directory too deep
```

**Expected behavior**

Files land at `node_modules/@types/react/index.d.ts`, and the language service resolves `import { useRef } from "react"` with no diagnostics.

Failing that, at minimum the install should not report success silently — but the right answer is to strip whatever single root directory the archive has, rather than matching one name.

**Full end-to-end reproduction**

**https://github.com/apostolos-geyer/worker-bundler-types-repro** — a small alchemy project that runs the Worker locally under workerd (`dev: true`, so it needs no Cloudflare account) and prints the before/after side by side against the real registry:

```
bun install
bun test ./integ/
```

```
before (as installed by 0.2.2):
  installed:              @types/react-dom@19.2.4, @types/react@19.2.18, csstype@3.2.3, react-dom@19.2.8, react@19.2.8, scheduler@0.27.0
  warnings:               []
  @types/react landed at: node_modules/@types/react/react/LICENSE
  @types/react/package.json readable: false
  TS7016 src/App.tsx:1     Could not find a declaration file for module 'react'.
  TS7016 src/App.tsx:5     Could not find a declaration file for module 'react/jsx-runtime'.
  TS7016 src/client.tsx:1  Could not find a declaration file for module 'react-dom/client'.
  TS7016 src/client.tsx:4  Could not find a declaration file for module 'react/jsx-runtime'.

after  (archive root stripped):
  @types/react landed at: node_modules/@types/react/LICENSE
  @types/react/package.json readable: true
  (no diagnostics)
```

The same project also checks that stripping the root does not just silence the checker: with `<input maxLength="ten" />` in the source, `TS2322 Type 'string' is not assignable to type 'number'` is still reported.

**Version**

`@cloudflare/worker-bundler@0.2.2`, and still present on `main` at 0efd545.

**Additional context**

Tar entries in an npm tarball share a single root component, so the root can be found rather than assumed. The cases worth being careful about are entries that are already flat, non-file entries whose names are not paths (`pax_global_header`, GNU `@LongLink`), directory entries, and archives whose entries genuinely do not share one root — those last ones must be left alone rather than half-stripped.

I have a patch and will open a PR shortly.

ben-reitz:
## Summary

**Context:** `installDependencies` assumed every npm tarball was rooted at `/package`, so some npm packages (such as DefinitelyTyped ones like `@types/react`) were installed one directory too deep and silently became unusable. This change detects and strips a tarball's shared root directory if it has one, whether it's called `package` or not.

### Changes

- infer and remove the directory shared by npm tarball file entries instead of assuming it is named `package`
- reconstruct POSIX USTAR paths from their prefix and name fields before root detection
- preserve flat and multi-root archives while ignoring tar metadata during root detection
- add synthetic tarball coverage for DefinitelyTyped, conventional, USTAR-prefixed long-path, leading-`./`, flat, and mixed archive layouts
- add a patch changeset for `@cloudflare/worker-bundler`

#### Before: only remove a root named package/

 ```text
   is-number tarball:
   package/index.js
           │ strip package/
           ▼
   node_modules/is-number/index.js              ✅

   @types/react tarball:
   react/index.d.ts
           │ react/ not recognized
           ▼
   node_modules/@types/react/react/index.d.ts    ❌ one level too deep
 ```

 #### After: detect and remove the actual shared root

 ```text
   is-number:
   package/index.js
           │ detected root: package/
           ▼
   node_modules/is-number/index.js               ✅

   @types/react:
   react/index.d.ts
           │ detected root: react/
           ▼
   node_modules/@types/react/index.d.ts           ✅
```

## Testing

- `pnpm --filter @cloudflare/worker-bundler test`
- `pnpm run check`

Fixes #2008

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/2016" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


threepointone (approved):
lgtm

PR #2022, opened 2026-08-03
Keep hibernation enabled with partial Agent options

### LINKED ISSUE #2012: Setting any `static options` on an Agent subclass silently disables hibernation
ai slop but still true:

**Describe the bug**

Setting any `static options` on an `Agent` subclass silently disables hibernation.

`Agent.options` defaults `hibernate: true`, and the docs state that *"Child classes can override specific options without spreading"* (`dist/index.js:285`, and in the shipped `.d.ts:3050`). But a subclass static **replaces** the inherited object rather than merging, and `partyserver` selects the connection manager from the raw static with no defaulting:

```js
// partyserver@0.5.9 dist/index.js:546,549
static options = { hibernate: false };
#ParentClass = Object.getPrototypeOf(this).constructor;
#connectionManager = this.#ParentClass.options.hibernate
  ? new HibernatingConnectionManager(this.ctx)
  : new InMemoryConnectionManager();
```

So `static options = { sendIdentityOnConnect: false }` leaves `hibernate === undefined` → `InMemoryConnectionManager`. Meanwhile `_resolvedOptions` (`agents dist/index.js:461`) *does* apply the default, so the SDK still reports `hibernate: true`. Two layers read the same static; only the one that doesn't pick the connection manager defaults it.

This is especially easy to hit because the SDK itself recommends the offending line. `agents dist/index.js:957` warns:

> `[Agent] ${ctor.name}: sending instance name "${this.name}" to clients via sendIdentityOnConnect ... If this name is sensitive, add `static options = { sendIdentityOnConnect: false }` to opt out.`

Following that advice, to fix a stated privacy concern, silently turns off hibernation.

**To Reproduce**

1. Define an Agent subclass that overrides a single option without spreading, exactly as the docs describe:
   ```js
   class MyAgent extends Agent {
     static options = { sendIdentityOnConnect: false };
   }
   ```
2. Connect a WebSocket to it.
3. Inspect `MyAgent.options.hibernate` — it is `undefined`, so `partyserver` constructs an `InMemoryConnectionManager` and accepts via `connection.accept()` instead of `ctx.acceptWebSocket(...)`.
4. Note that `_resolvedOptions.hibernate` still reports `true`, so nothing surfaces the mismatch.
5. Add `hibernate: true` back into the same object and the `HibernatingConnectionManager` is selected again.

**Expected behavior**

Overriding one option should leave the others at their documented defaults, so `hibernate` stays `true` unless explicitly set to `false`. At minimum, `_resolvedOptions` and the connection-manager selection should agree.

**Screenshots**

N/A.

**Version:**

`agents@0.20.1` with `partyserver@0.5.9`. Also present in `agents@0.19.0` / `partyserver@0.5.8` — the relevant `partyserver` lines are unchanged between the two.

**Additional context**

The impact is silent and billing-related rather than functional: `HibernatingConnectionManager.accept` calls `ctx.acceptWebSocket(...)`, while `InMemoryConnectionManager.accept` calls `connection.accept()`, so the Durable Object stays resident for the life of every WebSocket and bills wall-clock duration while idle. There is no error, and the SDK's own resolved options report the opposite.

Suggested fix: merge subclass statics into the defaults at class-definition time, so the raw `options` object `partyserver` reads is always complete — which is what the docs already promise. A dev-mode warning when `options` is set without `hibernate` would also catch it cheaply.

threepointone:
Fixes #2012.

## What changed

- Resolve the documented hibernation default on the raw subclass static options immediately before PartyServer initializes its connection manager.
- Preserve an explicit hibernate: false override.
- Replace rather than mutate the user-provided options object, which also supports frozen option objects.
- Add a Workers regression test that exercises the existing sendIdentityOnConnect-only subclass and verifies the socket is registered through the Durable Object hibernation API.
- Add an agents patch changeset.

## Reproduction

Before the fix, the regression test observed TestNoIdentityAgent.options.hibernate as undefined after connecting. PartyServer therefore selected its in-memory connection manager even though the Agents resolved options reported hibernation enabled.

## Validation

- focused reproduction failed before the fix and passes after it
- full Agents Workers suite: 91 files, 1,742 tests passed
- affected source and test TypeScript configs passed
- repository formatting and lint passed
- package export validation passed
- sherif passed with the two existing missing-package warnings for examples/harness-deck and examples/think-standalone

No documentation update is needed: the documented default was already correct, and this change makes runtime behavior match it.

### DIFF
--- a/.changeset/tidy-agents-hibernate.md
+++ b/.changeset/tidy-agents-hibernate.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
+Keep WebSocket hibernation enabled when an Agent subclass overrides unrelated static options.
--- a/packages/agents/src/index.ts
+++ b/packages/agents/src/index.ts
@@ -2333,6 +2333,18 @@ export class Agent<
   }
 
   constructor(ctx: AgentContext, env: Env) {
+    // PartyServer selects its connection manager while `super()` initializes
+    // the base-class fields, before this Agent instance exists. Ensure the raw
+    // subclass option it reads has the same hibernation default as our resolved
+    // options. Replacing (rather than mutating) the object also supports frozen
+    // user-provided option objects.
+    const ctor = new.target as typeof Agent;
+    if (ctor.options?.hibernate === undefined) {
+      ctor.options = {
+        ...ctor.options,
+        hibernate: DEFAULT_AGENT_STATIC_OPTIONS.hibernate
+      };
+    }
     super(ctx, env);
 
     this.mcp = this._withAgentSpan(
--- a/packages/agents/src/tests/static-options.test.ts
+++ b/packages/agents/src/tests/static-options.test.ts
@@ -1,0 +1,33 @@
+import { env, runInDurableObject } from "cloudflare:test";
+import { exports } from "cloudflare:workers";
+import { describe, expect, it } from "vitest";
+
+describe("Agent static options", () => {
+  it("keeps hibernation enabled when a subclass overrides another option", async () => {
+    const room = `partial-static-options-${crypto.randomUUID()}`;
+    const response = await exports.default.fetch(
+      `http://example.com/agents/test-no-identity-agent/${room}`,
+      { headers: { Upgrade: "websocket" } }
+    );
+
+    expect(response.status).toBe(101);
+    const client = response.webSocket;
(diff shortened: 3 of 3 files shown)


threepointone:
closing this to fix it in partykit instead

PR #2036, opened 2026-08-04
Fix/1984 synthetic compaction

### LINKED ISSUE #1984: bug: synthetic compaction messages are persisted as real rows, duplicating the summary in every later read
## Summary

After a Think session compacts, the synthetic `compaction_<recordId>` message that `getHistory()` substitutes for the compacted range gets **persisted as a real `assistant_messages` row** on the next turn. From then on `getHistory()` returns that id twice — once as the synthetic substitution, once as the stored row — permanently, for every read of that session.

Consequences:

- the compaction summary is sent to the model **twice on every subsequent turn** (wasted context, duplicated content);
- clients receive two messages with the same `id`, which breaks any keyed rendering (React logs `Encountered two children with the same key, row:compaction_<uuid>`);
- it **compounds** — each further compaction creates another synthetic message that gets filed the same way;
- the filed row becomes a **parent link in the message chain**, so it cannot simply be deleted without re-parenting its children.

## Versions

`agents@0.19.0`, `@cloudflare/think@0.15.0`, `ai@6.0.210`. Think subclass (not `AIChatAgent`), browser client via `useAgentChat` from `agents/chat/react`, WebSocket transport, `getInitialMessages: null`, `resume: true`.

## Root cause

1. `compact()` writes an `assistant_compactions` row and leaves the underlying messages in place. `getHistory()` then substitutes a synthetic message for the compacted range — `applyCompactions`, `agents/dist/experimental/memory/session/index.js:1061-1087`:

   ```js
   result.push({
     id: `${COMPACTION_PREFIX}${comp.id}`,
     role: "assistant",
     parts: [{ type: "text", text: comp.summary }],
     createdAt: new Date()
   });
   ```

   This message is **computed on read** and has no row of its own.

2. It is broadcast to clients like any other message, so it lives in `useChat` state, and the transport posts the full transcript back on the next turn — `agents/dist/chat/react.js:202-206`:

   ```js
   const bodyPayload = JSON.stringify({ messages: options.messages, trigger: options.trigger, ...extraBody });
   ```

3. Think persists **every** incoming message — `@cloudflare/think/dist/think.js:6264-6275`:

   ```js
   const reconciled = reconcileMessages(incomingMessages, serverMessages, sanitizeMessage);
   for (const msg of reconciled) await this._persistIncomingMessage(msg, serverMessages);
   ```

   `reconcileMessages` passes the synthetic message through unchanged (it exact-matches an id present in `serverMessages`).

4. `_upsertMessageInHistory` (`think.js:1401-1405`) decides insert-vs-update by looking the id up in storage:

   ```js
   if (await this.session.getMessage(safe.id)) await this.session.updateMessage(safe);
   else await this.session.appendMessage(safe, parentId);
   ```

   `getMessage` reads only `assistant_messages` (`session/index.js:791-796`). A synthetic compaction id has no row there, so the `else` branch runs and **`appendMessage` inserts it as a real row**.

## Reproduction

Clean thread, `Think` subclass with the standard `useAgentChat` browser client. I inspected the Durable Object's SQLite directly at each step (`.wrangler/state/v3/do/...`).

| Step | `assistant_messages` | rows with `compaction_%` id | `assistant_compactions` |
|---|---|---|---|
| Fresh thread, a few short exchanges | 10 | 0 | 0 |
| One large tool result (~670 KB) → **auto-compaction fires** | 12 | **0** | 1 |
| **One ordinary user message** | 15 | **1** | 3 |

That single ordinary message is what files the row. The resulting chain:

```
c4af8cb8-…                   | assistant | dsDPImLyNTLghz
compaction_bb61da07-3990-4ce | assistant | c4af8cb8-9505-    ← filed as a real row
OLWluQByjgtm24n5             | user      | compaction_bb6    ← now a parent link
```

Reproduced independently on two separate threads.

## Suggested fix

Skip synthetic compaction messages at intake. The repo already ships exactly the predicate needed — `isCompactionMessage` (`agents/dist/compaction-helpers-*.js:58-60`, exported from `agents/experimental/memory/utils`):

```js
function isCompactionMessage(msg) {
  return msg.id.startsWith(COMPACTION_PREFIX);
}
```

It is currently called in only one place (filtering prior compaction messages out of a range being re-compacted) and not on the persistence path. Guarding `_persistIncomingMessage` / `_upsertMessageInHistory` with it would prevent new occurrences.

Two related points worth considering:

- **Existing sessions need repair.** Any session that has compacted and taken one more turn already carries the row, and because it is a parent link, a cleanup must re-parent its children rather than delete outright.
- **A defensive guard in `appendMessage`** (reject ids using the reserved `compaction_` prefix) would make the invariant hard to violate from any path, not just this one.

## Workaround

Filtering the synthetic messages out of the outgoing transcript client-side stops new rows from being created:

```ts
useAgentChat({
  agent,
  prepareSendMessagesRequest: ({ messages }) => ({
    body: { messages: messages.filter((m) => !isCompactionMessage(m)) }
  })
});
```

This works because `prepareSendMessagesRequest`'s `body` is `Object.assign`ed into `extraBody`, which the payload spreads after `messages:`. Verified: with two unfiled synthetic messages pending, one ordinary turn produced zero new rows, and normal message persistence was unaffected.

## Related

#1966 raises the server→client direction of the same boundary (keeping server-only context out of client projections). This issue is the reverse: client→server intake persisting a message the server itself generated on read.

justanotherbyte:
# What this fixes
Fixes #1984 where a Think session's compaction summary got persisted as a real message row and duplicated on every later read — sending the summary to the model twice, breaking keyed rendering on the client, and compounding with each further compaction.

# How
Fixed at the shared chokepoint every provider and caller routes through, rather than patching Think's intake. `Session._appendMessage` now refuses to persist any message using the reserved compaction_ prefix, and applyCompactions in both the agent (SQLite) and Postgres providers filters compaction-prefixed rows out of the read projection so already-affected sessions self-heal on read with no migration — a no-op for healthy sessions.

Added testing coverage in the agents package (intake-persistence repro, pre-filed-row read-dedup repro, Session write-guard unit tests) plus a full Think test.

# Why
`getHistory()` substitutes a synthetic `compaction_<id>` overlay for the compacted range on read — it has no row of its own. The browser transport echoes the full transcript back on the next turn, so that overlay arrived as an incoming message and got filed as a real assistant_messages row; from then on `getHistory()` returned the id twice (once as the substituted overlay, once as the stored row).

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/2036" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


.changeset/reconnect-optimistic-send.md
```diff
@@ -0,0 +1,5 @@
+---
+"agents": patch
+---
+
+Fix `useAgentChat` dropping an in-flight optimistic send on reconnect. When a message was sent while the socket was down, PartySocket buffers the frame for delivery, but the idle-connect transcript the server replays on reconnect (`cf_agent_chat_messages`) doesn't include it yet — so the whole-array `setMessages` replace erased the just-sent message from the UI until the turn completed and the server rebroadcast. The reconnect replay now preserves a trailing local-only `user` message that was buffered while disconnected, while still letting the server snapshot win for a delivered send it deliberately rolls back (e.g. `messageConcurrency: "drop"`) and for a regenerate's assistant replacement. Also fixes `@cloudflare/ai-chat` and `@cloudflare/think`, which re-export the hook unchanged.
```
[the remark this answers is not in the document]
cjol: Did you remove the changeset? It’s important to include one otherwise the change won’t be released when we merge.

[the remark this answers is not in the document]
cjol: oh wait I now see that was an unrelated changeset. Can you add a new one for this change @justanotherbyte ?

packages/agents/src/experimental/memory/session/session.ts
```diff
@@ -618,6 +621,19 @@ export class Session {
...
+    // A client transport round-trips the full transcript back on the next
+    // turn, so an overlay arrives here as an incoming message; persisting it
+    // would duplicate the summary in every later read (#1984). Reject it from
+    // any write path so the invariant can't be violated caller-side.
+    if (isCompactionMessage(message)) {
+      console.warn(
+        `[Session] Refusing to persist a message with the reserved compaction id "${message.id}" — synthetic compaction overlays are computed on read and are never stored (#1984).`
+      );
+      return;
+    }
```
cjol, line 635: Is this necessary? Receiving a compaction message from the browser is expected, so we’re logging a warning under normal, happy-path conditions (potentially even on every single turn after a compaction?)

packages/agents/src/tests/experimental/memory/session/provider.test.ts
```diff
@@ -299,6 +299,69 @@ describe("AgentSessionProvider — tree-structured messages", () => {
     expect(compactions).toHaveLength(1);
   });
 
+  it("does not persist synthetic compaction overlays echoed back by a client (#1984)", async () => {
```
cjol, line 302: For completeness, it might be good to add a test for the postgres provider as well

packages/agents/src/tests/experimental/memory/session/provider.test.ts
```diff
@@ -299,6 +299,69 @@ describe("AgentSessionProvider — tree-structured messages", () => {
...
+
+    const overlay = (await agent.getHistory()).find((m) =>
+      m.id.startsWith("compaction_")
+    );
+    expect(overlay).toBeDefined();
+
+    // Simulate an already-corrupted session: a synthetic overlay was filed as
+    // a real row on a prior turn (before the intake guard existed), parented
+    // into the live chain via the raw insert seam.
+    await agent.rawInsertChildForTest("m5", overlay!.id);
```
cjol, line 355: For completeness, I might test that a child of the overlay message is restored and re-parented correctly

cjol (changes requested):
Overall shape is good, thank you! Definitely need a changeset, a few nits otherwise in comments.

threepointone:
I checked this against current `main`. The bug is still there:

1. `sessions/overlays.ts` `overlayMessage` gives the `compaction_<id>` message a new `createdAt` on every read.
2. So after any cache rebuild (a wake, for example), the client's echoed overlay no longer matches the server's copy byte for byte in `_persistIncomingMessage`.
3. It then goes through `upsertMessage`, then `appendMessage`, and is inserted as a real row. Nothing on the write path rejects the `compaction_` prefix.

This needs a full rewrite rather than a rebase. All three files it edits (`experimental/memory/session/{session,providers/agent,providers/postgres}.ts`) were removed in #2196, and the Postgres provider no longer exists. On `main`, the fix would be:

- **A write guard** in `SessionHandle.appendMessage` / `upsertMessage`, or earlier in Think's intake, that drops an echoed overlay silently. Per @cjol's earlier review, this should not `console.warn`, because the echo is normal after every compaction.
- **A read-side filter** in `SessionCore`. The legacy migration in `core.ts` copies old rows forward, so sessions that already have stray overlay rows still need cleaning.
- **A test for @cjol's other request:** a child of a stray overlay row is re-parented correctly.
- **A changeset.**

PR #2038, opened 2026-08-04
fix(think): preserve regeneration branch context

### LINKED ISSUE #2028: Think regeneration includes the previous response in the model prompt
**Describe the bug**

`regenerate()` creates a new response in the UI, but `@cloudflare/think` still sends the previous assistant response to the model.

Think then adds this message:

> Continue your previous response from exactly where it left off. Do not repeat any of it.

As a result, regeneration behaves like “continue” instead of generating a fresh alternative.

Keeping the old response as a stored branch is correct. Including it in the new model request is not.

**To Reproduce**

1. Send a user message and wait for the assistant response.
2. Call `regenerate()`.
3. Inspect the messages sent to the model.
4. The request contains the previous assistant response followed by the continuation instruction above.

This was also reproduced in the upstream Think test suite using a model test double.

**Expected behavior**

Regeneration should create a sibling response using conversation history only up to the user message being answered.

The previous assistant response should remain available as a stored branch, but it should not be included in the new model request.

**Screenshots**

Not applicable. The issue is visible in the model request.

**Version:**

- `@cloudflare/think@0.15.0`
- `agents@0.19.0`
- AI SDK v7

Also reproduced against `cloudflare/agents` main at commit `2b2b5980e1945cf55f5a11626bc395e7c460516f`, where Think is `0.15.1`.

**Additional context**

The existing regeneration tests verify that sibling branches are stored correctly, but they do not inspect the messages sent to the model.

This is different from #1351, which tracks displaying and navigating stored branches in the UI.

It is related to #1012, where the older `@cloudflare/ai-chat` implementation also left the previous assistant response in server-side context during regeneration.

ben-reitz:
Stacked on #2120.

## Bug

Regeneration means “answer this user message again” while keeping the previous answer available. Think already stored the new answer as a sibling, but it generated that answer from the wrong history.

```text
Stored conversation:

Earlier history → User message
                  └─ Old answer  ← active leaf

Expected model input: Earlier history → User message
Actual model input:   Earlier history → User message → Old answer → “Continue…”
```

The result looked like a new branch in storage, but the model had been asked to continue the old answer instead of producing an independent alternative.

## Fix

Think now uses one server-validated branch point for both model input and output storage. For regeneration, model history ends at the selected user message, and the generated answer is stored beneath that same message.

```text
Earlier history → User message
                  ├─ Old answer  ← preserved
                  └─ New answer  ← generated without seeing the old answer
```

The selected history also remains the inference source when lifecycle hooks or context compaction prepare the turn.

## Testing

- 87 client-tools tests
- model-boundary regression verifies the prior assistant is absent from regenerated inference context
- Session assertions verify the previous and regenerated assistants remain siblings

Fixes #2028

## Stack

1. #2120 — Session branch-local compaction
2. **This PR — Think regeneration context**
3. #2121 — transcript repair ownership
4. #2122 — branch-aware recovery
5. #2057 — restart durability

### DIFF
--- a/.changeset/fix-think-regeneration-context.md
+++ b/.changeset/fix-think-regeneration-context.md
@@ -1,0 +1,8 @@
+---
+"agents": patch
+"@cloudflare/think": patch
+---
+
+Generate Think response alternatives from the selected regeneration branch point instead of continuing the previous answer.
+
+Preserve the branch selection through compaction and chat recovery so interrupted regenerated responses remain siblings of the previous answer.
--- a/.changeset/fix-think-regeneration-context.md
+++ b/.changeset/fix-think-regeneration-context.md
@@ -1,0 +1,6 @@
+---
+"@cloudflare/think": patch
+"agents": patch
+---
+
+Generate Think response alternatives from byte-budgeted history ending at the selected regeneration branch point instead of continuing the previous answer.
--- a/packages/agents/src/chat/orphan-persist.ts
+++ b/packages/agents/src/chat/orphan-persist.ts
@@ -39,6 +39,8 @@ export interface PersistReconstructedOrphanOptions<
    * present.
    */
   fallbackId: string;
+  /** Parent for a newly reconstructed message in tree-structured stores. */
+  parentId?: string | null;
   /**
    * Finalize the reconstructed message before upsert — e.g. strip internal
    * parts or resolve the persist-target id. Return `null` to skip persistence
@@ -84,7 +86,7 @@ export async function persistReconstructedOrphan<
   if (existing) {
     await options.store.updateMessage(options.merge(existing, prepared));
   } else {
-    await options.store.appendMessage(prepared);
+    await options.store.appendMessage(prepared, options.parentId);
   }
   return true;
 }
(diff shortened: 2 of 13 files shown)


threepointone:
I checked this against current `main`. The bug is still there. The WebSocket chat handler works out `branchParentId`, but passes it only to `_streamResult` and not to `_runInferenceLoop`. So `_assembleModelMessages` builds the prompt from the active path, which still ends with the old answer, and #2028 reproduces.

Most of this diff can be dropped now. Sessions moved to `packages/agents/src/sessions` in #2196, and Think's `ThinkSession` already accepts a branch leaf in `getHistory(leafId)`, `getRecentHistory(budget, { leafId })` and `compact(leafId)`. The `experimental/memory/session` changes, their tests and the `agents` changeset entry are no longer needed. A rebase that keeps only the `think.ts` plumbing and the Think tests should be close to mergeable.

Two things to watch in the rebase:

1. **Easy to miss:** the call `getRecentHistory(budget, MODEL_RECENT_WINDOW, leafId)` must become `getRecentHistory(budget, { leafId })`. On `main`, a number in the second argument is accepted and ignored, and the leaf id is dropped with it, so a straight port compiles and silently does nothing.
2. The windowing test relies on hydration keeping at least `MODEL_RECENT_WINDOW` messages. `main` no longer guarantees that, so the expected prompt roles will change.

Also, regeneration now re-reads storage on every attempt. That is fine, because only regeneration does it, but it cuts against `0966a0b0`, so it is worth a comment.

PR #2039, opened 2026-08-04
fix(agents): preserve optimistic send dropped on reconnect (#1983)

### LINKED ISSUE #1983: useAgentChat: cf_agent_chat_messages handler replaces the message array wholesale, dropping client-buffered sends on reconnect
**Describe the bug**

`useAgentChat`'s transport applies the server's authoritative transcript frame (`cf_agent_chat_messages`) with a whole-array `setMessages(next)`. On an idle connect/reconnect this replaces the local list with the server snapshot, dropping any optimistic message the client just added via `sendMessage()` that the snapshot doesn't yet contain.

Concretely: the user sends a message while the socket is reconnecting. PartySocket buffers the frame (so the turn is reliably delivered) and `sendMessage` renders the message optimistically — but the reconnect snapshot doesn't include it yet, so `setMessages(next)` erases it from the UI. It reappears only after the turn completes and the server rebroadcasts. So it's not data loss; it's a UI-consistency issue — the SDK optimistically shows a send it has committed to delivering, then its own replay wipes it mid-flight.

**To Reproduce**

1. Render a chat with existing messages via `useAgentChat`.
2. Drop the network (toggle Wi‑Fi off), then let the socket reconnect.
3. During the reconnect window (socket not yet `OPEN`), call `sendMessage(...)`. The message renders optimistically; PartySocket buffers the frame.
4. On reconnect, the agent's idle-connect flow pushes the transcript as `cf_agent_chat_messages` (without the just-sent message).
5. `setMessages(next)` runs → the optimistic message disappears (the user sees the reply but not their own message), returning only after the turn ends.

**Expected behavior**

The reconnect transcript replay shouldn't drop an in-flight optimistic message the client just sent. Merge the snapshot with local state (preserve trailing local-only messages it doesn't include) instead of replacing wholesale — or expose a reconcile hook so hosts can merge it themselves.

**Screenshots**

The sent message is visible, disappears during reconnect, then reappears after the turn. 

Image

**Version:**

`agents@0.18.0` (also reproduces on `0.17.3`). Handler in `agents/dist/chat/react.js`, re-exported unchanged by `@cloudflare/think`.

**Additional context**

- Location: `useAgentChat` → `onAgentMessage`, `case "cf_agent_chat_messages"`:
  ```js
  let next = preserveProtectedStreamingAssistant(data.messages);
  // ...observing-accumulator merge...
  setMessages(next); // whole-array replace — drops client-local optimistic sends
  ```
- The frame is pushed on every idle connect/reconnect (e.g. `@cloudflare/think`'s `_buildIdleConnectMessages` when no stream is active; `AIChatAgent` has an equivalent replay).
- Not data loss: the send is buffered and flushed on reconnect, so the turn runs and persists. Visible recovery depends on the server rebroadcasting the transcript to the *sender* after the turn; if the sender is excluded, it stays gone until reload.
- Host-side workaround we use: merge trailing local-only `role: "user"` messages instead of replacing (scoped to `user` so a legitimate assistant replacement like regenerate isn't resurrected).

justanotherbyte:
Fixes #1983 `useAgentChat` dropping an in-flight optimistic send on reconnect. When a message was sent while the socket was down, PartySocket buffers the frame for delivery, but the idle-connect transcript the server replays on reconnect (`cf_agent_chat_messages`) doesn't include it yet — so the whole-array `setMessages` replace erased the just-sent message from the UI until the turn completed and the server rebroadcast. The reconnect replay now preserves a trailing local-only `user` message that was buffered while disconnected, while still letting the server snapshot win for a delivered send it deliberately rolls back (e.g. `messageConcurrency: "drop"`) and for a regenerate's assistant replacement. Also fixes `@cloudflare/ai-chat` and `@cloudflare/think`, which re-export the hook unchanged.

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/2039" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/reconnect-optimistic-send.md
+++ b/.changeset/reconnect-optimistic-send.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
+Fix `useAgentChat` dropping an in-flight optimistic send on reconnect. When a message was sent while the socket was down, PartySocket buffers the frame for delivery, but the idle-connect transcript the server replays on reconnect (`cf_agent_chat_messages`) doesn't include it yet — so the whole-array `setMessages` replace erased the just-sent message from the UI until the turn completed and the server rebroadcast. The reconnect replay now preserves a trailing local-only `user` message that was buffered while disconnected, while still letting the server snapshot win for a delivered send it deliberately rolls back (e.g. `messageConcurrency: "drop"`) and for a regenerate's assistant replacement. Also fixes `@cloudflare/ai-chat` and `@cloudflare/think`, which re-export the hook unchanged.
--- a/.changeset/reconnect-optimistic-send.md
+++ b/.changeset/reconnect-optimistic-send.md
@@ -2,4 +2,4 @@
 "agents": patch
 ---
 
-Fix `useAgentChat` dropping an in-flight optimistic send on reconnect. When a message was sent while the socket was down, PartySocket buffers the frame for delivery, but the idle-connect transcript the server replays on reconnect (`cf_agent_chat_messages`) doesn't include it yet — so the whole-array `setMessages` replace erased the just-sent message from the UI until the turn completed and the server rebroadcast. The reconnect replay now preserves a trailing local-only `user` message that was buffered while disconnected, while still letting the server snapshot win for a delivered send it deliberately rolls back (e.g. `messageConcurrency: "drop"`) and for a regenerate's assistant replacement. Also fixes `@cloudflare/ai-chat` and `@cloudflare/think`, which re-export the hook unchanged.
+Fix `useAgentChat` dropping an in-flight optimistic send on reconnect: a message buffered by PartySocket while the socket was down was omitted from the server's idle-connect transcript replay, so the whole-array `setMessages` erased it from the UI until the turn completed.
--- a/packages/agents/src/chat/react.tsx
+++ b/packages/agents/src/chat/react.tsx
@@ -588,35 +588,26 @@ function prependMissingHydratedMessages<ChatMessage extends UIMessage>(
   return [...missingHydratedMessages, ...currentMessages];
 }
 
-const SOCKET_OPEN = 1; // WebSocket.OPEN
-
-// Re-append trailing local user messages the snapshot omits, stopping at the
-// first already-present or non-user message (never resurrects assistants or
-// reorders history).
-function preserveTrailingLocalUserSends<ChatMessage extends UIMessage>(
+// Re-append the specific buffered sends a snapshot omits, restoring only the
+// tracked ids (in local order) so a message the server deliberately dropped is
+// left alone. Known limitation: a buffered send that is delivered and then
+// rolled back shares its id with the rollback snapshot, so that exact id can
+// still be resurrected under non-`queue` concurrency.
+function restoreBufferedSends<ChatMessage extends UIMessage>(
   snapshot: ChatMessage[],
-  local: readonly ChatMessage[]
+  local: readonly ChatMessage[],
+  bufferedIds: ReadonlySet<string>
 ): ChatMessage[] {
-  if (local.length === 0) {
+  if (bufferedIds.size === 0) {
     return snapshot;
   }
 
   const snapshotIds = new Set(snapshot.map((message) => message.id));
-  const trailing: ChatMessage[] = [];
-  for (let index = local.length - 1; index >= 0; index--) {
-    const message = local[index];
-    if (snapshotIds.has(message.id) || message.role !== "user") {
-      break;
-    }
-    trailing.push(message);
-  }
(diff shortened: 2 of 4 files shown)


.changeset/reconnect-optimistic-send.md
```diff
@@ -0,0 +1,5 @@
+---
+"agents": patch
+---
+
```
cjol, line 4: Are you sure that @cloudflare/ai-chat is actually affected by this bug? I’m not sure the server behanves the same as Think in this respect.

packages/agents/src/chat/react.tsx
```diff
@@ -1512,10 +1549,25 @@ export function useAgentChat<
     fallbackAckedResumeRequestIdsRef.current.clear();
     replayHydratedAssistantMessageIdsRef.current.clear();
     protectedStreamingAssistantRef.current = null;
+    pendingBufferedSendRef.current = false;
   }, [markInitialMessagesSeeded, setMessages, resetToolContinuation]);
 
   const sendMessageWithStreamingProtection: typeof sendMessage = useCallback(
     async (message, options) => {
+      // Non-OPEN socket => PartySocket buffers this send; flag it for the
```
cjol: I’m not sure inspecting the state of the socket is necessarily safe, because the socket could disconnect later (but still before the messsage is actually sent) in which case the messgae would buffer but we wouldn’t know about it. I think that PartySocket’s send() returns wheter the message was sent immediately, so maybe you can use that result instead of checking readyState?

packages/agents/src/chat/react.tsx
```diff
@@ -588,6 +588,37 @@ function prependMissingHydratedMessages<ChatMessage extends UIMessage>(
...
 
+const SOCKET_OPEN = 1; // WebSocket.OPEN
+
+// Re-append trailing local user messages the snapshot omits, stopping at the
+// first already-present or non-user message (never resurrects assistants or
+// reorders history).
+function preserveTrailingLocalUserSends<ChatMessage extends UIMessage>(
+  snapshot: ChatMessage[],
+  local: readonly ChatMessage[]
+): ChatMessage[] {
```
cjol: Is there any chance that `messagesRef.current` might contain messages that the server deliberately omitted (rejected?). If so, we’d have to track the actual message IDs that were buffered and only restore those, rather than assuming that everything that’s missing from the snapshot should be restored.

packages/agents/src/react-tests/reconnect-optimistic-send.test.tsx
```diff
@@ -0,0 +1,228 @@
...
+  beforeEach(() => {
+    errorSpy = vi.spyOn(console, "error").mockImplementation(() => {});
+  });
+
+  afterEach(() => {
+    errorSpy.mockRestore();
+    cleanup();
+  });
+
+  it("preserves a buffered optimistic send when the reconnect snapshot omits it", async () => {
```
cjol, line 168: In line with my unsafe comment earlier, could we extend coverage to test socket changes during `prepareSendMessagesRequest`? 
1. start OPEN, close before the actual send()
2. start CLOSED, reopen before the actual send()

ternarybits:
+1 for the optimistic-send fix, with an additional queued-send case to cover.

Source recheck on 2026-09-15: latest released agents 0.23.0 / ai-chat 0.12.0 / Think 0.18.0 and main `a7b29135acf127cdf4a44114652970d0a22300de` still route `CF_AGENT_CHAT_MESSAGES` through [`preserveProtectedStreamingAssistant`](https://github.com/cloudflare/agents/blob/5f7ad7e4edac2ec8dd1d6a31758f251cb52373fa/packages/agents/src/chat/react.tsx#L1312). That helper protects the assistant, but does not retain a local-only user send; the [snapshot handler](https://github.com/cloudflare/agents/blob/5f7ad7e4edac2ec8dd1d6a31758f251cb52373fa/packages/agents/src/chat/react.tsx#L1902) otherwise starts from the server list. This PR remains open/unmerged.

Our UI presents each Send immediately, before a durable submission is promoted into the transcript. A full sync snapshot arriving in that interval can omit the sender's own local user message; it disappears until the promotion broadcast. We currently bridge these attempts by user-message identity outside `useAgentChat`.

Please cover both the reconnect-buffered case in this PR and a send that has reached the server but is still awaiting durable queue promotion when a sync snapshot arrives. The latter may require a separate follow-up rather than extending this PR's disconnected-only preservation rule. Preserve genuine server rollback/drop/clear semantics: do not unconditionally merge every stale local user message. An acknowledgement/sync contract exposing request-to-user-message identity would also let callers reconcile deterministically.

This is current source-path validation and our integration use case, not a fresh end-to-end reproduction against the unmerged PR.

PR #2043, opened 2026-08-04
Python package support

abstractedfox:
Adds support for Python packages in worker-bundler, and thus to Python dynamic workers.

The current iteration of this feature leans on the Pyodide package index, with PyPI filling in as a fallback. In this state, it has parity with the way packages work in Pyodide normally, which is to say only one version of a package is made available by the index for a given version of Pyodide. For that reason, it retrieves this version (as long as the package is available) without handling version ranges (as the dependency string spec otherwise allows).

This has been reviewed by the Python Workers team in a series of PR's to a clone (https://github.com/abstractedfox/agents)

### DIFF
--- a/.changeset/early-weeks-live.md
+++ b/.changeset/early-weeks-live.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/worker-bundler": patch
+---
+
+Python package support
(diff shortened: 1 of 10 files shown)


dom96 (approved):
We've reviewed the changes here already as part of separate PRs in https://github.com/abstractedfox/agents/pulls?q=is%3Apr+is%3Aclosed. This is looking good and ready for merge.

Issue #1913, opened 2026-07-10
useAgentChat: "Maximum update depth exceeded" thrown during stream resume after a reconnect — surfaces as a terminal error mid-turn (not fixed by #1838)


mrgarcia1998:
**Describe the bug**

When a chat client's WebSocket closes mid-turn and PartySocket auto-reconnects, the resumed stream's replayed chunks drive a synchronous re-render cycle inside the shared chat hook (`agents/chat/react`, consumed via `@cloudflare/think/react`): `onAgentMessage → ReactChatState set messages → forceStoreRerender → …` repeating until React's update-depth guard throws `Maximum update depth exceeded`.

The exception escapes through `useAgentChat`'s catch path, so the client lands in `status: "error"` with `error` set **mid-turn** — while the durable turn on the agent DO is healthy and completes normally (verified in the DO message store: stream metadata `completed`, full assistant reply persisted).

#1838 does not fix this: with resumes serialized, a **single** close + its one resumed stream is still enough to trigger the loop. Reproduced in 3 separate runs on the versions below.

**To Reproduce**

1. A `Think`-based agent with chat over `useAgent` + `useAgentChat` (defaults — `resume` on).
2. Start a long streaming turn (a multi-second reply is enough).
3. Close the underlying WebSocket once mid-stream (e.g. grab the raw socket and call `.close()`; PartySocket auto-reconnects).
4. On reconnect the stream resumes and replays; the browser console shows the `Maximum update depth exceeded` React error, and `useAgentChat` reports `status: "error"` + a set `error` while the server-side turn keeps running.

**Expected behavior**

A resumed stream replays without tripping React's update-depth guard — and in general, a client-side rendering failure should not mark a turn as terminally errored while it is still running (and later completes) server-side. The false terminal error also cascades into a second bug (filed separately, linked below): once the client sits in `"error"`, stream resume never re-arms, so `isServerStreaming` can stick `true` forever.

**Version:**

`agents@0.17.3`, `@cloudflare/think@0.12.1`, `ai@6.0.205`, `@ai-sdk/react@3.0.207`, `react@19.2.7`, `partysocket@1.3.0`. Observed both under local `wrangler dev` and in production.

**Additional context**

Found while adopting the #1799 `connectionError` surface. Downstream we mitigate with an error-triggered debounce + a bounded `getMessages` reconciliation loop, but the root loop lives in the shared hook's message-replay path. Happy to provide the full captured stack / more repro detail if useful.

[...]

Issue #1733, opened 2026-06-11
ai-chat: double STREAM_RESUMING replay duplicates the streaming assistant text part


advaitpaliwal:
## Summary

When a `useAgentChat` client connects while a turn is mid-stream (e.g. the WebSocket opens after the turn started), the server sends `CF_AGENT_STREAM_RESUMING` **twice** to the same connection:

1. proactively in `onConnect` (`packages/ai-chat/src/index.ts` ~L744), and
2. again in response to the client's `CF_AGENT_STREAM_RESUME_REQUEST` (~L1059) — the request the React client sends once its message handler is registered (`resume` defaults to `true`).

`_notifyStreamResuming` has no per-connection dedupe and the `RESUME_ACK` handler has no double-replay guard, so the client ACKs both notifies and `replayChunks` re-sends the full chunk buffer twice.

The client *does* have a reset for replays — `resetMatchingHydratedAssistantForReplay`, gated on `pendingReplayResumeRequestIdsRef` — but the pending id is **deleted on the first replay's `start` chunk** (`packages/ai-chat/src/react.tsx` ~L1936-1949). When both RESUMING frames arrive before the first replay's `start` (the common ordering: both notifies are queued back-to-back on the socket), the second replay's `start` finds the pending id already deleted, skips the reset, passes the `status === "observing"` gate, and its `text-start` opens a **second text part**. All subsequent live deltas feed the new part while the first freezes — the user sees the same reply streaming as two stacked text blocks. The persisted message is clean, so the duplication snaps away on refresh/finalize.

## Repro

1. Agent with an active streamed turn; client connects mid-stream with `resume: true` (default). Easiest real-world trigger: start the turn over a detached HTTP request and let the WS connect a second later.
2. Log WS frames: you'll see two `cf_agent_stream_resuming`, two ACKs, and the turn's `start` chunk replayed twice.
3. If text deltas were already buffered when the second replay runs, the assistant message ends up with two `step-start`/`text` pairs while streaming.

## Fix that worked for us

Client-side, make replay idempotent — reset the matching trailing assistant message on **every** replayed `start`, not only when the request id is still pending:

```ts
// react.tsx, both the localRequestIds branch and the broadcast branch
if (data.replay && typeof chunkData.messageId === "string" && chunkData.type === "start") {
  pendingReplayResumeRequestIdsRef.current.delete(data.id);
  resetMatchingHydratedAssistantForReplay(chunkData.messageId);
}
```

[...]

PR #2050, opened 2026-08-05
fix(chat): batch replayed chunks on stream resume

ben-reitz:
Refs #1913. Thanks to @magicismight for the diagnosis and the fork branch that inspired this fix.

## The bug

Your WebSocket drops in the middle of an answer and reconnects a second later. Instead of the answer carrying on, the console fills with `Maximum update depth exceeded` and the chat shows an error - even though the server finished the answer fine and has it stored. Only a reload clears it.

## Why it happens

When a client reconnects mid-turn, the server replays the answer so far: every chunk it has stored, back to back, as fast as the socket will carry them.

The transport handed each chunk straight to the AI SDK. Every chunk rewrites the chat state, and every rewrite is a React render. React allows 50 renders in an unbroken row before it assumes something is looping and throws - and a replayed answer is easily 400 chunks, arriving with nothing in between to break the chain.

The throw lands deep inside the SDK's stream loop, which can't tell a rendering failure from a dead connection, so it marks the turn as failed and stops reading.

```mermaid
flowchart LR
  S["server replays<br/>400 stored chunks"] --> T{transport}
  T -->|"before: one at a time"| B1["400 chat-state writes"] --> B2["renders 1 to 50"] --> B3["render 51 throws<br/>status: error"]
  T -->|"after: collected and merged"| A1["about 5 writes"] --> A2["5 renders"] --> A3["answer restored<br/>status: ready"]
```

## The fix

The transport now catches replayed chunks in a short window instead of forwarding them one by one, and glues together the ones belonging to the same piece of text. A few hundred chunks become a handful of updates - roughly one per part of the message. The replayed text also appears at once, rather than visibly re-typing itself.

Holding chunks is only safe if they can never be stranded. Three rules take care of that:

- The window ends the moment the replay does.
- The transport empties the window before it closes a stream, so a second disconnect can't swallow what already arrived.
- If the same answer replays twice, the newer copy replaces the held one, so it can't be applied twice (#1733).

It's one new module, `chat/replay-batch.ts`, wired into the two resume paths in `ws-chat-transport.ts`. Nothing changes in the server, the wire format, or `chat/react.tsx`.

## Tests

18 unit tests in `chat/__tests__/replay-batch.test.ts` cover the merge rules, the window, the boundaries and an errored turn. Six tests in `react-tests/resume-replay-burst.test.tsx` drive the real hook with real wire frames - all six fail on `main`, and that is the only layer where the React throw is visible at all.

Suites: agents chat 532 · react 133 · workers 1801 · ai-chat 737 + react 82 · think 1000 · `pnpm run check`.

## Next steps

This massively reduces the chance of the bug appearing in the wild, but a few things remain for a follow-up:

| | Problem | Why it remains |
| --- | --- | --- |
| 1 | A replay of more than ~7 tool steps still hits the limit (measured: 6 steps fine, 8 throws) | Merging only joins chunks within a single part, and each tool step costs about 7 updates by itself |
| 2 | A fast enough live stream could do the same in theory | The live path still forwards one chunk at a time. Real models pause between tokens, which lets React reset its counter |
| 3 | Any client-side render error still marks the turn failed until a reload or a reconnect | The SDK's catch can't tell a render failure from a stream failure |

For 1 & 2, setting a default value for `throttle` on `useAgentChat` avoids the crash outright - which is what 25 of our 27 examples already do. Am going to raise a followup PR after this to set a default value for it in `useChatAgent` rather than leaving it unthrottled.

### DIFF
--- a/.changeset/eighty-donkeys-repeat.md
+++ b/.changeset/eighty-donkeys-repeat.md
@@ -11,10 +11,18 @@ loop and threw "Maximum update depth exceeded" — surfacing as a false terminal
 `status: "error"` for a turn the server had completed, with the replayed text
 visibly re-typing beforehand.
 
-Both transport-owned resume paths now buffer `replay: true` chunks and merge
-consecutive deltas of the same part, delivering the compacted batch once at the
-replay boundary (`replayComplete`, `done`, or the first live chunk). Applying a
+Both transport-owned resume paths now collect `replay: true` chunks in a short
+coalescing window and merge consecutive deltas of the same part. The window
+closes at the end of the event-loop turn that opened it, or earlier at a replay
+boundary (`replayComplete`, `done`, or the first live chunk). Applying a
 replayed prefix is now proportional to the number of message parts rather than
-the number of chunks. Cross-part ordering is unchanged, deltas carrying
-`providerMetadata` are kept whole, and an errored resumed turn still delivers
-its replayed content before the terminal error.
+the number of chunks.
+
+The window is bounded to a single turn on purpose: a burst spread over several
+turns cannot trip React's guard, and holding replayed chunks any longer would
+reorder them against the hook's own replay bookkeeping, which repairs a second
+replay of the same turn (#1733) by inspecting already-applied messages.
+
+Cross-part ordering is unchanged, deltas carrying `providerMetadata` are kept
+whole, and an errored resumed turn still delivers its replayed content before
+the terminal error.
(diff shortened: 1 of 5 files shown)


marcoaleixoo:
hey @ben-reitz , I'm facing this issue in production right now and it's getting frequent.
Do you think that this PR will be merged/released soon?
I was planning to patch locally with your proposed solution while you work on getting this released.

PR #2051, opened 2026-08-05
fix(agents): stream forwarded request bodies into sub-agents

### LINKED ISSUE #2015: forwardToFacet buffers the entire forwarded request body with no size limit
### Summary

`forwardToFacet` reads the whole forwarded request body into memory before
dispatching to the facet:

```ts
// packages/agents/src/sub-routing.ts:215
if (req.body && req.method !== "GET" && req.method !== "HEAD")
  forwardInit.body = await req.arrayBuffer();
```

No ceiling is applied. The same call ships in `agents@0.20.1` (current latest)
at `dist/sub-routing.js:113` and `dist/index.js:4103`.

### Why it matters

The read happens inside `Agent.fetch`, which returns before `onRequest` whenever
the path matches `/sub/{class}/{name}`:

```ts
async fetch(request) {
  const ctx = this.ctx;
  const match = parseSubAgentPath(request.url, {
    knownClasses: ctx.exports ? Object.keys(ctx.exports) : undefined,
  });
  if (!match) return super.fetch(request);
  // ...
  return this._cf_forwardToFacet(forwardReq, match);
}
```

An application's own request-intake limits therefore never run on this path. An
app that carefully bounds request bodies in `onRequest` still has an unbounded
read in front of it, executing in the parent Durable Object, which has a fixed
isolate memory budget.

### Suggested fix

Stream the body through rather than materialising it — `forwardInit.body =
req.body` with `duplex: "half"` where the runtime supports it — or apply a
configurable ceiling and reject above it.

### Measured impact

Peak resident memory across the `workerd` processes for a single POST, with a
handler that never reads the request body, so the growth is attributable to the
forwarder rather than to application code. One fresh worker per row.

| Request body | `…/{name}/diag` (canonical) | `…/{name}/sub/probe/child/diag` (facet) |
| --- | --- | --- |
| 16 MB | +1 MB | +71 MB |
| 64 MB | +1 MB | +284 MB |
| 128 MB | +1 MB | +556 MB |

The canonical path is flat because nothing reads the body. The facet path grows
at roughly 4.3–4.4x the body size, linearly, because the body is materialised as
an `ArrayBuffer` and then copied into a new `Request`. A Durable Object isolate
has a fixed memory budget, so a single sufficiently large request is enough to
put the parent object under pressure.

### Reproduction

`wrangler.jsonc` — the `experimental` flag matters: without it `ctx.facets` and
`ctx.exports` are unavailable and facet routing never engages.

```jsonc
{
  "name": "facet-probe", "main": "w.ts", "compatibility_date": "2026-07-31",
  "compatibility_flags": ["nodejs_compat", "experimental"],
  "durable_objects": { "bindings": [{ "name": "Probe", "class_name": "Probe" }] },
  "migrations": [{ "tag": "v1", "new_sqlite_classes": ["Probe"] }]
}
```

```ts
// w.ts
import { Agent, routeAgentRequest } from "agents";

export class Probe extends Agent<Env, { n: number }> {
  override initialState = { n: 0 };
  override async onRequest(): Promise<Response> {
    return Response.json({ reached: "onRequest" }); // never reads the body
  }
}
export default {
  async fetch(request: Request, env: Env): Promise<Response> {
    return (await routeAgentRequest(request, env)) ?? new Response("nf", { status: 404 });
  },
} satisfies ExportedHandler<Env>;
```

```bash
npx wrangler dev --local --config wrangler.jsonc --port 8830
# control — body ignored, memory flat
head -c $((64*1024*1024)) /dev/zero | tr '\0' 'a' | \
  curl -s -o /dev/null -X POST --data-binary @- \
  http://127.0.0.1:8830/agents/probe/a/diag
# facet — same body, forwarder buffers it
head -c $((64*1024*1024)) /dev/zero | tr '\0' 'a' | \
  curl -s -o /dev/null -X POST --data-binary @- \
  http://127.0.0.1:8830/agents/probe/a/sub/probe/child/diag
```

Sample `ps -o rss=` across the `workerd serve` processes during each request.

### Notes

- The facet API is marked `@experimental`, so this may be deliberate for now.
  Flagging it because the read sits *in front of* application-level validation
  rather than behind it, which means an application cannot bound it itself.
- Measured under `wrangler dev --local` on macOS/arm64 with `agents@0.20.1`. I
  have not measured this on deployed Workers.

AntoniTok:
Closes #2015

## The problem

When a request goes to a sub-agent (a URL with `/sub/...` in it), the parent used to read the *entire* upload into memory before passing it on. No limit. A 128 MB upload made the parent balloon by ~546 MB. And your app couldn't stop it, because this happened before your code ever ran.

## What changed

Two lines. Instead of reading the whole body first, the parent now just hands the body through as a stream — it passes the pipe along instead of filling a bucket and carrying it.

```diff
  if (req.body && req.method !== "GET" && req.method !== "HEAD") {
-   forwardedInit.body = await req.arrayBuffer();
+   forwardedInit.body = req.body;
  }
```

Both call sites had it: `Agent._cf_forwardToFacet` (`src/index.ts`) and `routeSubAgentRequest` (`src/sub-routing.ts`).

## Why it broke in the first place

#1443 fixed a WebSocket bug back in May and needed to set a header on upgrade requests. Doing that meant rebuilding the request from an explicit `RequestInit`, which doesn't inherit a body — so the body got re-attached by hand, the slow way, as a side effect. It streamed fine before that.

I verified this rather than assuming it. Reverting to the pre-#1443 form (`new Request(rewritten, req)`) makes all the new streaming tests pass, which confirms where the buffering came from. So this is a genuine regression, shipped in `agents@0.12.1`. In terms of behaviour this restores the streaming, but #1443's code stays the same. The `RequestInit` and `Upgrade` header handling are untouched and only the body attachment changes.

Worth noting *why* it mattered beyond raw memory: `Agent.fetch` returns before `onRequest` whenever the path matches `/sub/{class}/{name}`, so the unbounded read sat **in front of** application-level validation. An app that carefully bounded request bodies in `onRequest` still had an unbounded read ahead of it and no way to bound it. The cost was also **per hop** — a nested `/sub/.../sub/...` address re-materialised the same bytes at every level.

## Result

Measured with `wrangler dev --local` against the reporter's repro (a handler that never reads the body, so growth is attributable to the forwarder). Two tarballs built from the same commit, differing only in this fix. Fresh worker per row, peak RSS summed across the `workerd` processes for a single POST:

| Request body | facet route, before | facet route, after | canonical route (control) |
| --- | --- | --- | --- |
| 16 MB | +75 MB | +4 MB | +2 MB |
| 64 MB | +268 MB | +4 MB | +2 MB |
| 128 MB | +546 MB | +4 MB | +2 MB |

Flat now, no matter how big the upload. The before column reproduces the reporter's +71/+284/+556 MB; the after column matches the canonical control.

## The one caveat

The upload is now connected end to end. If a sub-agent replies without reading the body, the rest of the upload gets cut off instead of being quietly swallowed. That's the correct behaviour and what you'd want — but it *is* a change, so anyone relying on the old "swallow it all" habit would notice.

## Tests

New file `src/tests/sub-agent-body-forwarding.test.ts` (9 tests), covering both call sites:

- **dispatch before upload completes** — the child replies while the request body is still open. Deadlocks against the old code.
- **prefix of an unfinished body** — the child reads one chunk of a still-open body. This is the decisive one: it distinguishes real incremental streaming from "the runtime buffered internally but dispatched early".
- **integrity** — a 2 MB body arrives byte-exact (SHA-256) over one hop, over two nested hops, and via `routeSubAgentRequest`.
- **GET/HEAD** — unchanged.

The suite includes a canonical (non-facet) control that runs first. If the test harness itself couldn't stream a request body, every facet assertion would hang for unrelated reasons, so the control makes a failure interpretable. Fixtures are `BodyProbeSubAgent` (facet) and `BodyProbeRootAgent` (root control), sharing one handler.

The three streaming assertions fail against `main` with `expected 'pending' to be 'settled'` after 2.5s, and pass in ~25ms with this change.

## Verification

- `pnpm run test:workers` — 91 files / 1810 tests
- full `agents` suite — 131 files / 2552 tests
- `@cloudflare/ai-chat` 50/737, `@cloudflare/voice` 12/225, `nx affected -t test` 27/27
- `pnpm run check` — sherif, export check, oxfmt, oxlint, typecheck (120 projects)

The existing WebSocket sub-agent tests matter here specifically, since this touches the function #1443 changed — they pass.

## Not addressed

If an `onBeforeSubAgent` hook consumes the body and returns `void`, forwarding still throws (the stream is disturbed). That was already true with `arrayBuffer()`, just with a different message — orthogonal to this issue, so I left it alone.

---

<a href="https://app.devin.ai/review/cloudflare/agents/pull/2051" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/facet-forward-stream-body.md
+++ b/.changeset/facet-forward-stream-body.md
@@ -1,0 +1,22 @@
+---
+"agents": patch
+---
+
+Stream forwarded request bodies into sub-agents instead of buffering them in the parent Durable Object.
+
+`Agent._cf_forwardToFacet` and `routeSubAgentRequest` both did `forwardInit.body = await req.arrayBuffer()` before dispatching to a child facet, materialising the entire request body in the parent's isolate. Two consequences:
+
+- The read sat **in front of** application-level validation. `Agent.fetch` returns before `onRequest` whenever the path matches `/sub/{class}/{name}`, so an app that carefully bounded request bodies in `onRequest` still had an unbounded read ahead of it — and no way to bound it itself.
+- The cost was **per hop**. A nested `/sub/.../sub/...` address re-materialised the same bytes at every level.
+
+Both call sites now pass `req.body` through as a stream. Measured on `wrangler dev --local` with a handler that never reads the body, peak RSS across the `workerd` processes for a single POST:
+
+| Request body | facet route, before | facet route, after | canonical route (control) |
+| ------------ | ------------------- | ------------------ | ------------------------- |
+| 16 MB        | +75 MB              | +4 MB              | +2 MB                     |
+| 64 MB        | +268 MB             | +4 MB              | +2 MB                     |
+| 128 MB       | +546 MB             | +4 MB              | +2 MB                     |
+
+This restores the behaviour from before #1443, which switched to an explicit `RequestInit` in order to set a header on WebSocket upgrades and re-attached the body with `arrayBuffer()` as a side effect. The `Upgrade` header handling from that fix is unchanged.
+
+One behavioural note: backpressure now reaches the client. A child that returns without reading the body will cause the remainder of the upload to be cancelled, where previously the parent drained it in full.
--- a/packages/agents/src/index.ts
+++ b/packages/agents/src/index.ts
@@ -7736,8 +7736,12 @@ export class Agent<
     if (req.headers.get("Upgrade")?.toLowerCase() === "websocket") {
       forwardedHeaders.set(SUB_AGENT_OUTER_URL_HEADER, req.url);
     }
+    // Hand the body through as a stream. Reading it here (e.g.
+    // `await req.arrayBuffer()`) materialises the entire body in the
+    // parent DO's isolate, ahead of any application-level intake limit,
+    // and re-materialises it once per `/sub/` hop — see #2015.
     if (req.body && req.method !== "GET" && req.method !== "HEAD") {
-      forwardedInit.body = await req.arrayBuffer();
+      forwardedInit.body = req.body;
     }
     const forwarded = new Request(rewritten, forwardedInit);
     return fetcher.fetch(forwarded);
--- a/packages/agents/src/sub-routing.ts
+++ b/packages/agents/src/sub-routing.ts
@@ -424,8 +424,12 @@ export async function routeSubAgentRequest(
     method: req.method,
     headers: new Headers(req.headers)
   };
+  // Stream the body through rather than buffering it — see #2015. This
+  // helper runs in the caller's isolate, but the same request then hits
+  // `_cf_forwardToFacet` on the parent, so buffering here would put a
+  // second unbounded copy in front of the child.
   if (req.body && req.method !== "GET" && req.method !== "HEAD") {
-    forwardInit.body = await req.arrayBuffer();
+    forwardInit.body = req.body;
   }
   const forwardReq = new Request(forwardUrl, forwardInit);
 
--- a/packages/agents/src/tests/agents/index.ts
+++ b/packages/agents/src/tests/agents/index.ts
@@ -71,5 +71,7 @@ export {
   Sub_,
   ReservedClassParent,
   TestUnboundParentAgent,
-  TestMinifiedNameParentAgent
+  TestMinifiedNameParentAgent,
+  BodyProbeSubAgent,
+  BodyProbeRootAgent
 } from "./sub-agent";
--- a/packages/agents/src/tests/agents/sub-agent.ts
+++ b/packages/agents/src/tests/agents/sub-agent.ts
@@ -2223,3 +2223,100 @@ class _a extends Agent {
   }
 }
 export { _a as TestMinifiedNameParentAgent };
+
+// ── Request-body forwarding probes (issue #2015) ─────────────────────
+//
+// `_cf_forwardToFacet` and `routeSubAgentRequest` used to materialise
+// the whole forwarded body via `await req.arrayBuffer()` before
+// dispatching. These fixtures let a test observe *when* the child sees
+// the request relative to the client finishing its upload, which is
+// what distinguishes streaming from buffering.
+//
+// The same handler backs a facet child (`BodyProbeSubAgent`) and a
+// root Agent (`BodyProbeRootAgent`). The root is the control: it
+// proves the *test harness* can stream a request body, so a hang on
+// the facet path can be attributed to the forwarder rather than to
+// vitest-pool-workers.
+
+function toHex(buf: ArrayBuffer): string {
+  return Array.from(new Uint8Array(buf))
+    .map((b) => b.toString(16).padStart(2, "0"))
+    .join("");
+}
+
+/**
+ * Probe endpoints:
+ *   - `/probe/ignore`      — reply immediately, never touch the body.
+ *   - `/probe/first-chunk` — read exactly one chunk, then reply.
+ *   - `/probe/drain`       — read to completion, reply with size + digest.
+ */
+async function handleBodyProbe(
+  agentName: string,
+  request: Request
+): Promise<Response> {
+  const { pathname } = new URL(request.url);
+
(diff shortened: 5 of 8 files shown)


cjol:
> #1443 fixed a WebSocket bug back in May and needed to set a header on upgrade requests. Doing that meant rebuilding the request from an explicit `RequestInit`, which doesn't inherit a body — so the body got re-attached by hand, the slow way, as a side effect. It streamed fine before that.
> 
> I verified this rather than assuming it: reverting to the pre-#1443 form (`new Request(rewritten, req)`) makes all the new streaming tests pass. So this is a genuine regression, shipped in `agents@0.12.1`, and this restores it. #1443's `Upgrade` header handling is untouched.

Could you explain a little more why 1443 needed to rebuild the request in this way, and why reverting to the previous behaviour does not cause the original WebSocket bug not to come back?

cjol:
Do you think it would be appropriate to add the memory consumption benchmark as a regression test in CI somehow/

AntoniTok:
> > #1443 fixed a WebSocket bug back in May and needed to set a header on upgrade requests. Doing that meant rebuilding the request from an explicit `RequestInit`, which doesn't inherit a body — so the body got re-attached by hand, the slow way, as a side effect. It streamed fine before that.
> > I verified this rather than assuming it: reverting to the pre-#1443 form (`new Request(rewritten, req)`) makes all the new streaming tests pass. So this is a genuine regression, shipped in `agents@0.12.1`, and this restores it. #1443's `Upgrade` header handling is untouched.
> 
> Could you explain a little more why 1443 needed to rebuild the request in this way, and why reverting to the previous behaviour does not cause the original WebSocket bug not to come back?

In terms of behaviour this goes back to how it worked before #1443, meaning bodies stream again. But #1443's code stays the same. The `RequestInit`, the `Upgrade` check and the `SUB_AGENT_OUTER_URL_HEADER` line are all still there. The only change is `body = await req.arrayBuffer()` becoming `body = req.body`.

I only tried the literal pre-#1443 form (`new Request(rewritten, req)`) as a bisect step to find where the buffering came from. I'll make that clearer in the description.

#1443 rebuilt the request because It had to set `x-cf-agents-subagent-url` on upgrade requests so the parent knows which sub-agent path a socket belongs to. `onConnect` reads it back at `index.ts:2582`.

Incoming request headers are immutable in workerd, so setting one means copying the headers and building a fresh `Request`. The pathname is rewritten too, so it used a literal `RequestInit`, which doesn't carry the body over the way `new Request(url, req)` does.

The body was re-attached by hand and `arrayBuffer()` is what got written.

The tell is `routeSubAgentRequest`, which got the same rebuild but sets no header at all it just copies the shape.

So the WS fix stays intact. Upgrades never reach the changed line. `Agent.fetch` catches them at `index.ts:7031` and serves them from the parent, before the `_cf_forwardToFacet` call at 7039. Upgrades are also GET with no body, so the guarded line is skipped anyway. Anyhow all of #1443's tests pass.

AntoniTok:
> Do you think it would be appropriate to add the memory consumption benchmark as a regression test in CI somehow/

It could be worthwhile but it would have to be nightly and non-gating and it's not really what protects this.

I don't think workers can measure their own memory, so this can't be a `test:workers` test. which means we'd have to start a real `wrangler dev`, pushing 128 MB at it, and sampling RSS from outside which would be slow and noisy on shared runners, so the threshold would have to be loose.

And also the streaming tests already catch this regression on their own. "Child reads a prefix of a still-open body" can't pass against a buffering parent, because buffering holds the child back until the upload finishes. On `main` it's still pending after 2.5s. Here it's 25ms. No threshold to tune, and it's already in the gating job.

cjol (approved):
Okay, the key information I was looking for is that the body buffering wasn't actually part of the fix for 1443. That's fine then. Regarding CI, I still think it might be valuable to monitor memory consumption under various scenarios, because the streaming tests cover this specific regression but a more general lens might catch other similar failures. But that doesn't need to block this PR.

PR #2058, opened 2026-08-06
feat(chat): throttle chat UI updates by default

### LINKED ISSUE #1913: useAgentChat: "Maximum update depth exceeded" thrown during stream resume after a reconnect — surfaces as a terminal error mid-turn (not fixed by #1838)

ben-reitz:
A proposal, stacked on #2050. Fixes #1913 together with that PR.

## Why a second change

#2050 stops a resumed stream from writing chat state once per replayed chunk. It merges the chunks that belong to the same piece of text, so a few hundred chunks become a handful of writes.

That helps enormously, but it only merges *within* a part. A tool step costs about seven writes on its own, so a replayed turn of more than about seven tool steps still reaches React's 50-render limit and still shows a false error. Measured with the merge in place: six steps fine, eight throws.

A throttle does not care how many chunks arrive, so it covers what merging cannot.

```mermaid
flowchart LR
  C["chunks arrive"] --> M["#2050<br/>merge replayed chunks"] --> T["this PR<br/>throttle renders"] --> R["at most one render<br/>per 50ms"]
```

## The change

`useAgentChat` now coalesces chat updates every 50ms. Pass `throttle: false` to render every chunk as it arrives, or a number to pick your own interval.

Nothing else changes, and the first chunk of a stream is never delayed - the AI SDK throttles with `throttleit`, which runs the first call of an idle window immediately.

## Why 50ms

Any value above zero prevents the crash, because the update then arrives from a timer rather than from the current task. Verified at 1, 4, 16, 50, 100 and 250ms. So the size of the number is a cost decision rather than a safety one. Measured over a 200-chunk turn at ~100 chunks/sec (caveat that this is obvs on a powerful macbook):

| throttle | renders | render time |
| --- | --- | --- |
| off | 404 | 138ms |
| 16ms | 261 | 89ms |
| **50ms** | **90** | **37ms** |
| 100ms | 48 | 21ms |
| 250ms | 22 | 9ms |

50ms removes 78% of the renders. Larger values save progressively less while letting the text lag further behind the stream. It is also the value the AI SDK's own documentation uses, and 25 of our 27 examples already pass 100ms by hand, so throttling is not new behaviour for our own UIs - only the default is.

## Two things worth knowing

`@ai-sdk/react` v3 reads only `experimental_throttle`; v4 renamed it to `throttle`. Both majors are in our peer range, so the value is sent under both names. Turning it off omits both names rather than sending `0`, because both majors decide with `waitMs != null`, and only an omitted option reaches their unthrottled path.

This proposed change provides a mitigation rather than a guarantee. `useChat` throttles its store subscription, but its `getSnapshot` returns a new messages array on every chunk, and React forces a synchronous re-render whenever that identity moved during a render - a path the throttle never sees. That is vercel/ai#6166, with a fix open in vercel/ai#17893. Writing state less often is the only thing that bounds it, which is why #2050 does the heavier lifting.

## Tests

Eight unit tests in `chat/__tests__/chat-throttle.test.ts` cover the precedence rules and both spellings. Two tests in `react-tests/default-throttle.test.tsx` drive the real hook with a 12-step tool turn: it stays healthy on defaults, and `throttle: false` restores the old failure, which proves the caller's value reaches the SDK.

Suites: agents chat 540 · react 135 · ai-chat 737 · think 920 + react 5 · `pnpm run check`.

### DIFF
--- a/.changeset/silent-paws-shine.md
+++ b/.changeset/silent-paws-shine.md
@@ -1,0 +1,17 @@
+---
+"agents": minor
+---
+
+Throttle chat UI updates by default in `useAgentChat`
+
+Streaming writes chat state once per chunk, and each write re-renders. When
+chunks arrive in a burst — a resumed stream replaying a long turn, for
+example — React reaches its 50-render limit and throws "Maximum update depth
+exceeded", which the AI SDK reports as a failed turn even though the server
+completed it (#1913).
+
+`useAgentChat` now coalesces those updates every 50ms, which removes about 78%
+of renders on a fast stream and matches the value the AI SDK documents. The
+first chunk of a stream is never delayed. Pass `throttle: false` to render
+every chunk as it arrives, or a number to change the interval. The deprecated
+`experimental_throttle` is still honoured.
--- a/.changeset/silent-paws-shine.md
+++ b/.changeset/silent-paws-shine.md
@@ -14,4 +14,6 @@ completed it (#1913).
 of renders on a fast stream and matches the value the AI SDK documents. The
 first chunk of a stream is never delayed. Pass `throttle: false` to render
 every chunk as it arrives, or a number to change the interval. The deprecated
-`experimental_throttle` is still honoured.
+`experimental_throttle` is still honoured. Message snapshots, functional
+updates, and streamed continuations resolve against the current chat store, so
+coalescing renders cannot roll assistant content back to an older snapshot.
--- a/packages/agents/src/chat/__tests__/broadcast-state.test.ts
+++ b/packages/agents/src/chat/__tests__/broadcast-state.test.ts
@@ -243,37 +243,35 @@ describe("broadcast stream state machine", () => {
 
   // ── continuation response ────────────────────────────────────────
 
-  it("creates accumulator with existing parts for continuation", () => {
-    const messages = makeMessages("hi", "first response");
+  it("seeds a continuation from the current messages update", () => {
+    const currentMessages = makeMessages("hi", "current response");
+    const staleRenderedMessages = makeMessages("hi", "stale response");
 
     const result = transition(idle, {
       type: "response",
       streamId: "s1",
       messageId: "fallback-id",
-      chunkData: textChunk(" continued"),
+      chunkData: { type: "text-delta", delta: " continued" },
       continuation: true,
-      currentMessages: messages
+      currentMessages: staleRenderedMessages
     });
 
     expect(result.state.status).toBe("observing");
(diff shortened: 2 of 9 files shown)


ben-reitz:
☝🏼 Looks like they're actually turning on `throttle` by default (with 50ms as the default value) in the next version of the `ai` SDK, partly off the back of this PR...

PR #2084, opened 2026-08-10
fix(skills): run compiled scripts loaded from R2

### LINKED ISSUE #2068: Skill scripts served from R2 cannot run: both documented paths fail
**Packages:** `agents@0.20.1` (via `@cloudflare/think@0.15.1`)

Agent Skills documents two ways to give a dynamically-sourced skill an executable
script. Neither works. A skill loaded through `skills.r2()` can only run a script
that ignores both instructions.

## The documented paths

**1. Ship TypeScript.** The Think README states:

> TypeScript scripts are compiled with `@cloudflare/worker-bundler`.

There is no such compilation at runtime. `@cloudflare/worker-bundler` is not a
dependency of `agents` and appears nowhere in `dist/`. A `.ts` entry resolves to
runtime `typescript`, so `prepareJavaScriptSource()` falls past its
`runtime === "javascript"` branch and throws:

> Skill script "…" must be compiled to a self-contained JavaScript module before
> it can run.

**2. Compile before upload.** The error above, and `src/skills/compile.ts`'s own
module docstring, both direct you to the same fix:

> skills served from R2 or other dynamic sources should be compiled with
> `compileSkillScript` before upload

Following that instruction produces a *worse* failure than ignoring it.
`compileSkillScript` emits esbuild's default-export form:

```js
function run(input, ctx) { /* … */ }
export { run as default };
```

`prepareJavaScriptSource()` routes a compiled bundle to `rewriteBundledSource()`,
which strips the export **and rebinds the default** (`const __skillRun = run;`).
That branch is gated on `resource.precompiled === true`.

**`skills.r2()` never sets `precompiled`.** It derives every resource descriptor
from the object path alone — `resourceKind()`, `resourceEncoding()`,
`resourceMimeType()`, plus `size`. `listAllObjects()` calls
`bucket.list({ prefix, cursor })` without `include: ["customMetadata"]`, so
there is no channel for the flag to arrive on either.

So the compiled bundle misses the `precompiled` branch and takes the single-file
raw branch instead. `scriptModule()` then applies:

```js
stripStrayExports(source.replace(/^\s*export\s+default\s+/m, "const __skillRun = "))
```

The regex does not match `export { run as default }`, so nothing is rebound, and
`stripStrayExports()` deletes the export line. Executed against a real Worker
Loader, the compiled bundle fails with:

> Skill script failed: Skill script default export must be a function (input, ctx).

The script *does* have a default export. The runner cannot see it in the form its
own compiler emits, and reports the opposite.

## The loop

Executed end to end against a live `worker_loaders` binding:

| Script served from R2 | Result |
| --- | --- |
| `.ts`, `export default` | `must be compiled to a self-contained JavaScript module … e.g. with compileSkillScript` |
| `.js` compiled by `compileSkillScript` | `Skill script default export must be a function (input, ctx).` |

The first error instructs you to produce the second.

## What this costs

`compileSkillScript` is the bundler — turning several modules into one
self-contained file is the whole reason it exists. So while the symptom is "a
compiled bundle will not run", the consequence is broader: **a dynamically-sourced
JavaScript or TypeScript skill script cannot span more than one file.** The only
shape that runs is a single `.js` that imports nothing, so any skill whose logic
does not fit comfortably in one module has no path at all.

The asymmetry is the sharp part. A multi-file *Python* skill from the same R2
source runs today, sibling imports and all, because `runPythonScript()` never
touches this code path and materializes every resource onto the filesystem. So the
runtime with the weaker story upstream is the only one that can be structured, and
a team that wants modules is pushed to Python by a bug rather than by a decision.

## Reproduction

The descriptor drop reproduces directly (vitest, workers pool):

```ts
import { r2 } from "agents/skills";

const files = new Map([
  ["skills/demo/SKILL.md", "---\nname: demo\ndescription: demo skill\n---\nbody"],
  // exactly what compileSkillScript emits
  ["skills/demo/scripts/run.js",
   "function run(i,c){return{ok:true}}\nexport { run as default };"],
]);

const bucket = {
  async list({ prefix }) {
    return {
      objects: [...files.keys()].filter(k => k.startsWith(prefix))
        .map(key => ({ key, size: files.get(key).length })),
      truncated: false,
    };
  },
  async get(key) {
    const body = files.get(key);
    return body === undefined ? null : { text: async () => body };
  },
};

const content = await r2(bucket, { prefix: "skills/" }).load("demo");
const script = content.resources.find(r => r.path.endsWith("run.js"));

script.kind;        // "script"
script.precompiled; // undefined  <-- the compiled bundle is not recognised
```

## Expected

Either `skills.r2()` propagates `precompiled` for compiled scripts — via
`customMetadata` on the R2 object, or by treating `compileSkillScript` output as
recognisable — or `scriptModule()` handles the `export { x as default }` form the
way `rewriteBundledSource()` does, so the binding survives on both branches.

Also worth correcting: the README's `@cloudflare/worker-bundler` sentence
describes runtime compilation that does not exist, which is what leads a reader
to upload TypeScript in the first place.

## Notes

- Every result above was executed against a live `worker_loaders` binding, not
  inferred. A plain single-file `.js` using `export default` runs fine from the
  same source, which is the only JavaScript shape that currently works.
- Python is unaffected. `runPythonScript()` bypasses this path entirely and
  materializes every resource onto the Pyodide filesystem, so a multi-file Python
  skill served from R2 runs today — verified, including a sibling module import
  resolved from the documented `/skill` mount point.

ben-reitz:
Fixes #2068.

## What is the bug

A skill loaded through `skills.r2()` cannot run the output of `compileSkillScript()`, even though compiling before upload is the documented path for TypeScript and multi-file scripts.

Raw TypeScript tells the user to compile first. The compiled JavaScript then fails because the runner reports that its default export is not a function.

## Why it happens

`compileSkillScript()` emits esbuild's ESM export-list form:

```js
function run(input) {
  return input.text.toUpperCase();
}

export { run as default };
```

The runner already knows how to turn that export back into its internal `__skillRun` binding, but only when the resource descriptor says `precompiled: true`. The Vite plugin preserves that flag; R2 (and other dynamic sources) might only preserve the actual compiled content.

Without the flag, a `.js` bundle takes the raw JavaScript path, where its export list is removed without rebinding `run`. A bundle stored under a `.ts` path, or next to sibling script resources, is rejected as if it still needed compilation.

## The fix

`compileSkillScript()` now stamps its output with a versioned first-line header:

```js
// cloudflare-agents:compiled-skill-script:v1
```

When a dynamic source does not have a `precompiled` descriptor, the runner requires both that header and esbuild's final bundled default-export block before normalizing the script. Recognition happens before the file-extension and sibling-script checks, so header-marked output can keep its original `.ts` path and resource list. The marker check tolerates a UTF-8 BOM or CRLF line endings added by publish tooling.

The compiler removes an entry-point shebang from the generated bundle because skill code is embedded inside a function wrapper rather than launched as a CLI. The existing `precompiled` path remains in place for Vite manifests and sources that can preserve the descriptor. This also keeps the runtime compiler-free: raw TypeScript and unbundled multi-file scripts are still rejected. Dynamic publishers compile in Node and upload `compiled.content` unchanged.

## Why use a header

Recognizing `export { binding as default }` alone was too guessy. That text could appear anywhere, including inside a comment or string, and should not be enough to opt a file into the compiled path.

The header makes this an explicit contract between `compileSkillScript()` and the runner. Unrelated JavaScript syntax can no longer accidentally look precompiled. Keeping the export-shape check as well makes sure the marked file still has the structure the runner knows how to normalise.

## Documentation

The Think docs, Agent Skills example, and skills design notes now describe build-time esbuild compilation instead of the removed runtime `@cloudflare/worker-bundler` behavior. Dynamic-source examples preserve the generated content, including its header.

## Tests

Regression coverage drives the public path from a fake R2 bucket through `skills.r2()`, `SkillRegistry`, `skills.runner()`, and a live Worker Loader. It covers header-marked content stored as `.js`, content stored as `.ts`, a compiled entry with a sibling TypeScript resource, and BOM/CRLF-normalized uploads.

The compiler tests pin the generated header and esbuild default-export shape and verify that retained shebangs are removed. Runner tests verify that unmarked export-shaped source does not bypass the compile-required path, while export-shaped text inside a marked bundle's comments or strings is preserved.

Suites: skills runner/R2 49 · compiler/Vite 14 · full agents 2554 · `pnpm run check`.

### DIFF
--- a/.changeset/swift-skills-bundle.md
+++ b/.changeset/swift-skills-bundle.md
@@ -1,0 +1,5 @@
+---
+"agents": patch
+---
+
+Fix precompiled JavaScript and TypeScript skill scripts loaded from R2 or other dynamic sources. Output from `compileSkillScript` can now run without separately preserving the `precompiled` resource descriptor.
--- a/.changeset/swift-skills-bundle.md
+++ b/.changeset/swift-skills-bundle.md
@@ -2,4 +2,4 @@
 "agents": patch
 ---
 
-Fix precompiled JavaScript and TypeScript skill scripts loaded from R2 or other dynamic sources. Output from `compileSkillScript` can now run without separately preserving the `precompiled` resource descriptor.
+Fix precompiled JavaScript and TypeScript skill scripts loaded from R2 or other dynamic sources. Header-marked output from `compileSkillScript` can now run without separately preserving the `precompiled` resource descriptor.
--- a/design/skills.md
+++ b/design/skills.md
@@ -255,9 +255,10 @@ returned by `run_skill_script` without mutating the workspace. The Agents Vite
 plugin compiles bundled JavaScript and TypeScript with esbuild before deployment.
 R2 and other dynamic sources use `compileSkillScript` in Node-based publish
 tooling and store its self-contained ESM output unchanged. The compiler prepends
-an Agents skill-script format header, and the runner requires that header plus
-esbuild's `export { binding as default }` bundle form. Dynamic sources therefore
-do not need to preserve the optional `precompiled` resource descriptor.
+an Agents skill-script format header that selects the compiled runner path; the
+runner then normalizes esbuild's expected `export { binding as default }` bundle
+form into its function wrapper. Dynamic sources therefore do not need to
+preserve the optional `precompiled` resource descriptor.
 
 Python (`.py`) and Bash (`.sh`/`.bash`) keep the path-based contract, matching
 CLI-oriented Agent Skills. The runner mounts:
--- a/design/skills.md
+++ b/design/skills.md
@@ -254,9 +254,10 @@ tools; and `ctx.output.writeFile(name, content)` records scratch artifacts
 returned by `run_skill_script` without mutating the workspace. The Agents Vite
 plugin compiles bundled JavaScript and TypeScript with esbuild before deployment.
 R2 and other dynamic sources use `compileSkillScript` in Node-based publish
-tooling and store its self-contained ESM output. The runner recognizes esbuild's
-`export { binding as default }` bundle form, so dynamic sources do not need to
-preserve the optional `precompiled` resource descriptor.
+tooling and store its self-contained ESM output unchanged. The compiler prepends
+an Agents skill-script format header, and the runner requires that header plus
+esbuild's `export { binding as default }` bundle form. Dynamic sources therefore
+do not need to preserve the optional `precompiled` resource descriptor.
 
 Python (`.py`) and Bash (`.sh`/`.bash`) keep the path-based contract, matching
 CLI-oriented Agent Skills. The runner mounts:
--- a/design/skills.md
+++ b/design/skills.md
@@ -251,8 +251,12 @@ removed); `ctx.workspace` exposes async `readFile`/`listFiles`/`glob`/`stat`/
 `writeFile` gated by the workspace permission; `ctx.tools` exposes
 `tools.call(name, input)` and `tools.<name>(input)` for explicitly granted
 tools; and `ctx.output.writeFile(name, content)` records scratch artifacts
-returned by `run_skill_script` without mutating the workspace. TypeScript and
-multi-file scripts are compiled/bundled with `@cloudflare/worker-bundler`.
+returned by `run_skill_script` without mutating the workspace. The Agents Vite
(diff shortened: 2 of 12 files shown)


ben-reitz:
@aron-cf's looking at re-working this entirely, so closing for now as this isn't a great solution anyway

PR #2093, opened 2026-08-12
chore(worker-bundler): General tidy-up and move Python logic into a separate file

abstractedfox:
General refactor; cleans up a few vestiges of earlier work, removes outdated comments, and should make the code a little more descriptive. Also moves the Python logic into a separate file per an earlier review comment from @ryanking13 (https://github.com/abstractedfox/agents/pull/6#discussion_r3643456980)

I deliberately left the commits intact to make it easier to review, the first one is just the move into a new file and the second one has all the little changes I made afterward

### DIFF
--- a/packages/worker-bundler/src/installer-python.ts
+++ b/packages/worker-bundler/src/installer-python.ts
@@ -21,7 +21,7 @@ export interface PyprojectToml {
   };
 }
 
-interface PypiSimpleFile {
+interface PyPASimpleFile {
   filename: string;
   url: string;
   hashes?: Record<string, string>;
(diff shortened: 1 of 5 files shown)


packages/worker-bundler/src/installer-python.ts
```diff
@@ -0,0 +1,594 @@
+import {
+  isTextFile,
+  fetchWithTimeout,
+  DEFAULT_TIMEOUT_MS
+} from "./installer.ts";
```
ryanking13: As this is a refactoring PR, I would say cyclic imports between installer.ts and installer-python.ts is bad.

I think JS bundler handles this, but it would be probably better to extract out these common functions into a separate helper file.

ryanking13 (approved):
I left a minor comment about code structure, otherwise looks good!

ryanking13:
Also please update the PR title to be more descriptive

packages/worker-bundler/src/installer-python.ts
abstractedfox: Ahh, good catch, thank you

PR #2100, opened 2026-08-12
Remove skill runner

aron-cf:
Agent Skills currently combine loading task instructions with optional script execution. This makes
 `SkillRegistry` responsible for runtime permissions and adds a runner, build-time compilation, and
 execution-specific dependencies to an otherwise read-only resource system.

 This change removes `skills.runner`, `run_skill_script`,  `agents/skills/compile`, execution-only
types, and the Think runner hook.  Vite now preserves script files as inert resources. The
first-party examples  use `read_skill_resource` or existing workspace tools instead. The Turndown
stub remains because Think's workspace Bash tool still imports `just-bash`.

 The package builds and repository checks pass. The changeset, package README, examples, and Think guides document the breaking API removal and direct applications toward explicit tools when they need execution.


ben-reitz (approved):
You think it's worth adding an example for how people can allow their agents to run scripts that are included in skills now?

(I don't know how commonly skills include scripts, but it is part of the skills spec...)

PR #2156, opened 2026-08-24
feat(agents): WebSockets speaks the Agent protocol for plain hosts; Cap'n Web transport as a socket class

mattzcarey:
## What

Makes a plain `DurableObject` composed with the `WebSockets` capability work with `useAgent` and `AgentClient` exactly like an `Agent`, on either wire.

**Server — the capability speaks the Agent protocol.**
- On connect it sends the identity frame (`cf_agent_identity`), which resolves the client's `ready`/`identified`. `identity: false` opts out; `Agent` passes it because it sends its own under `sendIdentityOnConnect`.
- It answers `rpc` frames (what `call()` and `stub` send) against `callables`, on both wires. A `ReadableStream` result streams as `done: false` chunks then a final `done: true`. `Agent`'s decorated target is marked host-served so `Agent.onMessage` keeps answering its own frames (facet bridging, `StreamingResponse`).
- The experimental `?__agents_rpc=capnweb` callables endpoint from 0.23.0 is **removed**. `useAgent().stub`/`.call()` reach the same `RpcTarget` on either transport, so it had no remaining purpose. `callablesFromDecorated` goes with it: `Agent` answers its own `rpc` frames and sets neither `callables` nor `identity` on the capability.
- `LifecycleServices` gains `name` and `className`.

**Cap'n Web transport — a socket class, not a parallel client.**
- `CapnWebSocket` is a WebSocket-shaped class over a Cap'n Web session. `useAgent({ transport: "capnweb" })` and `AgentClient({ transport })` pass it to PartySocket as its `WebSocket` option, so PartySocket keeps owning reconnection, buffering, and backoff. No second hook, no second client, no native call path.
- The transport also rides in the query, so PartySocket's socket key changes and it reconnects when `transport` changes.
- The server session root carries exactly one method, the frame pipe. A session is registered before `onConnect` runs, so `getConnections()` includes it during the connect handler, same as a hibernating socket. Handlers are wire-agnostic.
- Default is `"cf-websocket"`. Cap'n Web connections do not hibernate.

## Example: `examples/next/websockets`

One plain `RoomObject`, one `RpcTarget`, two transports. The React client opens the same room twice with `useAgent`, once per transport, and calls `stub.say()`, `stub.members()`, and a streaming `call("countdown")` on each. The README has the transport comparison table. Tests cover identity, rpc frames including a streamed result, the transport pipe, cross-wire broadcast, tags, and the HTTP push path.

## Compared with the previous revision of this PR

Removed `capnweb-client.ts` (hand-rolled reconnect), `use-capnweb-socket.ts` (parallel hook), the native `callNativeCapnWeb` path in `react.tsx`, callables on the session root, and the legacy streaming projection. Every Devin finding on the earlier revision was in code that no longer exists. The previous working tree is preserved as checkpoint commit `47ed21ab`.

## Docs

- `docs/agents/lifecycle.md`: WebSockets section rewritten — plain hosts with `useAgent`, and the two wires.
- `docs/agents/client-sdk.md`: Transport section.

## Testing

- Workers project: 123 files, 1900 tests pass, including the transport e2e suite and plain-host tests (identity, `rpc` frames, streaming, unknown method, non-rpc passthrough) on both wires.
- React browser: `useAgent` capnweb suite and client suite, 43 tests pass.
- Example: 6 tests pass; Vite client builds with no externalized Node modules.
- `sherif`, `oxlint`, `oxfmt --check`, root typecheck (122 projects) clean.

## Follow-up

State sync for plain hosts: once #2179 (`agents/state`) lands, `WebSockets` takes a `state` option and pushes `cf_agent_state` on connect, routes incoming state frames to `state.set(value, connection)`, and broadcasts changes excluding the source. #2246 (`examples/next/routing`) stacks on this branch.


mattzcarey:
Superseded by https://github.com/cloudflare/agents/pull/2249: same branch content, head moved from the fork to cloudflare/agents so CI runs on origin. The previous revision is preserved as checkpoint commit 47ed21ab.

PR #2168, opened 2026-08-27
feat(tasks): durable replayable execution as a Lifecycle capability (agents/tasks)

mattzcarey:
Adds `agents/fibers` — durable, replayable background execution as one Lifecycle capability. This is Phases 1–4 of the accepted durable-execution RFC (`design/rfc-fibers.md`), **shipped under the name Tasks** (`agents/tasks`) — "fiber" stays the legacy engine's vocabulary (`runFiber`/`startFiber`), so one word no longer means two things in the package. Built directly on the Lifecycle capability contract and alarm-contribution model that #1897 shipped. A companion `design/rfc-streams.md` (proposed) sketches the Streams capability — the durable chunk-log half of the pattern the Think migration validated.

## What this adds

**The capability.** One `Tasks` instance per Durable Object declares its named definitions in the constructor — a Scheduler-style map, so the registry is rebuilt on every wake and recovery of in-flight runs is correct by construction:

```ts
export class ReportObject extends DurableObject<Env> {
  readonly tasks = new Tasks({
    definitions: {
      "build-report@v1": async (input: ReportInput, step: TaskStep) => {
        const research = await step.do(
          "research",
          { retries: { limit: 4, delay: "2 seconds", backoff: "exponential" } },
          ({ signal }) => this.research(input.topic, { signal })
        );
        await step.sleep("editorial-delay", "30 seconds");
        await step.do("publish", ({ idempotencyKey }) =>
          this.publish(input.reportId, research, { idempotencyKey })
        );
        return { reportId: input.reportId };
      }
    }
  });
  readonly lifecycle = Lifecycle.install(this).use(this.fibers);
}

const receipt = await this.tasks.run("build-report@v1", input, {
  idempotencyKey: `report:${input.reportId}`
});
```

`tasks.run(name, input, options)` and `tasks.handle(name)` (a typed lens scoped to one definition) are typed against the map. Framework-internal definitions attach through an internal composition-root resolver aperture mirroring the Scheduler's callback-name resolver.

**Replay as the execution model.** Every attempt runs the handler from the top: completed `step.do()` steps return journaled results without re-executing, sleeps consult their first persisted deadline, and execution reaches the first unfinished step. Process loss leaves a claimed run whose claim deadline wakes the object; the interrupted step runs again under a fresh generation (at-least-once, with stable per-step idempotency keys for downstream dedup).

**The rest of the surface.** Durable acceptance with receipts and idempotency-key joins; per-step retry/timeout policy with `NonRetryableError`; `step.status()` progress behind a replay live gate (old progress is never re-published as new); cooperative cancellation; `get`/`getByIdempotencyKey`/`list`/`delete` inspection; retention control; visible failures for journal divergence, duplicate step names, and definitions missing after a deployment.

**Custom recovery (Phase 2).** A definition may pair its handler with a `recover` callback — `{ run, recover }` in the map — that owns unclean interruptions instead of automatic replay. It receives the run input, metadata, and the interrupted step (name, attempt, stable idempotency key, and the last `checkpoint()` the lost attempt wrote), and decides `replay` (immediate or deferred), `complete`, `fail`, or `cancel`. Clean step failures never reach recovery — the retry policy owns them. Recovery is itself interruption-safe (a `recovering` claim with the same generation fencing and claim-deadline backstop), and a throwing `recover` retries on exponential backoff with a bounded budget before the run fails.

**Agent integration (Phase 3).** `Agent` installs the capability automatically as experimental `this.tasks`. Subclass definitions go on the overridable `taskDefinitions` field (rebuilt every wake, resolved lazily so field order never matters); framework-internal definitions attach through a composition-root aperture mirroring the Scheduler's callback resolver; handlers run in the Agent's invocation context; and interrupted runs dispatch at the same startup point as the legacy fiber scan — before the user's `onStart`.

**Think + AIChatAgent replatform (Phase 4).** Chat turns in both frameworks and Think's messenger replies now execute on the capability: the live closure runs as one journaled step in the caller's invocation context with checkpoint-backed `stash()`, and an unclean interruption synthesizes the legacy `FiberRecoveryContext` and routes through the **unchanged** `_handleInternalFiberRecovery` → ChatRecoveryEngine seam (messenger recovery likewise, with re-entry checkpoints persisted in host storage). The recovery brain did not move — which is why the entire think suite passes untouched. Legacy `runFiber()`/`startFiber()` user APIs are unchanged and still recovered by their own scan; facet-hosted turns stay on the legacy engine until routed Tasks land.

**Alarm integration.** Tasks follows the Scheduler pattern exactly: run deadlines live in `cf_agents_task_runs.next_at`, `getNextAlarm()` contributes the minimum, `onAlarm()` claims due runs in a bounded batch, and every deadline change calls `alarms.rearm()`. It never touches the physical alarm.

## The architecture decision

The original RFC draft proposed a central Lifecycle-owned `cf_lifecycle_work` trigger table with claim/lease fencing and an `onWork()` dispatch hook. The RFC as landed (`design/rfc-fibers.md`, status: accepted, amended) rejects that in favor of the shipped alarm-contribution model — the dispatcher duplicated state the capability already owns, doubled the fencing layers, and would have re-architected arbitration #1897 just shipped and regression-tested. The full reasoning is recorded under Alternatives considered.

Also included: `examples/next/tasks`, a plain-Durable-Object example smoke-tested end to end under `wrangler dev` (draft → durable sleep through the real alarm → publish, with idempotency-key joins and cancellation over HTTP).

## Deliberately not in this PR (per the RFC)

- Delegating the released `runFiber()`/`startFiber()` user APIs onto the new engine — they keep their exact behavior and their own recovery scan; deprecation waits for migration evidence.
- `waitForCompletion` on `run()` — open question 9 (input gates / leaked waiters) deserves a workerd-verified answer first; internal frameworks use an attached internal start instead.
- Runs on routed sub-agents — `run()` throws a clear error; facet-hosted chat turns stay on the legacy engine until the routed protocol lands.

The whole surface is `@experimental` and documented as such (`docs/agents/fibers.md`).

## Verification

- 30 new fiber tests on real Durable Objects (26 capability-level plus 4 Agent-level, including shared-alarm coexistence with Agent schedules and interruption reclaim on a real Agent) with real Lifecycle startup, real SQLite, and real platform alarms — including replay memoization proven via instance counters (journal hits never execute), interrupted-attempt reclaim from a seeded dead generation, sleep first-deadline authority, the status live gate, cancellation fencing, Scheduler alarm coexistence, the bounded alarm batch, and every recovery path (complete/replay/deferred-replay/fail/cancel decisions, clean-failure isolation from recovery, checkpoint persistence, and the throwing-recover backoff and budget).
- Full suites on the migrated engine: agents 1,938/1,938, **think 887/887**, ai-chat 655/655 — the think suite passing unchanged is the strongest evidence the recovery seam held.
- `sherif`, `check:exports` (with rebuilt dist), repo-wide `oxfmt --check`, `oxlint`, and the root typecheck all pass.


mattzcarey:
Superseded by #2173, which now carries the full Tasks + Streams + chat-replatform work rebased onto the Lifecycle work-queue rework (#2175) and the WebSockets capability extraction (#2169), with the API slimmed to pure replay after end-to-end use.

PR #2179, opened 2026-08-28
feat(agents): move state into an opt-in Lifecycle capability (agents/state)

AntoniTok:
State was one method (`_setStateInternal`) doing four jobs inside the Agent god-class — validate, persist, broadcast, notify — with the state row, in-memory cache, and schema all threaded through the class. This moves storage and change ordering into a `State` capability (`agents/state`), so any Lifecycle host gets durable, validated state without inheriting `Agent`. Same pattern as the WebSockets (#2169) and MCP client (#1895) capabilities.

Agent's public API and wire protocol are unchanged.

### API

```ts
import { State } from "agents/state";

class Counter extends DurableObject {
  readonly state = new State<{ count: number }>({
    initialState: { count: 0 },
    validateStateChange: (next, source) => { /* throw to reject */ },
    onChanged: (state, source) => { /* sync or async */ }
  });
  readonly lifecycle = Lifecycle.install(this).use(this.state);
}

counter.state.get();            // { count: number } | undefined
counter.state.set({ count: 1 }); // "server" source
counter.state.set(next, connection); // client-originated
```

`StateOptions` is plain data: `initialState`, `validateStateChange`, `onChanged`. `StateChangeSource` is `Connection | "server"`. Async `onChanged` is tracked and its rejection logged; `set()` stays synchronous. The capability reaches Lifecycle only for `storage` and uses `#` private fields like its siblings.

### Architecture

```
+--------------------------------------+
| Agent                                |
|  get state -> _state.get()           |
|    (seeds initialState if undefined) |
|  setState  -> _state.set()           |
|  _handleStateChanged():              |
|    3. broadcast --------> clients    |
|    4. notify hook                    |
|         ^ onChanged                  |
+---------|----------------------------+
+---------+----------------------------+
| State (capability)                   |
|  onStart: own schema version key,    |
|           create table, legacy clean |
|  get(): load / cache                 |
|  set(): 1.validate 2.persist         |
|         3. onChanged(state, source)  |
+---------+----------------------------+
          | this.lifecycle.sql / storage
          v
   cf_agents_state (SQLite)
```

The capability never references connections, the Agent, env, or ctx. Agent passes two closures (`validateStateChange`, `onChanged`) as static options and does the WebSockets-specific work itself. Agent's generic parameter is renamed `State` → `TState` (positional for consumers, so not a breaking change) so the class can be imported as plain `State`.

### Data flows

Server sets state (`source = "server"`, broadcast to all):

```
setState(next) -> _state.set(next, "server")
  1. validate  (Agent.validateStateChange)
  2. persist   -> cf_agents_state
  3. onChanged -> Agent._handleStateChanged(next, "server")
       broadcast CF_AGENT_STATE to all protocol clients
       waitUntil -> onStateChanged / onStateUpdate dev hook
```

Client sends state over WS (`source = connection`, echo excludes sender):

```
onMessage (stays in Agent): parse, readonly check, CF_AGENT_STATE_ERROR
  -> _state.set(next, connection)
       1. validate  2. persist
       3. onChanged -> _handleStateChanged(next, connection)
            broadcast to all EXCEPT connection.id
            dev hook
```

### What moved vs. what stays

| Concern | Owner |
|---|---|
| `cf_agents_state` table, state row, load/cache, validate + persist | State |
| legacy `cf_state_was_changed` cleanup | State (its v1 migration) |
| own schema version key (`cf_agents:state_schema_version`) | State |
| `initialState` field and first-access seeding | Agent (`state` getter) |
| broadcast to connections | Agent (`_handleStateChanged`) |
| dev hooks (`validateStateChange`, `onStateChanged`/`onStateUpdate`) | Agent (override surface) |
| `onMessage` parse / readonly / error responses | Agent / WebSockets |

`initialState` stays an Agent field because subclass fields initialize after the base constructor. Rather than the capability reaching back into the host lazily, Agent's `state` getter seeds it on first read when the capability returns `undefined` (not JSON-representable, so it uniquely means "no row"). The seed goes through `set()`, so it persists, broadcasts, and runs the hook as before.

### Table ownership

`State` is the only owner of `cf_agents_state`. Agent used to keep its global schema version as a row in that table, which was the only reason it still created the table. That version now lives under the `cf_agents:schema_version` KV key, read synchronously via `storage.kv` so the constructor can still gate DDL — the same convention every capability uses. A DO created under the old layout has the row read once, moved to the key, and deleted on its next construction.

### Compatibility

`state`, `setState()`, `onStateChanged`, and the `CF_AGENT_STATE` frames behave identically. The full Agent state suite and the schema suite pass; the fixture reach-ins to the removed `_state` cache are replaced by eviction (`evictDurableObject`) or by driving the capability's own migration.

### Tests

Capability suite installs `State` on a bare Durable Object through `withCapabilityHarness`: persist/read, access before lifecycle start, initial-state seeding, falsy-value row existence, rehydration across eviction, validation rejection, `onChanged` source on both origins, sync/async hook failures. Schema suite adds a legacy schema-version-row migration case.


packages/agents/src/index.ts
```diff
@@ -1792,55 +1829,25 @@ export class Agent<
...
-    // update db and return the initial state
-    this._setStateInternal(this.initialState);
-    return this.initialState;
+  /**
+   * @internal Drop the StateManager's in-memory cache so the next `state`
+   * read reloads from storage and re-seeds the initial state when the row is
+   * absent. Mirrors a hibernation wake-up; for host-internal use and tests
+   * that exercise the lazy-load path in a single live instance.
+   */
+  __DO_NOT_USE__resetStateCacheForTesting(): void {
```
mattzcarey: no testing code in the main class

packages/agents/src/index.ts
```diff
@@ -2413,9 +2425,18 @@ export class Agent<
     this.lifecycle
       .use(this.scheduler)
       .use(this.mcp)
+      .use(this.#state)
```
mattzcarey: this._state.

packages/agents/src/index.ts
```diff
@@ -1570,6 +1576,38 @@ export class Agent<
...
+   * is invariant. Holding it as a `StateManager<State>` field would propagate
+   * that invariance to `Agent`'s own `State` parameter and break
+   * `Subclass -> Agent<Env, unknown>` assignability for every consumer
+   * (sub-agents, `DurableObjectNamespace<Subclass>`, etc.). Erasing to
+   * `unknown` keeps the field's runtime typing intact while leaving
+   * `Agent<State>` as variance-compatible as it was before this capability
+   * existed; the typed `State` boundary is re-established at the delegating
+   * call sites below.
+   */
+  readonly #state: StateManager<unknown> = new StateManager<State>({
```
mattzcarey: call this State. 


let me pass a hook to do stuff on changed. and pass the initial state. dont need a setter I think it can be static.

packages/agents/src/index.ts
```diff
@@ -2999,33 +3021,47 @@ export class Agent<
...
+    const { state: nextState, sourceId } = event;
 
     // Broadcast state to protocol-enabled connections, excluding the source
     this._broadcastProtocol(
       JSON.stringify({
         state: nextState,
         type: MessageType.CF_AGENT_STATE
       }),
-      source !== "server" ? [source.id] : []
+      sourceId !== undefined ? [sourceId] : []
```
mattzcarey: huh?

packages/agents/src/index.ts
```diff
@@ -2999,33 +3021,47 @@ export class Agent<
...
     );
 
     // Notification hook (non-gating). Run after broadcast and do not block.
     // Use waitUntil for reliability after the handler returns.
     const { connection, request, email } = agentContext.getStore() || {};
+    const source: Connection | "server" =
```
mattzcarey: 🪦

AntoniTok:
- Rebased onto current main; PR is mergeable.
- Moved state storage, caching, validation, and notifications into StateManager.
- _state is public.
- Added direct initialState and onChanged options.
- Preserved the real state-change source.
- Removed _setStateInternal.
- Removed all shipped test-only cache reset methods.
- Exposed the Lifecycle SQL helper.
- Added public agents/state export and type tests.
- State table is ensured before Lifecycle startup.
- onChanged supports sync and async hooks; errors are handled while setState() stays synchronous.
- Tests now use real eviction or fresh instances.
- Added regression tests; all 2,001 Workers tests pass locally.

AntoniTok:
I assume you didn’t want me to change StateManager to State because State already names the agent’s state type and would just be confusing. Anyway #state is now _state.

Anyways now initialState passes the starting value, and onChanged passes the hook that runs after state changes; no initial-state setter is needed.

The wierd stuff marked by your last 2 comments was from splitting the state-change logic into two functions connected by an event causing us to lose the original connection value, so the code had to fake it back. Fixed now.

Devin flagged that async onChanged promises were not handled. Agent state changes now use the existing ctx.waitUntil, while standalone StateManager instances catch rejected promises. No Lifecycle API or behavior changed.

mattzcarey:
Superseded by https://github.com/cloudflare/agents/pull/2253: same branch content with current main merged in, head moved from the fork to cloudflare/agents so CI runs on origin.

mattzcarey (approved):
Reviewed and iterated on this together: State capability owns cf_agents_state outright, Agent's schema version moved to a KV key, static options, TState rename, # privates. CI green, no open threads.

PR #2193, opened 2026-09-01
refactor(agents): extract facets into a dynamic-agents capability, reposition for isolation not chat sessions

mattzcarey:
## What this does

Facets ("sub-agents") had grown into ~2,400 of `index.ts`'s ~12,150 lines, inlined in the `Agent` god class, and the docs positioned them as the recommended way to model many chat sessions per user. Investigation (workerd source, cloudflare-os, and this repo's git history of facet-lifecycle bugs) says otherwise: facets are a **colocated isolation primitive** — separate isolate, own SQLite, no independent alarms, machine-pinned tree — built for code the parent *supervises* (dynamically-loaded/generated code, per-run tool agents, sandboxes), not for an open-ended set of independent peers.

### 1. Extraction (zero behavior change)
Moves the facet machinery into `packages/agents/src/dynamic-agents/` — types, identity, RPC bridges, the sub-agent registry, lifecycle routing, resolution/teardown/fiber-recovery, and WebSocket forwarding/virtual connections — behind an explicit host port. Registered as a Lifecycle capability (`capabilityId: "dynamic-agents"`); its hot paths stay composition-root wired because the capability-runner hook contract can't express request-rewrite-and-continue or post-claim WebSocket forwarding (documented on the class). No wire- or storage-visible identifier changes. `index.ts`: 12,154 → ~10,480 lines.

### 2. Public API: `this.dynamicAgents`
The new surface: `this.dynamicAgents.{get,abort,delete,has,list}` on every `Agent`, plus the `DynamicAgentClass` and `DynamicAgentStub` type names, all marked `@experimental`. `subAgent()` / `abortSubAgent()` / `deleteSubAgent()` / `hasSubAgent()` / `listSubAgents()` are unchanged in behavior, `@deprecated` in place, and now delegate through the same capability. This delegation also replatforms inherited Think and AIChatAgent facet usage without changing either package or their chat topology. `/sub/` URLs, `useAgent({ sub })`, `parentAgent()`, and `onBeforeSubAgent` are untouched. Runtime and type-level tests pin the new facade and typed child stub.

### 3. Docs (`docs/agents/sub-agents.md` rewrite)
Verified workerd semantics section, a corrected claim (frames do **not** flow directly to the child post-upgrade — every frame wakes the root parent and is forwarded over RPC), and an explicit decision rule: *a facet is a child whose code/lifecycle the parent supervises and which must live inside the parent; an independent peer you address by name should be its own top-level Durable Object.*

### 4. Two new examples, both with React + Vite UIs and workers-pool tests
- **`examples/next/dynamic-agents`** — what facets are for: a supervisor stores user-submitted Durable Object code, loads it via Worker Loader, and runs it as a facet with isolated storage, supervised abort, and code upgrades over stable state. The editor loads the selected gadget's stored source before deployment.
- **`examples/next/chats`** — the recommended many-chats topology: one top-level `ChatAgent` DO per conversation plus a per-user `UserAgent` index DO. Listing and cross-chat search read only the index; each Chat pushes title/preview metadata best-effort after writes; the User DO assigns deterministic activity order and ignores older delivered timestamps; deletion is `destroy()` plus one row. The example intentionally treats the index as an eventually consistent derived projection and leaves production idempotency, repair, deletion fencing, and User-gated routing to the separate RFC.

`examples/multi-ai-chat`'s README gets a one-paragraph caveat pointing at `examples/next/chats` for the many-chats case; it still works and is left as-is.

### 5. Future topology RFC, no implementation
`design/rfc-user-chat-durable-objects.md` proposes the follow-up architecture discussed here: one top-level User agent owns the catalog, shared workspace/memory, MCP registry/connections, OAuth callbacks, and cross-chat jobs; one independent top-level Chat agent owns each conversation. A browser normally keeps one User socket plus one socket for the active Chat. Switching chats replaces only the Chat socket. The RFC covers identity, User-gated routing, metadata consistency, deletion, repair, rollout from facet-backed apps, and separate `AIChatAgent` / `Think` verification. It explicitly does not implement that migration in this PR or propose a framework `Chats` base class.

## Verification
- `packages/agents`: full `test:workers` project green (1976 tests / 114 files), plus the earlier `ai-chat` (655 tests) and `think` package suites (both subclass `Agent` via protected members)
- Repo-wide: `pnpm run check` clean (`sherif`, exports, formatting, lint, and all 121 TypeScript projects)
- `examples/next/chats`: 10 tests green; `examples/next/dynamic-agents`: 5 tests green

## Not in this PR
No existing application or package is migrated from facet-backed chat sessions to top-level Chat DOs. `Think` and `AIChatAgent` remain functionally unchanged; inherited `subAgent()` calls simply delegate through the new capability. The possible User-agent + Chat-agent topology is documentation-only in this PR and requires separate implementation work.

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/agents/pull/2193" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


packages/agents/src/index.ts
```diff
@@ -36,6 +35,36 @@ export type {
...
+  RootFacetRpcSurface,
+  DynamicAgentClass as SubAgentClass,
+  DynamicAgentConnectionMeta as SubAgentConnectionMeta,
+  DynamicAgentPathInvokeEndpoint as SubAgentPathInvokeEndpoint,
+  DynamicAgentStub as SubAgentStub
+} from "./dynamic-agents/types";
+export type {
+  DynamicAgentClass as SubAgentClass,
+  DynamicAgentStub as SubAgentStub
+} from "./dynamic-agents/types";
```
ben-reitz, line 69: do we want to export these new types under the new names as well rather than **only** exporting them under the _legacy_ names?

.changeset/dynamic-agents-capability.md
```diff
@@ -0,0 +1,11 @@
+---
+"agents": patch
```
ben-reitz: should this be a `minor` release considering the new public API?

ben-reitz (approved):
the architectural change lgtm!

packages/agents/src/index.ts
mattzcarey: Yes — exported `DynamicAgentClass` and `DynamicAgentStub` from the main `agents` entry in e76fa00b, while retaining `SubAgentClass` / `SubAgentStub` as compatibility aliases. Added type-level coverage for both names.

.changeset/dynamic-agents-capability.md
mattzcarey: Agreed — changed the `agents` changeset from patch to minor in e76fa00b because `Agent#dynamicAgents` and the new type names are public API additions.

PR #2196, opened 2026-09-01
feat(agents): move sessions into a Lifecycle capability

mattzcarey:
## Summary

Adds the experimental `agents/sessions` Lifecycle capability and the `agents/context` module, and moves Think and AIChatAgent conversation persistence onto them.

Sessions owns durable conversation storage and nothing else: a tree of messages with branches and compaction overlays, streamed and byte-budgeted reads, opt-in full-text search, and a content-addressed attachment store.

The old `agents/experimental/memory` provider stack, its Postgres providers, and SessionManager are removed. A user-facing conversation directory remains the responsibility of a parent or router Durable Object.

## Nothing is truncated, and nothing is too large to store

Two independent mechanisms, split by **type** rather than by size.

**Attachments leave the message.** A part that declares a non-text media type and carries its bytes inline — an image, an audio clip, a PDF — is stored outside the row under its SHA-256. The part keeps its shape and `mediaType`; only the payload becomes an `attachment:sha256:<hex>` pointer. Reads put it back verbatim.

The rule is uniform: an image is extracted at 8 KB and at 8 MB, and prose is never extracted at any size. A message's stored shape therefore never depends on how large an image happened to be.

**Row chunking is the size backstop.** A message whose serialized JSON still exceeds `MAX_INLINE_ROW_BYTES` (1.5 MiB) is split across numbered continuation rows and reassembled on read, cut on UTF-8 byte boundaries and never through a surrogate pair. There is no ceiling and no size error.

The two never interact: media leaves before the row is measured, so a message carrying a large image usually has no continuation rows at all.

Payload lifetime is derived from reference rows; bytes go when the last reference does. Content addressing buys idempotency on a retried write, not a space saving — nothing in the design assumes payloads repeat.

### Why this is not the layer an earlier revision removed

That one extracted only when a row was over budget, which made it a rescue mechanism competing with chunking — and it could not do the job it existed for, because extraction needed an extractable leaf. A row over budget from bloated metadata or thousands of small parts had nothing to extract and failed. Chunking won that argument; this store is an ingest rule, not a rescue.

### What it costs

A 200 KB image bills four row writes — the message, one payload chunk, its metadata, one reference — where inlining billed one. A 2 MiB image bills five. Text is unaffected at one row. The bench asserts these exactly.

What that buys is a message row that stays a few hundred bytes however large the payload is, which is what makes pointer-mode reads cheap and the byte budget meaningful.

## Hydration budgets bound memory

`getRecentHistory(maxContentBytes)` charges each row its stored bytes, its continuation rows, **and** the payloads it points at, at inlined size. Extraction makes the row of a message carrying a 2 MB image tiny, so counting rows alone would let a budget admit a window that hydrates far larger than it measured.

The budget is a hard ceiling with no message-count floor beneath it. A floor that admitted rows regardless of size is not a floor under a budget, it is a hole in one — twenty media-heavy rows would be admitted whatever the limit said. The newest message is always returned. `minRecentMessages` is gone.

This does change Think: its window can be shorter than `MODEL_RECENT_WINDOW` when messages are unusually large. With the 32 MiB default it only ever binds on media, since 32 MiB of text is far past what any model could read.

## Storage economics

Every table is `WITHOUT ROWID` with a composite primary key and no secondary index, ordered by a per-session `seq`. A text append with search off bills one row write, down from two. An unchanged update writes nothing: no row, no continuation, no FTS churn, no event. State is derived rather than maintained.

The attachment reference table carries no index on `hash`: it takes a write on every media append, and the reachability check that reads it is a scan of a small table, which is far cheaper than maintaining an index on the write path.

## Context is not conversation storage

Context blocks, frozen prompts, and the search provider move to a new `agents/context` module. Think declares blocks through a new `configureContext()` hook and reaches them through `this.context`; `configureSession()` keeps compaction and search. The `Session` handle stores messages and knows nothing about prompts.

History shaping deliberately stays with the hosts for now. Truncating old tool output slides its boundary a little further every turn, and a sliding boundary rewrites the prompt prefix, which is what invalidates a cache. Moving it here before that is understood would relocate the problem into the module meant to solve it. See #2200.

Intake shaping — the read-path half of #2201 — is stacked on this branch as #2203.

### Think

Keeps its message cache, reconciliation, broadcasts, model assembly, branching, and public arrays. `hydrationByteBudget` defaults to 32 MiB and now bounds memory. Media eviction is unchanged in purpose and simpler in code: it reads inlined history, decodes a `data:` URL directly, and its Workspace write, `[evicted …]` marker and `/attachments/evicted/` paths are untouched. Rewriting a message to a marker now also drops its attachment reference, so eviction genuinely reclaims session storage rather than only shortening context. Think no longer reads Sessions tables with raw SQL.

### AIChatAgent

Keeps its mutable `messages` array, destructive regeneration, retention, broadcasts, and v4 conversion. Boot no longer loads the transcript synchronously in the constructor: the legacy lift and one bounded hydration run in `onStart`, the live array mirrors the Sessions change feed, and `get-messages` streams its response.

## Removed

The synchronous storage aperture (`session.importMessage()` replaces its import use), the R2 tier, the aged-row maintenance pass, the lossy eviction mode, `SessionMessageTooLargeError` and the size ceiling it guarded, `sessionAttachments`, `sanitizeToolPairs`, the token-counter and `CompactContext` plumbing, `onCompactionError`, the `missingUpdate` option, the per-field option thunks, and a duplicate copy of the sanitize helpers.

`SessionStats` loses `totalContentBytes` and `pathLength`. Nothing read either, and both measured only the active branch — excluding other branches, other sessions in the object, and the attachment tables — so as a size signal they answered a different question than the one anyone would ask them. A real signal against the 10 GB ceiling has to sum the tables and deserves its own function.

## Migration

Legacy `assistant_*` message and compaction rows are lifted in SQL. Each source is dropped only after every one of its rows is verified to have a copy holding the same payload; there are no tombstones, because 5 GB of duplicated history in a 10 GB object is its own outage. A lift that cannot verify itself leaves the source in place and does not stamp the schema version, so a later start retries rather than stranding the rows.

AIChatAgent lifts `cf_ai_chat_agent_messages` into a linear chain and drops it only when every row imported — a row that failed to parse is *skipped*, not migrated, and dropping the table on a skip would delete it permanently.

The Sessions schema itself has never been deployed, so it needs no migration of its own.

## Measured

Against 251,895 real messages from local Claude Code and pi transcripts: three exceed the 1.5 MiB row budget, all three in pi, and all three also exceed SQLite's 2 MB ceiling — so without chunking they would have been outright insert failures rather than degraded writes. The largest purely model-generated message anywhere is 215 KB, about 13.6% of the budget. Inbound tool content, not model output, is what gets large.

Prior deployed runs also confirmed the 32 MiB budget genuinely bounding hydration (19 of 536 messages, 31.11 MiB) and Think/AIChat migrations landing byte-identical. Full detail in `design/sessions.md`.

## Validation

- `agents`: 118 files, 1887 tests passing
- `@cloudflare/think`: 44 files, 882 tests passing
- `@cloudflare/ai-chat`: 50 files, 664 tests passing
- 117 typecheck projects, `oxfmt`, `oxlint`, `sherif`, and the package exports check all clean

Follow-ups filed: #2204 (optional R2 tier for the attachment store), #2205 (the change-feed cache mirror is duplicated in every host).


mattzcarey:
Rebased on main (one conflict, `design/rfc-think-multi-session.md` — took main's text, which records the same supersession plus the replacement RFC) and worked through the Devin findings.

**Fixed — AIChat legacy lift could delete history.** The drop guard was `imported + skipped === order.length`, but a *skipped* row is one that could not be parsed or imported. Accounting for a row is not migrating it, so any malformed row was destroyed when the source table went. Now drops only when `skipped === 0 && imported === order.length`. The lift is idempotent, so leaving the table means a later start retries the rows that failed rather than losing them.

**Fixed — Sessions migration never retried.** `onStart` stamped the schema version even when `migrateLegacy` reported an incomplete copy, so the rows it left behind stayed unreachable through the new API forever. `migrateLegacy` now returns completeness and the version is stamped only on success.

**Fixed — the hydration budget was not a bound.** `getRecentHistory` admitted rows whenever fewer than `minRecentMessages` had been taken, whatever they weighed. A floor that ignores size is not a floor under a budget, it is a hole in one. `minRecentMessages` is removed from core, handle, Think and AIChat; the budget is a hard ceiling that always returns at least the newest message.

This does change Think: its window can now be shorter than `MODEL_RECENT_WINDOW` when messages are unusually large, and the two hydration tests that asserted the old floor were updated. The trade is deliberate — with the production 32 MB budget this only ever binds on media, since 32 MB of text is far past what any model could read anyway. New test appends eight ~533 KB media messages against a 1.5 MB budget and asserts the window stays inside it.

**Not a bug.** The `hooks.test.ts` ledger-sweep failure passes in isolation (107/107); it is the known alarm-timing flake.

The remaining findings target code this PR deletes — the old `attachments.ts`, `reapUnreferenced`, offloaded prose, `attachmentBytes` — and Devin has already marked three of them resolved for that reason.

Suites: agents 1860/117, ai-chat 659/50, think 881/44. 117 projects typecheck; oxfmt, oxlint, sherif, check-exports clean.

Intake shaping moved out to #2203, stacked on this.

https://claude.ai/code/session_01MXpoKqJrUFTqgmDrehDti4

Issue #2201, opened 2026-09-02
Shape tool output at the boundary, the way pi does, instead of only after the fact


mattzcarey:
## Summary

Everything we do about context happens *after* a tool has already returned. A tool reads a 3.5 MB file, the whole thing lands in the transcript, and later passes truncate it, evict it, or summarize it away. Pi does the opposite: it caps at the tool boundary, before anything is stored, and hands the model a way to ask for the rest.

We should do the same in `agents/context`, because the tools that produce oversized output are mostly written by users of this SDK, and a guardrail every tool author has to implement themselves is one almost nobody has.

## Where we are against pi

| | pi | Think today |
| --- | --- | --- |
| text read cap | 2,000 lines or 50 KB, whichever hits first | 2,000 lines, 3.5 MB |
| over-cap behavior | truncates and tells the model `offset=N` to continue | truncates the line |
| images | optional `imageProcessor`, `autoResizeImages` on by default | nothing |

Two differences matter more than the numbers.

Pi's cap is about seventy times smaller. `DEFAULT_MAX_BYTES` is 50 KB (`pi-agent-core/harness/utils/truncate.ts`); Think's `MAX_MODEL_FILE_BYTES` is 3.5 MB (`packages/think/src/tools/workspace.ts`).

And pi's truncation carries a **continuation affordance**: `[Showing lines 1-40 of 900 (50KB limit). Use offset=41 to continue.]`. That is what makes an aggressive cap safe — the model is never stuck, it just fetches the next window if it actually needs it. Our truncation is a dead end, which is why our cap has to be generous, which is why large files reach the model.

Think also has no image path. Pi resizes before the bytes enter the request; a full-resolution screenshot from one of our tools enters context at whatever size the tool produced.

## Proposal

Put intake shaping in `agents/context` as a wrapper applied to tool results on the way in:

1. A byte budget for any tool result, defaulting to pi's numbers.
2. Truncation that always emits a continuation hint naming the exact way to get the next window.
3. An image step that downscales before the bytes become part of the request, with a hook so a host can supply its own encoder.

Hosts get it by installing the module rather than by each tool author remembering. A tool that genuinely needs to return something huge opts out explicitly.

## Open questions

- Does shaping run before or after the result is persisted? Before means the transcript never holds what the model never saw, which is smaller and simpler. After means a later policy change can still recover the full output. Storage is lossless either way, so this is a context decision, not a storage one.
[...]

Issue #2200, opened 2026-09-02
Investigate which context-reduction mechanisms break prompt caching


mattzcarey:
## Summary

We have three context-reduction mechanisms and no measurement of what any of them cost in prompt-cache hit rate. Providers cache on a byte-identical prefix, so any technique that rewrites an earlier part of the prompt invalidates the cache from that point on. Some of ours rewrite the prefix once, which is a fine trade. At least one appears to rewrite it on **every turn**, which would be a straight loss: cached input tokens are billed at a fraction of fresh ones, so a technique that trims tokens while destroying the cache can cost more than sending the untrimmed prompt.

Nobody has measured it. This issue is to measure it and then fix or document what we find.

## The three mechanisms

| mechanism | what it rewrites | how often |
| --- | --- | --- |
| `truncateOlderMessages` (`agents/chat`) | old tool outputs to 500 chars, old text to 10,000 | suspected every turn |
| media eviction (`@cloudflare/think`) | a file part becomes an `[evicted …]` marker | once per message |
| compaction overlays (`agents/sessions`) | a span of turns becomes a summary | once per span |

## The specific suspicion

`truncateOlderMessages` keeps the most recent `keepRecent` messages (default 4) intact and truncates everything older. The boundary is computed from the length of the message list, so it **slides forward as the conversation grows**. On the turn after a message crosses that boundary, its full text is replaced by a truncated copy, in the middle of the prompt prefix. Every request after that has a different prefix from the one that was cached.

If that reading is right, a long conversation invalidates its cache on a regular cadence rather than converging on a stable prefix, and the cheapest fix may simply be to anchor the truncation boundary to something stable (a message id, or a compaction anchor) instead of an offset from the end.

Media eviction and compaction should both be one-shot: they rewrite the prefix once and the new prefix is then stable. That is the acceptable shape, and it is worth confirming rather than assuming.

Worth noting that `freezeSystemPrompt` already exists precisely to keep the *system* prompt byte-stable across turns, so the intent is established; the message prefix just never got the same treatment.

## What to measure

Run a long multi-turn conversation against a caching provider and record, per turn, the cache-read and cache-write token counts from the model's usage metadata (`usage.cachedInputTokens` is already surfaced through Think's step context, and pi reports `cacheRead` / `cacheWrite`).

Do it four ways so each mechanism is isolated:

1. everything off
2. `truncateOlderMessages` only
3. media eviction only (`mediaEviction: true`, truncation disabled)
4. compaction only

[...]

PR #2203, opened 2026-09-02
feat(context): bound what a tool result contributes to a request

mattzcarey:
Stacked on #2196. Implements the read-path half of #2201.

## What

`shapeMessage` and `shapeHistory` bound what a stored tool result contributes to a model request:

- a byte cap and a line cap, pi's 50 KB / 2,000 lines by default
- truncation that always carries a continuation hint naming the offset to resume from
- `dropFields`, a set of host-named fields to strip

```ts
shapeMessage(message, { maxBytes: 50_000, dropFields: ["details"] })
```

## Why the read path

Capping is lossy — whatever is cut, the model cannot recover it. So storage keeps the full result and the cap applies to the copy being assembled into a request. A limit can then change next month without having destroyed the bytes in the meantime.

That is the same line drawn everywhere else in this design: sessions is lossless, context shapes. It is also the answer to the open question in `design/context.md` about whether intake shaping runs before or after persist — attachment extraction is lossless so it runs before, in #2196; capping is lossy so it runs after, here.

## Why this is safe for prompt caching

A cache hit needs a byte-identical prefix, which is why sliding history truncation is deferred (#2200) rather than moved here.

These limits are a function of ONE message: how large that tool result is and which fields it carries. The same message shapes to the same bytes on turn 3 and on turn 300, whatever surrounds it, so the prefix is stable. There is a test asserting exactly that.

## `dropFields` comes from a measurement

Across 251,895 real messages from local Claude Code and pi transcripts, three crossed the 1.5 MiB row budget. Pi persists a raw provider payload beside the content it renders, and that duplicate alone accounted for 2.6 MB and 1.65 MB of two of them. Dropping redundant payloads is worth more than the line caps, and costs the model nothing because it never needed both copies.

Which field is redundant is the host's call, so the module names none itself.

## Scope

- Only tool parts are touched. A long assistant answer passes through untouched.
- Images pass through at full resolution, deliberately. Downscaling re-encodes a user's own bytes, and buys less than it appears to since providers scale images down before tokenizing — so it saves upload bytes and latency, not context. A host that wants fewer image bytes in the window evicts the image rather than degrading it.
- Nothing calls this yet. Whether a host wants a cap, and at what size, is a policy decision for the host.

## Tests

9 new tests covering the caps, the continuation hint, line-before-byte ordering, `dropFields`, nested tool output, the pass-through cases, byte-budget cutting that never splits a surrogate pair, streaming, and prefix stability.

https://claude.ai/code/session_01MXpoKqJrUFTqgmDrehDti4

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/agents/pull/2203" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


blev8824-ai:
The lossless-storage/lossy-shaping split is the right design: capping is destructive, so keeping the full result in the session and applying the cap only when assembling the request lets the limit change later without having destroyed bytes. And the continuation hint naming the resume offset is a good rescue for truncated structure.

One gap worth naming: shaping bounds what the model sees, but a tool result can also land in agent state that is not model-visible on this request path, for example values the agent dereferences later or feeds into follow-up tool arguments (spans). If the intent is to bound what the agent can act on, the cap has to apply upstream of anything that becomes a future tool arg, not just to the assembled prompt message. Otherwise a large result can still leak across the boundary through a span, even though the message is capped.

Does shaping also constrain tool results that flow into subsequent tool-call arguments, or only the model-visible message? If it is message-only, a poisoned result can still propagate through a later tool invocation.

Separating the prompt/message path from the enforcement path, so truncation and control are not the same mechanism, is the line AgentKey draws between context and authorization (https://agentkey.us).

PR #2215, opened 2026-09-04
feat(streams): R2 chunk log — segment write-ahead log behind `new Streams({ r2 })`

mattzcarey:
## What

`agents/streams` gets a second chunk log. Hand `Streams` an R2 binding and chunks leave DO SQLite:

```ts
readonly streams = new Streams({
  r2: this.env.BUCKET,
  r2Prefix: `streams/${this.ctx.id}/`,              // default "streams/"
  r2Checkpoint: { everyChunks: 25, everyMs: 1000 }  // the defaults
});
```

Nothing else changes: `open`, `append`, `read`, `status`, `list`, `delete` behave as before and `append()` stays synchronous. Stream rows (state, tag index, metadata, cursor) stay in SQLite, where a point read or tag lookup costs nothing. Chunks go to the bucket.

## Why

Cost. SQLite bills one row per stored chunk and again to delete it. R2 bills one Class A op per checkpoint, deletes are free, and storage is about 13× cheaper with no 10 GB ceiling. One R2 put costs the same as 4.5 SQLite row writes, so the R2 log is cheaper whenever a checkpoint covers more than ~4.5 stored rows: unpacked producers, or a loss window of a few seconds. Against chat's 10-to-1 packed SQLite at a 1 s cadence it is a wash on ops and wins only on storage. The docs say this in the same words.

Per 400-chunk chat turn: $0.0008 SQLite unpacked, $0.00008 packed, $0.00009 R2 at the default cadence, $0.00002 at 5 s.

## How

R2 has no append, rejects bodies of unknown length, and stores nothing from a put that has not completed (all measured on real R2, see below). So the log is a write-ahead log of segment objects:

- Appends go into an in-memory line log that live readers tail, exactly like the SQLite log's in-isolate wakeups. Memory holds only the unflushed tail plus a 256 KB hot window of landed lines; once a segment's put resolves its lines are evicted and a reader further behind reads that segment from R2. A stream's length never grows isolate memory.
- Every `everyChunks` appends or `everyMs` after the first unflushed one, the new lines are put as one immutable segment `<prefix><id>/seg/<epoch>/<from>-<to>`. Each landed segment is the durability: when the isolate dies, everything up to the last landed segment is in R2. Puts are chained so they land in order, retried with backoff, and a range that still fails folds into the next checkpoint so nothing is skipped.
- The row's cursor is stamped from landed segments, throttled to one row write per 5 s, so `status()` never reports more than R2 holds and the stamp costs a fraction of the puts. `open()` after a death re-stamps it exactly.
- `open()` on a stream whose isolate died lists the segments once, keeps the contiguous chain, deletes keys it does not cover, and continues in a new epoch from the chain's end without loading it into memory. A resumed producer starts at the durable cursor, so the Tasks resume contract holds and a discarded generation can never be spliced back in.
- `close()`/`error()` settle the row synchronously, then in the background stream the segments back through one exact-size put at `<id>/body` and drop them (keys are tracked, no list; deletes batched at 1000). Segments stay readable until the body lands. `await streams.flush(id)` waits for it.
- Replay of a settled stream caches bodies up to 1 MiB whole; larger bodies get a line-offset index and ranged gets per page. The read cache is an 8 MiB LRU.

SQLite accounting on the R2 backend: 100 appends read and write 0 rows; the cursor stamp is 1 row write per 5 s of streaming; `status()` reads 1, tag lookup 2, settle reads 2 and writes 2.

Chat's `ResumableStream` uses the synchronous SQLite aperture and is unaffected; it throws if given an R2-backed `Streams`.

## Measured on real R2 (2026-09-03)

| Experiment | Result |
|---|---|
| `put(key, unknownLengthStream)` | rejected: "Provided readable stream must have a known length" |
| `FixedLengthStream` put, slow writes | streams as written; a 150 s put from a DO completed |
| object visible mid-put? | no; not in `list`/`head` until complete |
| producer aborts or closes short | put rejects, nothing stored |
| multipart with tiny parts | `complete` rejects (10011, below 5 MiB minimum) |
| small put latency | p50 145 ms, p95 195 ms |
| tag lookup, 1000 streams | R2 prefix scan 542 ms, key-encoded index 41 ms, SQLite 0 ms |

Those are why there is no streaming file put (it never contributes durability) and why the rows stay in SQLite.

## Tests

- `src/tests/streams/r2-backend.test.ts` (miniflare R2 binding `STREAMS_R2`): cursor + live tail + body at close; no chunk rows in SQLite; `error()` path; resume from the segment chain after a simulated restart; replay before recovery never pins a truncated log; delete; SQLite accounting; aperture refusal.
- Existing streams, write-accounting, bench, migration and chat suites unchanged and green (46 tests).
- `r2-backend.test.ts` also covers memory: 2000 × 500 B chunks keep under half the lines in memory while a reader far behind is served from segments, and the 1 MB settled body replays from any cursor through the offset index.
- Real R2 with the package build: 200 chunks tailed over `sseResponse` while streaming, body landed, segments cleaned; `ctx.abort()` at chunk 120 left five segments and a durable cursor of 100, resume continued from 100, final body 200 lines. A 4000 × 500 B stream (2.2 MB) settled through the streamed compaction with no segments left, and the same stream killed at chunk 925 resumed from 925 and settled to 4000 lines.

## Not in this PR

- A SIGKILL e2e for the R2 backend (the workers-pool restart simulation covers the same path).
- Per-stream cadence overrides; `r2Checkpoint` is capability-wide.

### DIFF
--- a/.changeset/streams-r2-chunk-log.md
+++ b/.changeset/streams-r2-chunk-log.md
@@ -1,0 +1,7 @@
+---
+"agents": minor
+---
+
+feat(streams): store chunk logs in R2 with `new Streams({ r2: env.BUCKET })`.
+
+Stream rows (state, tag index, metadata, cursor) stay in DO SQLite; chunks go to R2 as a write-ahead log of segment objects, checkpointed every 25 chunks or 1 s (`r2Checkpoint`), so a Durable Object that dies mid-stream leaves everything up to its last checkpoint in R2 and a restarted producer resumes from it. Settlement compacts the segments into one exact-size object; `Streams.flush()` awaits it. The synchronous storage aperture used by chat stays SQLite-only.
--- a/docs/agents/streams.md
+++ b/docs/agents/streams.md
@@ -183,28 +183,38 @@ rejects bodies of unknown length, and stores nothing from a put that has
 not completed. So the log is a write-ahead log of segment objects:
 
 - Appends go into an in-memory line log that live readers tail, exactly
-  like the SQLite log's wakeups.
+  like the SQLite log's wakeups. Memory holds only the unflushed tail plus
+  a 256 KB hot window of landed lines: once a segment's put resolves its
+  lines are evicted, and a reader further behind reads that segment from
+  R2. A stream's length never grows the isolate's memory.
 - Every `everyChunks` appends or `everyMs` after the first unflushed one,
   the new lines are put as one immutable segment under `<id>/seg/`. Each
   landed segment is the durability. When the isolate dies, everything up
   to the last landed segment is in R2, and the loss window is the cadence.
-- The row's cursor is stamped only after a segment's put resolves, so
-  `status()` never reports more than R2 holds.
-- `open()` on a stream whose isolate died rebuilds the log from the
-  contiguous segment chain, deletes keys the chain does not cover, and
-  continues in a new epoch. A resumed producer starts at the durable
-  cursor, so the Tasks resume contract holds and no discarded generation
-  can be spliced back in.
+- The row's cursor is stamped from landed segments, throttled to one row
+  write per 5 s, so `status()` never reports more than R2 holds and the
+  stamp costs a fraction of the puts. While a stream is live the row can
+  lag the chain by a few seconds; `open()` after a death re-stamps it
+  exactly. A segment put that fails after retries folds its range into
+  the next checkpoint, so nothing is skipped.
+- `open()` on a stream whose isolate died lists the segments once, keeps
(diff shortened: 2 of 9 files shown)


mattzcarey:
Superseded by #2216: the R2 segment log was measured against the SQLite alternatives and lost on cost (a Class A put is 4.5 SQLite row writes; the row-write cost was the cleanup, not the append). #2216 keeps stream blocks in the same DO storage as the session message and makes the handoff one transaction. The R2 measurements and the spike stay on this branch for reference.

PR #2232, opened 2026-09-10
perf(think): hydrate attachment bytes only for the recent tail

mattzcarey:
## Summary

Think hydrated every attachment in its cache window as base64 and kept it resident for the object's lifetime: `getRecentHistory(budget)` re-inlined every pointer, `_cachedMessages` held the result, every broadcast serialized it, and every turn walked it — while `truncateOlderMessages` only replays the newest 4 messages at full fidelity and older media is not needed by the model.

This PR hydrates media only for the recent tail.

### Sessions (`agents/sessions`)

- `getRecentHistory(budget, { inlineAttachments })` — policy-only option, default `"all"` (unchanged behaviour). `{ newest: N }` puts bytes back only for the newest N rows of the window; older rows keep the stored pointer form (`attachment:sha256:<hash>` in the field that held the payload) and their attachment rows are never read.
- **Pointer round-trip fix** (latent bug, independent of the option): writers derived attachment references only from media extracted on that write, so re-writing a message that already carried a pointer with any other change (`updateMessage`) dropped the reference and collected the bytes under a live pointer; appending a pointer-form copy under a new id took no reference at all. `extractAttachments` now also reports the pointers a message already carries (`references`), and `append`/`update`/`import` record those. Tests cover both cases.
- Bench (`storage-ops-bench.test.ts`): 10 image messages, tail of 2 → 16 fewer rows read than full inlining, resident bytes ~5x smaller. Fails on `main`.

### Think

- `mediaHydration: MediaHydrationConfig | "all"`, default `{ keepRecentMessages: 4 }` (clamped to the model's recent window). The cache holds pointers for older messages; before a turn a part still carrying a pointer reaches the model as `[image/png omitted: older than the recent window]` (`media-hydration.ts`). Nothing durable changes: rows keep their pointers, `getHistory()` still returns bytes.
- The media-eviction scheduler treats a tail-hydrated cache like a windowed one (aged rows' bytes are not in memory to inspect), so eviction still runs once per distinct stored size.
- Test agent: 8 × 120 KB images — resident cache 1.28 MB → ~0.64 MB (only the newest 4 carry bytes); model messages contain no pointer; a pointer-form cached message written back with a metadata stamp still resolves to its bytes on a full read.

## Behaviour changes a reviewer must sign off

1. **Model context**: with default config, images older than the newest 4 messages are no longer sent to the model at full size — they become a marker. Previously `truncateOlderMessages` passed file parts through untouched at any age (only media eviction removed them, and only when a Workspace is configured). This is the intended saving, but it is a visible change for hosts without a Workspace. `mediaHydration: "all"` restores the old behaviour.
2. **Wire form to clients**: messages older than the window are broadcast with `attachment:sha256:…` URLs. There is no client-side attachment fetch route, so a client cannot render those images from the transcript alone. With media eviction on (default) plus a Workspace, aged images were already replaced by markers after a pass; without a Workspace this is new. Options if this is unacceptable by default: flip the default to `"all"` (Think opt-in), or add a Think route that serves an attachment by hash (out of scope here).
3. **Reconcile CPU**: a client that still holds the inline form of an old message (received before a DO restart) re-sends it each turn; the cached pointer form no longer JSON-equals it, so Sessions re-hashes the payload to conclude "unchanged" (no write, cache not re-inflated). Clients that received pointer form send it back unchanged and short-circuit as before.

## Tests / gates

- `packages/agents`: `pnpm vitest --run --project workers src/tests/sessions` — 77 passed (2 new attachment tests + 1 new bench).
- `packages/think`: `pnpm vitest --run -c src/tests/vitest.config.ts` — 898 passed (4 new).
- `packages/ai-chat`: 653 passed.
- `oxfmt --check .`, `oxlint`, `tsc --noEmit` in agents and think, `check:exports`: clean.

## Docs

`docs/agents/sessions.md` (inlineAttachments, pointer write-back), `docs/think/index.md` (Media hydration section + config table row).


mattzcarey:
Closing: the default behaviour change (placeholders for older images in model context, pointer URLs on the wire) is not wanted. The attachment-ref fix found along the way will be extracted into its own PR.

PR #2234, opened 2026-09-10
perf(think): memoise per-turn tool sets; skip skills refresh on continuations

mattzcarey:
## Problem

`_prepareInferenceInvocation` rebuilt every tool set on every inference attempt — each new turn, each auto-continuation (client tool result, connection-less continuation, `runTurn({ continuation: true })`), and each reactive-overflow retry:

`createWorkspaceTools`, `createFetchTools`, `getTools()`, `_compileActionTools()` (re-iterating `getActions()` and re-deriving metadata), `extensionManager.getTools()`, `_refreshSkillsIfChanged()` (which always called `SkillRegistry.refresh()` → re-`list()` every source, then re-projected the workspace, and only *then* compared fingerprints), `context.tools()`, `_skillRegistry.tools()`, MCP `getAITools()`, client tools.

(Correction to the original premise: proactive compaction does not re-enter this path — it reuses `_activeTurnTools`. The re-entries are continuations and the overflow retry loop.)

## What changed

| Set | Before | After |
|---|---|---|
| workspace + fetch | rebuilt per attempt | built once; keyed on `workspace` / `workspaceBash` / `fetchTools` identities |
| actions | `getActions()` + compile every attempt | `getActions()` still called every attempt (user hook, may be dynamic); compile memoised per `Action` descriptor (same tool name + kind). `_activeTurnActionMetadata` derived from the memo; approval-descriptor map still reset per attempt |
| extension tools | rebuilt per call | `ExtensionManager.getTools()` memoised; dropped on load / restore / unload (execute closures already re-check membership) |
| context-block tools | rebuilt per call | `ContextBlocks.tools()` memoised on block shape (labels + writable + searchable). Content edits / `refreshSystemPrompt()` keep the set |
| skill tools | rebuilt per call (new `z.enum`) | `SkillRegistry.tools()` memoised on catalog names + script-runner presence |
| skills refresh | every attempt | at most once per `intervalMs` (default 60s); never on a continuation or an overflow retry of the same request |
| `getTools()`, MCP, client tools | per attempt | unchanged (user hook / documented per-call / per-request executor) |

New option on `Think`:

```ts
skillsRefresh: "every-turn" | "on-start" | { intervalMs: number } = { intervalMs: 60_000 };
```

**Default rationale.** `skills.r2` already indexes with a 60s TTL (`refreshIntervalMs`), so 60s is the shortest interval at which a refresh can observe anything new from the one remote source we ship; a longer default would only add staleness on top of the R2 TTL without saving a network round-trip. `fromManifest` is build-time constant, so any interval is free there. `"every-turn"` remains for hosts with a custom source that has no TTL of its own. Documented in `docs/think/index.md`; changeset `.changeset/quiet-skills-refresh.md` (think minor, agents patch).

## What a catalog change touches (audit)

When a refresh finds a new fingerprint, Think does exactly three things:

1. `context.refreshSystemPrompt()` — re-renders every block and overwrites the **persisted frozen prompt**. This is the only "rewrite", and it is the intended one: it feeds the next turn's system message.
2. `_configSet("skillsFingerprint", …)` — one `think_config` row.
3. `_configureSkillWorkspace()` → `seedWorkspace()` — writes `SKILL.md` for skills whose file is absent; **existing files are preserved** (`onConflict` default).

Nothing walks or rewrites transcript rows. The `activate_skill` tool result from an earlier turn keeps the old body, and the system prompt an earlier turn was sent is not stored per message, so it cannot be rewritten.

One thing worth knowing, pre-existing and not changed here: with `skillWorkspace` enabled, a source change updates the catalog prompt but **not** an already-seeded `SKILL.md` (preserved by design so user edits win), so `activate_skill` can serve the old body under the new description until the file is removed or `onConflict: "replace"` is set.

## Numbers (from `tool-memo.test.ts`)

Turn → continuation → turn with `"every-turn"`:

| | before | after |
|---|---|---|
| skill source `list()` calls (incl. `onStart` load) | 4 | 3 |
| skill source `refresh()` calls | 3 | 2 |
| tool objects reused across attempts (workspace / action / skill / context) | no | yes |
| `getTools()` / `getActions()` calls | 3 / 3 | 3 / 3 (unchanged by design) |

Default `{ intervalMs: 60_000 }`: two back-to-back turns → 1 refresh (was 2). `"on-start"`: 0 refreshes.

Skill mutated between turns (`ThinkSkillChangeTestAgent`): rows persisted before the change are byte-identical afterwards; the v1 body stays in the transcript and v2 never appears in it; model calls saw prompts `[v1, v1, v2, v2]` — the prefix changed exactly once, on the turn after the refresh, and was stable after that.

## Tests

- `packages/think/src/tests/tool-memo.test.ts` (+ `agents/tool-memo.ts`): counters across turn → continuation → turn; identity assertions on built tools; the three policies and the default; the skill-change / persisted-rows test. All fail on `main`'s `think.ts`.
- `extension-manager.test.ts`: `getTools()` identity across load/unload.
- `agents/src/tests/context/context.test.ts`: `tools()` memo on shape.
- `agents/src/tests/skills.test.ts`: `tools()` identity across refresh with same/changed catalog.
- Full Think suite green; `oxfmt --check .`, `oxlint`, `tsc` clean.

## Open question

Should `getTools()` get an opt-in static mode? Left alone because the `fetchTools` docs explicitly promise `getTools()` runs every turn.


mattzcarey:
Closing after review: the skills refresh this skips is already TTL-guarded and fingerprint-guarded, so the saving is speculative; the action memo never hits for inline action() construction; and public methods would return a shared mutable ToolSet under a patch bump.

PR #2238, opened 2026-09-10
fix: allow interfaces as agent props

### LINKED ISSUE #1886: Allow proper types for agent props

mattzcarey:
## Summary

- Allow interfaces with named fields to be used as `Props` for `Agent`, `AIChatAgent`, legacy `McpAgent`, and `Lifecycle`.
- Apply the same `object` bound to `getAgentByName`, `AgentGetOptions`, and request-routing options.
- Keep `Record<string, unknown>` as the default, so untyped callers still read arbitrary props as `unknown`.
- Add compile-time regressions for the exact `getAgentByName(namespace, name, { props: config })` report, typed Agent namespaces, routing, Lifecycle, MCP, and AI Chat. Primitive props remain rejected.

## Why this is still needed

PR #1906 was closed in favor of changing the bound in PartyServer. Since then, `Agent` stopped extending PartyServer and now composes the SDK-owned `Lifecycle`, where the same `Record<string, unknown>` constraint remained. The upstream PartyServer change no longer fixes this API by itself.

## Proof

Before the fix, the new regression produced the reported error on current `main`:

```text
TS2322: Type 'RolePlayAgentConfig' is not assignable to type
'Record<string, unknown>'.
Index signature for type 'string' is missing in type 'RolePlayAgentConfig'.
```

After the fix:

- A standalone strict TypeScript consumer compiled successfully against the built `agents` and `@cloudflare/ai-chat` package entry points, using an interface as props without casts or an index signature.
- `pnpm run check` passed, including all 119 TypeScript projects.
- `pnpm exec nx affected -t test --base=origin/main` passed for 22 projects. This included 3,195 Agents tests, 735 AI Chat tests, and 898 Think tests.

This is a type-only change with no runtime behavior change.

Fixes #1886

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/agents/pull/2238" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


packages/agents/src/agent-routing.ts
```diff
@@ -385,15 +403,18 @@ export async function routeAgentRequest<Env>(
...
  * @param namespace - Agent Durable Object namespace.
  * @param name - Agent instance name.
  * @param options - Placement, startup properties, and retry options.
  * @returns The initialized Agent stub.
  */
 export async function getAgentByName<
   Env extends Cloudflare.Env = Cloudflare.Env,
-  T extends Agent<Env> = Agent<Env>,
-  Props extends Record<string, unknown> = Record<string, unknown>
+  T extends Agent<Env, unknown, object> = Agent<Env>,
```
stof, line 416: this should define a relation between the `Agents` props and the `Props` template type of `getAgentByName`. Otherwise, typescript will not enforce passing compatible Props, loosing the benefits of type-safety.

packages/ai-chat/src/tests-d/agent-props.test-d.ts
```diff
@@ -0,0 +1,28 @@
...
+
+declare abstract class RolePlayChatAgent extends AIChatAgent<
+  typeof env,
+  unknown,
+  RolePlayAgentConfig
+> {}
+
+declare const config: RolePlayAgentConfig;
+declare const namespace: DurableObjectNamespace<RolePlayChatAgent>;
+getAgentByName(namespace, "role-play", { props: config });
```
stof, line 19: a separate test should be added, which should call `getAgentByName` with `props` being an object missing some properties of the `RolePlayAgentConfig` interface. This test should expect a TS error on such call to prove it detects such type mismatch.

packages/agents/src/tests-d/agent-props.test-d.ts
```diff
@@ -0,0 +1,71 @@
...
+  RolePlayAgentState,
+  RolePlayAgentConfig
+> {
+  override onStart(props?: RolePlayAgentConfig): void {
+    props?.role satisfies string | undefined;
+  }
+}
+
+declare const rolePlayNamespace: DurableObjectNamespace<RolePlayAgent>;
+getAgentByName(rolePlayNamespace, "role-play", { props: config });
```
stof, line 38: a separate test should be added, which should call `getAgentByName` with `props` being an object missing some properties of the `RolePlayAgentConfig` interface. This test should expect a TS error on such call to prove it detects such type mismatch.

PR #2253, opened 2026-09-11
feat(agents): move state into an opt-in Lifecycle capability (agents/state)

mattzcarey:
Supersedes #2179, whose head lived on a fork; this PR's head is `cloudflare/agents:feat/state-capability` so CI runs on origin. Includes the merge of current main and the Devin fixes.

State was one method (`_setStateInternal`) doing four jobs inside the Agent god-class — validate, persist, broadcast, notify — with the state row, in-memory cache, and schema all threaded through the class. This moves storage and change ordering into a `State` capability (`agents/state`), so any Lifecycle host gets durable, validated state without inheriting `Agent`. Same pattern as the WebSockets (#2169) and MCP client (#1895) capabilities.

Agent's public API and wire protocol are unchanged.

### API

```ts
import { State } from "agents/state";

class Counter extends DurableObject {
  readonly state = new State<{ count: number }>({
    initialState: { count: 0 },
    validateStateChange: (next, source) => { /* throw to reject */ },
    onChanged: (state, source) => { /* sync or async */ }
  });
  readonly lifecycle = Lifecycle.install(this).use(this.state);
}

counter.state.get();            // { count: number } | undefined
counter.state.set({ count: 1 }); // "server" source
counter.state.set(next, connection); // client-originated
```

`StateOptions` is plain data: `initialState`, `validateStateChange`, `onChanged`. `StateChangeSource` is `Connection | "server"`. Async `onChanged` is tracked and its rejection logged; `set()` stays synchronous. The capability reaches Lifecycle only for `storage` and uses `#` private fields like its siblings.

### Architecture

```
+--------------------------------------+
| Agent                                |
|  get state -> _state.get()           |
|    (seeds initialState if undefined) |
|  setState  -> _state.set()           |
|  _handleStateChanged():              |
|    3. broadcast --------> clients    |
|    4. notify hook                    |
|         ^ onChanged                  |
+---------|----------------------------+
+---------+----------------------------+
| State (capability)                   |
|  onStart: own schema version key,    |
|           create table, legacy clean |
|  get(): load / cache                 |
|  set(): 1.validate 2.persist         |
|         3. onChanged(state, source)  |
+---------+----------------------------+
          | this.lifecycle.sql / storage
          v
   cf_agents_state (SQLite)
```

The capability never references connections, the Agent, env, or ctx. Agent passes two closures (`validateStateChange`, `onChanged`) as static options and does the WebSockets-specific work itself. Agent's generic parameter is renamed `State` → `TState` (positional for consumers, so not a breaking change) so the class can be imported as plain `State`.

### Data flows

Server sets state (`source = "server"`, broadcast to all):

```
setState(next) -> _state.set(next, "server")
  1. validate  (Agent.validateStateChange)
  2. persist   -> cf_agents_state
  3. onChanged -> Agent._handleStateChanged(next, "server")
       broadcast CF_AGENT_STATE to all protocol clients
       waitUntil -> onStateChanged / onStateUpdate dev hook
```

Client sends state over WS (`source = connection`, echo excludes sender):

```
onMessage (stays in Agent): parse, readonly check, CF_AGENT_STATE_ERROR
  -> _state.set(next, connection)
       1. validate  2. persist
       3. onChanged -> _handleStateChanged(next, connection)
            broadcast to all EXCEPT connection.id
            dev hook
```

### What moved vs. what stays

| Concern | Owner |
|---|---|
| `cf_agents_state` table, state row, load/cache, validate + persist | State |
| legacy `cf_state_was_changed` cleanup | State (its v1 migration) |
| own schema version key (`cf_agents:state_schema_version`) | State |
| `initialState` field and first-access seeding | Agent (`state` getter) |
| broadcast to connections | Agent (`_handleStateChanged`) |
| dev hooks (`validateStateChange`, `onStateChanged`/`onStateUpdate`) | Agent (override surface) |
| `onMessage` parse / readonly / error responses | Agent / WebSockets |

`initialState` stays an Agent field because subclass fields initialize after the base constructor. Rather than the capability reaching back into the host lazily, Agent's `state` getter seeds it on first read when the capability returns `undefined` (not JSON-representable, so it uniquely means "no row"). The seed goes through `set()`, so it persists, broadcasts, and runs the hook as before.

### Table ownership

`State` is the only owner of `cf_agents_state`. Agent used to keep its global schema version as a row in that table, which was the only reason it still created the table. That version now lives under the `cf_agents:schema_version` KV key, read synchronously via `storage.kv` so the constructor can still gate DDL — the same convention every capability uses. A DO created under the old layout has the row read once, moved to the key, and deleted on its next construction.

### Compatibility

`state`, `setState()`, `onStateChanged`, and the `CF_AGENT_STATE` frames behave identically. The full Agent state suite and the schema suite pass; the fixture reach-ins to the removed `_state` cache are replaced by eviction (`evictDurableObject`) or by driving the capability's own migration.

### Tests

Capability suite installs `State` on a bare Durable Object through `withCapabilityHarness`: persist/read, access before lifecycle start, initial-state seeding, falsy-value row existence, rehydration across eviction, validation rejection, `onChanged` source on both origins, sync/async hook failures. Schema suite adds a legacy schema-version-row migration case.

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/agents/pull/2253" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


mattzcarey:
Closing in favour of the original #2179, per Matt. The origin branch feat/state-capability carries a merge of current main (one trivial conflict in Agent's lifecycle.use chain) if it is useful; the waitUntil change was reverted.

PR #2275, opened 2026-09-15
Add durable Web Channel for browser chat

cjol:
This PR adds `agents/channels/web`, a Web Channel that serves existing `useAgentChat` clients from arbitrary Lifecycle Durable Objects, and adds Host-owned durable response recording and canonical history resolution. It supersedes the closed draft #2188.

## Why

- Channels could stream text to provider adapters, but they could not preserve the rich message, reasoning, tool, source, file, metadata, and data events required by browser chat clients.
- Live WebSocket delivery alone cannot recover output missed during a disconnect or isolate replacement, and a finalized transcript cannot represent an active or interrupted response attempt.
- We could persist encoded Web frames, but that would make the durable source unusable by Voice and future Channels. This records normalized `ChannelChunk`s once at the Host boundary and leaves wire projection with each adapter.
- We could let Web own approval, cancellation, reset, and history policy, but those actions affect application state. They instead cross transport-neutral `ChannelHost` interfaces while Web retains only connection and resume protocol mechanics.

## Public API Surface

| Symbol | Kind | Notes |
|---|---|---|
| `agents/channels/web` | New entry point | Exports `web`, `WebChannel`, `WebChannelOptions`, `WebConnectionIdentity`, `WebChatIngressPayload`, and `WebChatSurface`. |
| `ChannelChunk` | Expanded union | Adds message and part boundaries, structured tool input/output and approval events, files, metadata, application data, and stable part identities. Existing text, reasoning, tool-status, and source chunks remain valid. |
| `ChannelFinishReason`, `ChannelJsonValue`, `ChannelJsonObject`, `ChannelProviderMetadata` | New types | Describe the dependency-neutral rich chunk vocabulary. |
| `ChannelConversationChunk`, `ChannelConversationMessage`, `ChannelConversationSnapshot`, `ChannelMessageResolver` | New types | Represent canonical application-owned conversation history, including participant-scoped content. |
| `ChannelResponseContext` | New type | Supplies stable response, conversation, and canonical assistant-message identities for durable recording. |
| `ChannelPartRenderingOptions` | New type | Configures optional tool-status and reasoning projections for Slack and Telegram. |
| `ChannelClientTool`, `ChannelToolResult`, `ChannelCancelRequest`, `ChannelConversationResetRequest` and input variants | New event types | Normalize browser tool results, cancellation, and reset without exposing Web protocol identifiers to applications. |
| `ChannelHostOptions.streams`, `resolveMessages`, `onToolResult`, `onCancel`, `onConversationReset` | Additive options | Configure durable output, canonical history, and application-owned ingress policy. |
| `ChannelEventContext.operationId` | Additive field | Correlates events from one application operation without exposing transport internals. |
| `ChannelStreamOptions.response` | Additive option | Required when the Host is configured with Streams. |
| `toChannelChunks`, `fromUIMessageStream`, `toUIMessageStream` | New functions | Convert AI SDK full and UI streams through the shared Channel chunk contract. |
| `ConsumeChunksOptions` | New type | Adds abort-signal support to chunk consumers. |
| `SlackChannelOptions.renderParts`, `TelegramChannelOptions.renderParts` | Additive options | Opt into provider-appropriate tool and reasoning rendering. |

## Architectural Changes

```text
useAgentChat
    ↕ existing browser protocol
Web Channel
    ↕ neutral ingress and ChannelChunk output
ChannelHost ─── canonical history resolver
    │
    └── Streams: one durable neutral log per response attempt
          ↑
          └── Web replay reads prefix, then follows the live tail
```

- Canonical snapshots answer what the conversation contains now. Response streams answer what happened during one response attempt. Neither replaces the other.
- `ChannelHost.stream()` assigns stable message and part boundaries, appends each neutral chunk before delivery, and records once above fallback or fanout policy.
- Web stores request correlation, participant ownership, continuation state, and advertised browser tools as stream metadata rather than embedding transport details in chunks.
- Reconnecting clients keep the existing offer and ACK protocol, preserve the original Web request ID, replay from the durable cursor, and follow the same live stream without a second catch-up layer.

## Code Changes

- Add rich `ChannelChunk` conversion for AI SDK full streams and UI-message streams, while retaining provider-specific text projections.
- Add Host service binding so configured Channels can share canonical history and read durable response streams without taking ownership of application persistence.
- Add the Web adapter with participant-aware snapshots, multi-connection fanout, client-tool continuation, approvals, cancellation, reset, durable replay, and replay-to-live handoff.
- Serve `/get-messages` through Web Channel ingress using the same identity resolution and participant filtering as socket hydration.
- Keep incoming `cf_agent_chat_messages` ignored because it is a lossy client projection, not authoritative application state.
- Extend the local Web fixture with a React `useAgentChat` client, real socket scenarios, and process-restart coverage.

## Compatibility

- Existing Channel text streams remain valid. Rich chunk variants are additive.
- A Host configured with `streams` must pass `options.response` to every `stream()` call so the durable record has stable identities.
- Web applications should persist canonical messages through `resolveMessages` and set `syncMessagesToServer: false`; client transcript replacement is intentionally not accepted as authoritative state.
- The adapter covers browser chat requests, streaming, tools, approvals, cancellation, reset, snapshots, and reconnect replay. Generic `useAgent` state and RPC remain separate WebSockets capabilities.
- AIChatAgent turn recovery and programmatic message-update broadcasts remain application behavior. This PR extracts the browser transport boundary; it does not migrate the AIChatAgent turn engine.


packages/agents/src/channels/channel.ts
```diff
@@ -11,6 +20,32 @@ import type {
 
 export type Awaitable<T> = T | Promise<T>;
 
+export type ChannelChunkAudience =
+  | { type: "conversation" }
+  | { type: "participant"; participantId: string };
```
cjol: I need. a strong justiviation for why this exists; I don't like that different participants have different views onto the transcript. I think it's something to do with client tools? If the justification is not strong enough I'll ask to remove it.

packages/agents/src/channels/channel.ts
```diff
@@ -11,6 +20,32 @@ import type {
...
+
+export type ChannelConversationChunk = ChannelChunk & {
+  /** Omit for conversation-visible content. */
+  audience?: ChannelChunkAudience;
+};
+
+export type ChannelConversationMessage = {
```
cjol: Generally ths "ChannelConversationMessage"/ChannelConversationChunk types, I'm not sure why it's needed compared to just updating ChannelChunk directly. I'm worried we're just adding indirection wrappers and introducing different types for users to get confused about later.

packages/agents/src/channels/channel.ts
```diff
@@ -83,6 +96,23 @@ export type ChannelDeliveryOptions = {
   delivery?: ChannelDeliveryContext;
 };
 
+/** Optional projections of non-text stream parts into a provider message. */
+export type ChannelPartRenderingOptions = {
```
cjol: I don't think this type shold be central; it's up to each channel to decide an appropriate set of options

packages/agents/src/channels/chunks.ts
```diff
@@ -0,0 +1,150 @@
...
+  | { type: "step-start" }
+  | { type: "step-finish" }
+  | ({ type: "text-start"; id: string } & ProviderContext)
+  | ({ type: "text-end"; id: string } & ProviderContext)
+  | ({ type: "text"; text: string; id?: string } & ProviderContext)
+  | ({ type: "reasoning-start"; id: string } & ProviderContext)
+  | ({ type: "reasoning-end"; id: string } & ProviderContext)
+  | ({ type: "reasoning"; text: string; id?: string } & ProviderContext)
+  | {
+      type: "tool";
```
cjol: Perhaps `type: "tool-status"` would be clearer?

packages/agents/src/channels/chunks.ts
```diff
@@ -0,0 +1,150 @@
...
+      mediaType: string;
+    } & ProviderContext)
+  | {
+      type: "data";
+      name: string;
+      id?: string;
+      data: unknown;
+      transient?: boolean;
+    }
+  | ({ type: "custom"; kind: `${string}.${string}` } & ProviderContext);
```
cjol: this escape hatch makes me nervous and invites future agents to abuse it. Let's remove if we don't have a strong justification

packages/agents/src/channels/ai-sdk-stream.ts
```diff
@@ -0,0 +1,372 @@
...
+          id: chunk.id,
+          data: chunk.data,
+          transient: chunk.transient
+        });
+      }
+      throw new Error(`Unknown AI SDK UI event: ${chunkType(chunk)}`);
+  }
+}
+
+function chunkType(chunk: never): string {
```
cjol: inline this

packages/agents/src/channels/ai-sdk-stream.ts
```diff
@@ -0,0 +1,372 @@
+import type { UIMessageChunk } from "ai";
+import type { ChannelChunk } from "./channel";
+
+function abortError(reason?: string): Error {
+  return new Error(reason ?? "The generation was aborted");
+}
+
+function copyDefined<T extends object>(value: T): T {
```
cjol: I'm not sure this is a useful indirection, we can tolerate absent fields as easily as we can undefineds.

packages/agents/src/channels/ai-sdk-stream.ts
```diff
@@ -0,0 +1,372 @@
...
+          data: chunk.data,
+          transient: chunk.transient
+        })
+      ];
+    default:
+      return assertNever(chunk);
+  }
+}
+
+function assertNever(value: never): never {
```
cjol: inline this

packages/agents/src/channels/ai-sdk.ts
```diff
@@ -57,105 +59,328 @@ export function createSendMessageTool(
   });
 }
 
+function defined<T extends object>(value: T): T {
```
cjol: This function keeps cropping up -- please scan the whole PR forit and ensure it's definitely needed ot delete it everywhere if not.

packages/agents/src/channels/ai-sdk-stream.ts
```diff
@@ -0,0 +1,372 @@
...
+  ) as T;
+}
+
+function isDataChunk(
+  chunk: UIMessageChunk
+): chunk is Extract<UIMessageChunk, { type: `data-${string}` }> {
+  return chunk.type.startsWith("data-") && "data" in chunk;
+}
+
+export function uiChunkToChannelChunk(
```
cjol: Is this really necessary? I'd like to avoid two converters fromthe AI SDK domain. We already can convert text stream parts, so can we compose with AI-sdk native utilities? e.g. maybe keep only our UIMessageChunk converter, and map AI SDK parts to UIMessageChunks first before calling our converter

packages/agents/src/channels/channel.ts
```diff
@@ -93,6 +123,8 @@ export type ChannelStreamOptions = {
   title?: string;
   /** Caller-owned correlation an Adapter may use where the provider supports it. */
   delivery?: ChannelDeliveryContext;
+  /** Required when the Host has durable Streams configured. */
+  response?: ChannelResponseContext;
```
cjol: I think we can simplify significantly if we just make this mandatory. Say every streamed response has a stable identity and normalized message/part boundaries.

packages/agents/src/channels/host/index.ts
```diff
@@ -189,8 +283,52 @@ export class ChannelHost {
...
+    surface: ChannelMessageSurface,
+    chunks: ChannelChunkSource,
+    options: ChannelStreamOptions
+  ): Promise<ReadableStream<ChannelChunk>> {
+    const response = options.response;
+    if (!response) {
+      await chunks.cancel().catch(() => {});
+      throw new Error(
+        "ChannelHost.stream requires options.response when Streams are configured"
+      );
```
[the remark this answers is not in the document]
cjol: let's resolve this by making response identity mandatory everywhere

packages/agents/src/channels/host/index.ts
```diff
@@ -332,6 +523,157 @@ export class ChannelHost {
...
+  let implicitPart: { type: "text" | "reasoning"; id: string } | undefined;
+
+  function release(): void {
+    if (released) return;
+    released = true;
+    reader.releaseLock();
+  }
+
+  function append(chunk: ChannelChunk): void {
+    writer.append(chunk as unknown as import("../../streams").StreamJson);
```
cjol: ugly type cast here, can we do better?

packages/agents/src/channels/chunks.ts
```diff
@@ -0,0 +1,150 @@
...
+  | "content-filter"
+  | "tool-calls"
+  | "error"
+  | "other";
+
+/**
+ * One event in a progressively generated answer.
+ *
+ * Simple producers may emit only `text`. Rich producers preserve message/part
+ * boundaries, tool calls and results, and structured content in this same stream.
```
cjol: I don't want a distinction between "simple" producers and"rich" producers. There must be a single producer contract, which will need to be the richer one. If we later want to add abstractions to simplify further, we can layer those on on top. The current separation adds complexity throughout, e.g. normalization between the formats in multiple places. If we need to make our producers more complex, so be it.

packages/agents/src/channels/ingress.ts
```diff
@@ -29,9 +29,19 @@ export type ChannelActorInput = Omit<ChannelActor, "identity"> & {
...
+  readonly description?: string;
+  /** JSON Schema, including boolean schemas. */
+  readonly inputSchema?: Record<string, unknown> | boolean;
+};
+
 export type ChannelEventContext = {
   /** Immutable Channel-scoped identity, independent of application routing. */
   readonly eventId: string;
+  /** Opaque identity shared by events belonging to one application operation. */
+  readonly operationId?: string;
```
cjol: I'd like to avoid introducing a new "operation" entity; not clear what it really corresponds to and widens the taxonomy. Can we use responseId or eventId? Can we avoid the two-paths problem by making whatever it is mandatory?

packages/agents/src/channels/host/index.ts
```diff
@@ -100,13 +155,43 @@ export class ChannelHost {
...
+        resolveMessages: options.resolveMessages,
+        responseStreams: options.streams
+      });
+      const bindIngress = (
+        channel as Channel & Partial<BindableChannelIngress>
+      )[bindChannelIngress];
+      bindIngress?.call(channel, (envelope) =>
+        this.#dispatch(channelKey, channel, envelope)
+      );
+    }
```
cjol: This smells of patching our abstraction with hidden mechanisms instead of evolving the Channel shape. Let's discuss why it's needed and what first-class APIs might look like.

.changeset/durable-channel-response-streams.md
```diff
@@ -0,0 +1,9 @@
+---
```
cjol: we don't need all these changesets, they can be merged.

packages/agents/README.md
```diff
@@ -135,11 +135,191 @@ import { ChannelHost } from "agents/channels";
 import { email } from "agents/channels/email";
 import { slack } from "agents/channels/slack";
 import { telegram } from "agents/channels/telegram";
+import { web } from "agents/channels/web";
 ```
 
 See the [Voice](../../docs/agents/voice.md) and
 [Channels](../../docs/agents/channels.md) references.
 
+Slack and Telegram can include tool status and provider-exposed reasoning in a
```
cjol: Revert this and any other docs, I don't want any AI-authored docs right now forchannels.

packages/agents/src/channels/live-tests/README.md
```diff
@@ -1,9 +1,21 @@
 # Live delivery tests
 
-These local-only tests call the real `telegram()`, `slack()`, and `email()`
-adapters through `ChannelHost`, then read each destination through an independent
-provider API. They are not part of normal package tests, Nx affected tests, or
-CI.
+These local-only tests call the real `telegram()`, `slack()`, `email()`, and
```
cjol: Your changes to this README put way too much emphasis on Web specifically. It's supposed to be one amongst peers.

PR #2284, opened 2026-09-16
fix(playground): make the recurring schedule demo actually recur

### LINKED ISSUE #2283: [playground] the recurring schedule demos fire once using `this.schedule()` instead of `this.scheduleEvery()`
Hi Sunil @threepointone et al 👋 Grateful to see that the agents architecture and docs are just as user-friendly as the partykit ones are!

I found a bug in the playground example, and here's some LLM-generated text matching the style y'all seem to be speaking in other github issues :)

---

## Summary

On `/core/schedule` in `examples/playground`, "Schedule Recurring" fires exactly once, though the UI presents it as an interval that repeats until cancelled.

[`scheduleRecurring`](https://github.com/cloudflare/agents/blob/a7b29135acf127cdf4a44114652970d0a22300de/examples/playground/src/demos/core/schedule-agent.ts#L21-L31) passes a number to `this.schedule()`, which is a one-shot delay, and puts `recurring: true` in the payload — where the scheduler never sees it. It's payload data handed to the callback, not a scheduling option.

```ts
const schedule = await this.schedule(intervalSeconds, "onRecurringTask", {
  label,
  recurring: true
});
```

The list rendering has the matching symptom: [`ScheduleDemo.tsx#L336`](https://github.com/cloudflare/agents/blob/a7b29135acf127cdf4a44114652970d0a22300de/examples/playground/src/demos/core/ScheduleDemo.tsx#L336) labels a row `Every ${intervalSeconds}s` when `schedule.type === "interval"`, which never matches, so a recurring row renders a one-shot fire time instead.

## Repro

1. `/core/schedule`, set the interval to 10s, click **Schedule Recurring**
2. One `recurring_executed` entry arrives in the log; no further ticks
3. The row in **Active Schedules** shows a fire time rather than `Every 10s`, and disappears after it fires

## Fix

`scheduleEvery()` is the interval API. Branch, diff and patch:

- https://github.com/cloudflare/agents/compare/main...patcon:agents:patcon/fix-recurring-schedule-demo
- https://github.com/cloudflare/agents/compare/main...patcon:agents:patcon/fix-recurring-schedule-demo.diff
- https://github.com/cloudflare/agents/compare/main...patcon:agents:patcon/fix-recurring-schedule-demo.patch

It swaps the call in the agent and in the demo's "Recurring intervals" code sample, which quotes the same wrong pattern. 7 insertions, 6 deletions; `oxfmt --check` and `tsc --noEmit` pass. Happy to open it as a PR if that's easier than applying the patch.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (~320 words of issue text from ~250 words of human prompts)

mattzcarey:
The playground's "Schedule Recurring" demo called `this.schedule(intervalSeconds, ...)` with `recurring: true` in the payload. `schedule()` with a number is a one-shot delay and the scheduler never reads the payload, so the task fired once and the row rendered as a one-shot fire time.

This switches the agent and the demo's quoted code sample to `scheduleEvery()`, which is the interval API. The list rendering already handles `type === "interval"`, so the row now shows `Every Ns` and persists until cancelled.

Patch by @patcon from #2283, applied with authorship preserved since non-collaborators cannot open PRs here.

Example-only change, no changeset.

Closes #2283

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/agents/pull/2284" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


patcon:
Thanks! Really appreciate the care around author preservation, and for thinking about how open source works in this strange new world :)

PR #2289, opened 2026-09-16
fix(think): preserve scheduled retries across restarts

### LINKED ISSUE #2276: Think recovery skips an interrupted opened stream with no assistant content
### Summary

Think 0.17.0 has two recovery windows that can terminalize a recoverable chat submission after a Durable Object restart.

An opened durable stream with zero assistant content carries a `streamId`. `_recoverablePreStreamUserId` rejects that state before checking whether the durable leaf is still the originating user message. Recovery settles the submission as `skipped` and never retries the prompt.

A retry callback can also be durably scheduled before the current chat fiber disappears. On the next startup, `_hasScheduledRecoveredContinuation` recognizes `_chatRecoveryContinue` alone. A pending `_chatRecoveryRetry` therefore fails the running submission as interrupted before the retry executes.

### Reproduction

Environment:

- Node 24.18.0
- `@cloudflare/think` 0.17.0
- `agents` 0.22.0
- Miniflare with a SQLite Durable Object

Empty-stream case:

1. Submit one user message through `Think.submitMessages`.
2. Return a readable model stream that opens and emits zero chunks.
3. Restart the Durable Object.
4. Let the same model complete on its next call.

Actual result: the submission settles as `skipped` with `Recovery skipped.` and `onChatResponse` does not fire.

Expected result: Think retries the same durable user leaf, completes once, and fires one response callback.

Scheduled-retry case:

1. Leave a submission running before the model stream opens.
2. Schedule `_chatRecoveryRetry` for that submission and remove the interrupted fiber, reproducing the durable state between recovery classification and callback execution.
3. Restart the Durable Object.

Actual result: startup changes the submission to `error` with `Submission was interrupted after messages were applied.`

Expected result: the submission remains running while the retry callback is pending.

### Proposed fix

- Let `_recoverablePreStreamUserId` accept a `streamId` when durable assistant content is empty, then retain its existing latest-user-leaf validation.
- Let `_hasScheduledRecoveredContinuation` recognize `_chatRecoveryRetry` alongside `_chatRecoveryContinue`.

A Miniflare regression covering baseline completion, empty-stream retry, partial continuation, pending retry preservation, and cancellation passes with those two changes.

ben-reitz:
## Problem

After a restart, Think could mark a submission as failed even though its recovery was still in progress — or had already succeeded. Review found the reported bug plus adjacent gaps, all breaking the same invariant: a running submission must always be resolvable to live or terminal evidence at startup.

1. Startup checked for scheduled continuations but overlooked scheduled retries (the reported bug).
2. Recovery handed off to its replacement turn before that turn was durably recorded, and the replacement ran under a new request ID that startup couldn't match.
3. A finished turn's recovery task could disappear before the submission ledger was updated — and any newly starting turn reclaimed the terminal stream that served as completion evidence.
4. Facet-hosted recoveries never signaled handoff, holding the recovery step for the full model turn against its 15-minute timeout.
5. The recovery lookup matched submission and request identities without precedence, so a collision could touch the wrong submission.

## Fix

Recognize both kinds of recovery work at startup; keep the submission bound to whichever side currently owns it across the handoff (the handoff signal is scoped to the intended successor's call chain, on both the Tasks and facet-fiber paths); and settle the ledger from the turn's retained terminal stream without rerunning the turn or repeating callbacks.

`ResumableStream` gains a small additive lifecycle for that evidence: rows can be retained at settlement, are skipped by reclamation, and become reclaimable again on `release()`. Think pins a running submission's terminal stream until its ledger settles and releases pins at every settlement path and on startup. Exact submission IDs take precedence over request-ID fallback, and new recovery payloads carry the stable submission identity (old payloads still match).

```diff
 on restart
- only a scheduled continuation keeps the submission running
+ scheduled retries, the accepted replacement turn, and a finished
+ turn's retained terminal stream all keep or settle the submission correctly
```

## Scope note

This fixes more than the original startup-detection gap: closing only item 1 still failed deterministic interruption tests for the reported symptom (see the reporter's third scenario in #2276, whose request-ID analysis matches item 2 exactly). The commits land on clean boundaries — `f620ff54` (item 1), `2984fba5` + `dfa76d65` (item 2), `91ee77ba` (items 3–5 + the `agents` retention API, changeset: minor) — and can be split on request.

**Validation:** Deterministic red-first tests for every window above (including foreign-turn interference, foreign-stream reclamation with a real process restart, facet handoff, and ID collisions), full Think (936) and agents (3,244) suites, submission-recovery e2e, build, and `pnpm run check` — all passing with retries disabled.

Fixes #2276


ben-reitz:
Superseded by #2297 (identical content on a fresh branch) to work around a stuck automated review run on this PR.

Issue #2264, opened 2026-09-12
Think: `TurnResult.message` is the session leaf, not the message this turn produced


MatthewHager:
## Summary

`TurnResult.message` is documented as part of the wait-mode result, but `_enrichTurnResult` fills it from `session.getLatestLeaf()` after the turn queue has released. That is the newest message in the conversation, not the assistant message this turn produced.

Two consequences:

1. A queued follow-up can already have appended its user message, so a completed wait turn resolves with `message: undefined` (see #2243).
2. After `submitMessages` + `inspectSubmission`, there is no message at all: `ThinkSubmissionInspection` has `submissionId`, `status`, `error`, timestamps — no `message`. A workflow that admitted the turn by submission id cannot tell which assistant message that turn wrote.

## Environment

- `@cloudflare/think@0.17.0`
- `runTurn({ mode: "wait" })` and `submitMessages` / `inspectSubmission`

## What the types allow

```ts
type TurnResult = SaveMessagesResult & {
  message?: SessionMessage;
  continuation: boolean;
};

type ThinkSubmissionInspection = {
  submissionId: string;
  idempotencyKey?: string;
  requestId?: string;
  status: ThinkSubmissionStatus;
  error?: string;
  metadata?: Record<string, unknown>;
  createdAt: number;
  startedAt?: number;
  completedAt?: number;
};
```

## Reproduction (wait)

Two `runTurn({ mode: "wait" })` calls on the same DO, overlapping:

```ts
await Promise.all([
  stub.runTurn({ mode: "wait", input: "q1" }),
  stub.runTurn({ mode: "wait", input: "q2" }),
]);
```

The first result can be `{ status: "completed", message: undefined }` because the leaf is already the second turn's user message. #2243 has the `_enrichTurnResult` / `TurnQueue` trace.

## Reproduction (submit)

```ts
const { submissionId } = await agent.submitMessages([userMessage], {
  idempotencyKey: "import:instance:capture:homes",
});
const inspection = await agent.inspectSubmission(submissionId);
// inspection.status === "completed"
// inspection.message is not a field
```

The caller has to walk `getMessages()` and guess which assistant row belongs to this submission.

## Workaround

Record the user message id at admit time and take the first assistant after it in `getMessages()`. That is stable for both wait and submit. For wait, do not trust `result.message`.

## Suggested fix

[...]

PR #2290, opened 2026-09-16
fix(think): return each wait turn's own answer

### LINKED ISSUE #2243: think: runTurn({ mode: "wait" }) resolves with message: undefined when a second turn is queued on the same agent
## Summary

`runTurn({ mode: "wait" })` reports the finished turn's assistant message by reading the session's newest message *after* the turn queue has released its slot. When a second turn is already queued on the same agent, that queued turn can append its user message first, so the completed turn resolves with `status: "completed"` and `message: undefined`.

`@cloudflare/think@0.17.0`, `dist/think.js` 3758–3768:

```js
async _enrichTurnResult(result, continuation) {
  let message;
  if (result.status === "completed") {
    const leaf = await this.session.getLatestLeaf();
    if (leaf?.role === "assistant") message = leaf;
  }
  return { ...result, continuation, message };
}
```

`_runTurnWait` awaits `_runProgrammaticMessagesTurn` (which holds the `TurnQueue` slot) and only then calls `_enrichTurnResult`, so the lookup happens outside the serialized region:

```js
const result = await this._runProgrammaticMessagesTurn(crypto.randomUUID(), input, { ... });
return this._enrichTurnResult(result, false);
```

`TurnQueue.enqueue` (`agents/dist/chat/index.js` 164–189) releases in a `finally` before the caller's continuation runs, so the next turn's `execute` — which begins by appending its user message — is free to interleave.

The turn itself is fine: both turns run, both messages persist, and the session is correct afterwards. Only the reported `message` is wrong, which makes a correct turn look like a failure to the caller.

## Environment

- `@cloudflare/think@0.17.0`, `agents@0.22.0` (current npm latest)
- Cloudflare Workers / workerd, `Think` subclass in a Durable Object
- Same code on `main`: `packages/think/src/think.ts` `_enrichTurnResult`

## Reproduction

Two RPC calls on one agent instance, each calling `runTurn({ mode: "wait" })` with a function input:

```ts
await Promise.all([
  stub.appendTurn({ q: "any 2 bedrooms?" }),
  stub.appendTurn({ q: "book a tour" }),
]);
```

Probe inside the first call's continuation:

```
{"status":"completed","leafRole":"user"}
```

Expected `leafRole: "assistant"` for the turn that just completed; the leaf is the *other* turn's question. Reproduces reliably on a cold install and intermittently otherwise — it is a timing race, so it surfaces as a flaky test or an occasional 5xx in production rather than a hard failure.

## Suggested fix

Resolve the assistant message from the run that produced it rather than from session-global state — either capture the leaf inside the admitted turn body (still holding the queue slot) and pass it out with the result, or look it up by the turn's `requestId`. Reading `getLatestLeaf()` after the slot is released is only correct when exactly one turn ever exists.

## Workaround

Callers can record the id of the user message they submit and take the first assistant message after it in `getMessages()`, which is stable regardless of what else is queued.

ben-reitz:
## Problem

Two queued `runTurn({ mode: "wait" })` calls can finish successfully, but one caller receives the other turn's answer—or no answer. We looked up the session's latest message after releasing the queue, when another turn or response hook could have changed it.

## Fix

Track which assistant message each wait turn saved, in SDK-owned per-call context, and read that exact stored message when building the result.

```diff
 build the completed turn's result
- read whichever message is newest
+ read the assistant message this turn saved
```

Review caught that an earlier version passed this through an overridable method's return value, which a legacy subclass override could silently drop; the capture now stays internal, and the protected contracts are unchanged.

**Validation:** Concurrent-turn, continuation, legacy-override, response-hook, and no-output regressions; full Think Workers/React suites, build, and `pnpm run check` passed.

Fixes #2243


threepointone:
I checked this against current `main`. The bug is still there: `_enrichTurnResult` reads `session.getLatestLeaf()` after `_runProgrammaticMessagesTurn` returns, and by then the queue has released the slot. When a second wait turn is queued right behind, it can append its user message first, so the first call returns `completed` with no `message`. #2341 fixed only the submission side of this (`ThinkSubmissionInspection.messageId`), and its description says the wait-mode fix is left to this PR.

The approach looks right. It uses the same `AsyncLocalStorage` pattern as `continuationOutputContext`, and the tests cover queued turns, a queued continuation, aborts and the legacy override. The PR now conflicts with `main`, and it needs these changes:

1. **Rebase.** `_runTurnWait` gained a second continuation branch for an overridden `continueLastTurn`, plus `captureOutput: true`, so `waitTurnResultContext.run(...)` needs to wrap the new body. `_persistAssistantMessageWithCutover` now takes a fifth `submission` argument. The PR's `_discardStreamAtCutover` is now `_streamCutoverOptions`. `_streamResultToRpcCallback` also calls the persist function now, and it should not set the capture.
2. **Write the capture once, or key it by `requestId`.** As written, any later persist that inherits the context overwrites `capture.messageId`. For example, a fire-and-forget `saveMessages` started from `onChatResponse` would replace this turn's answer.
3. **Changeset:** mention that a continuation that persists nothing now returns no `message` instead of the previous assistant leaf.
4. **Link #2264.** Its other half (the submission side) is fixed by #2341, so this PR closes #2264 as well as #2243.

PR #2292, opened 2026-09-16
feat(agents): add the named browser session core

ben-reitz:
## What this is

A session engine. The host names a browser (e.g. `"checkout"`, `"research"`) and one call does the right thing:

```text
resolve("checkout")
  browser still alive?    → reattach to it        (restarted: false)
  dead / closed / swept   → create a replacement  (restarted: true —
                                                   "heads up: fresh browser,
                                                    your login may be gone")
```

Internal module (`browser/session-core.ts`), plus two small backward-compatible additions to already-public helpers - this is the only public behaviour change in this stack.

## Why

A Browser Rendering session is a real Chrome browser: slow to start, and it forgets logins, cookies, and tabs when it dies. Today, keeping one alive across model turns means the *model* does the bookkeeping — deciding survival with `cdp.startSession()` and carrying raw protocol ids between turns — and evaluation showed it keeps fumbling that (stale-id failures, attach ceremony). This engine moves all of it host-side:

- **Names are the only identity that crosses a boundary.** Models never see platform session ids, so they cannot lose them.
- **Death is loud, not cryptic.** `restarted: true` replaces `CDP error -32001`; first-ever use is not a restart.
- **Settings stick to the name.** `keepAliveMs`, `recording`, and `guardrails` are stored durably and reapplied to every replacement browser. Keep-alive defaults to the 600-second platform max.
- **Cleanup is bounded.** `close()` tombstones then deletes (best-effort — the tombstone is the durable outcome, and the pinned `keep_alive` reclaims any browser whose delete failed); `sweep()` closes idle browsers and prunes aged tombstones (10-minute default; alarm scheduling lands in the next PR).
- **Active browsers are never reaped.** CDP commands on a socket from `connect()` refresh the session's idle clock (throttled to 60s, capped at half the sweep window so a short custom `sweepIdleMs` cannot outrun the touch cadence), so a sweep only closes browsers that are genuinely idle.
- **Restart evidence survives recovery.** Dead sessions are tombstoned — never deleted — while a replacement is created, so a resolver racing the recovery also reports `restarted: true` instead of mistaking it for first use.
- **The trivial case stays trivial.** `openOneShotBrowserSession()` is store-less create-use-delete. Kitesurf will use this option.

Storage discipline mirrors the existing connector: locks wrap storage reads/writes only, never Browser Run network calls; concurrent create races resolve first-commit-wins with best-effort cleanup of the loser. The tests include a detector that fails if any network call happens while a lock is held.

# Public API impact

Two additions to existing `agents/browser` exports, both backward compatible:

```diff
 createBrowserSession(browser, {
   keepAliveMs, includeTargets, recording,
+  guardrails: { allowedDomains: [...], allowedDomainSets: [...] }
 })

-connectBrowserSession(browser, sessionId, timeoutMs?)
+connectBrowserSession(browser, sessionId, { timeoutMs?, onClose?, onActivity? })
```

The bare-number `timeoutMs` form still works but is marked `@deprecated` (overload-level, so editors flag it) — the options object is the canonical form and internal callers are migrated to it.

- `guardrails` are just an addition to the existing POST body to acquire a Browser Run session (recently added to their REST contract) and are fixed for the session's life, every connection included. The type names are exported (with the changeset) in the docs PR at the top of the stack.
- allow passing in `onClose`, run exactly once when the session reaches a terminal state — explicit close, peer closure, or socket error
- allow passing in `onActivity`, invoked on every CDP command sent — the activity signal the session core uses for idle tracking

The engine uses a new `browser:session:<name>` keyspace, disjoint from the connector's `cdp:exec:` / `cdp:reuse:` entries — the two coexist, and `browser_execute` behavior is untouched.

## Reviewing

- `session-core.ts` reads top to bottom: resolve → connect → close → sweep → one-shot.
- 23 new tests in `browser-session-core.test.ts`: first-create, live reattach, dead/tombstoned recreate, a concurrent dead-recovery race (both resolvers report `restarted: true`, first-commit-wins, redundant browser reclaimed), option reapplication, per-name isolation, lock discipline, activity touches (refresh, throttle, derived interval under a short sweep window, no resurrection of closed sessions), sweep behavior, close with a failed platform delete, one-shot cleanup including the attach-failure and peer-closure paths plus the dispose-once guard, Kitesurf rejection.


ben-reitz:
Superseded by #2302 (identical commit, recreated so the review bot can pick it up). The stacked PRs above it (#2293, #2294) have been retargeted onto the new branch.

PR #2293, opened 2026-09-16
feat(agents): add the browser-sessions Lifecycle capability

ben-reitz:
## What this is

The host wiring for the session engine one PR down: `BrowserSessions`, a Lifecycle capability any Agent subclass or plain Durable Object composes in one line — and that cleans up after itself. Internal: not exported for now.

```ts
readonly browser = new BrowserSessions({ browser: env.BROWSER });
readonly lifecycle = Lifecycle.install(this).use(this.browser);

await this.browser.resolve("checkout");        // reattach-or-create
await this.browser.sessions();                 // all names, live/closed + timestamps
await this.browser.liveView("checkout", { mode: "tab" }); // fresh ~5-minute human link
```

## The full API surface

Construction — five options, only `browser` required:

```ts
new BrowserSessions({
  browser: env.BROWSER, // required — the Browser Rendering binding
  store,                // custom BrowserSessionStore; defaults to a store over the object's own DO storage. A store without `list` disables the auto-sweep and sessions()
  create: {             // durable — reapplied on every create, incl. restarts
    keepAliveMs,        //   default: the platform max (600s)
    recording,          //   opt into rrweb session recording
    guardrails          //   allowedDomains / allowedDomainSets, fixed at launch
  },
  timeoutMs,            // default CDP command timeout for connect()
  sweepIdleMs           // idle window + sweep cadence, default 10 min
});
```

Six host methods:

| Method | Does |
|---|---|
| `resolve(name?)` | reattach-or-create → `{ name, sessionId, restarted, createdAt, updatedAt }` |
| `connect(name?)` | resolve + attach a CDP socket → `{ name, sessionId, restarted, cdp }` |
| `close(name?)` | tombstone the named session (the next resolve reports `restarted: true`) |
| `sweep()` | run one idle sweep now, besides the scheduled cadence |
| `sessions()` | all named sessions, live/closed + timestamps |
| `liveView(name?, { mode })` | fresh human links, or `undefined` when unknown, closed, or dead |

Every `name` defaults to `"default"`. There is also capability plumbing (`onStart`, `onJob`, `capabilityId`, `claims`) that Lifecycle itself calls — technically visible, not for hosts.

## Why

The engine below this still leaves a host three chores: supply a store, remember to schedule sweeping, and build its own observability. This capability removes all three:

- **Store: automatic.** Defaults to a `DurableBrowserSessionStore` over the host object's own storage. Custom stores still plug in; a store without `list` support means manual sweeping — documented, not silent.
- **Sweeping: runs itself.** One singleflight Lifecycle job — armed when a session exists, re-armed on start after a crash, rescheduled while entries remain, gone when the store empties. An idle host carries no recurring alarm. Six hardenings from review: the job id is namespaced (`browser-sessions:sweep`) so it can never collide with a host's own job ids (the jobs table keys ids globally); session ops supersede a concurrently retiring sweep dispatch instead of trusting its soon-to-be-stale row; a sweep that fails terminally (`onJobError`) reschedules one idle window out rather than silently retiring the recurring job; session ops keep the earliest pending deadline instead of pushing it out, so one busy name cannot starve idle names and tombstones of cleanup; startup re-derives a lost physical alarm when the sweep row survived a crash (same discipline as Scheduler and Tasks); and minting a Live View link counts as session activity, gated on the refresh actually landing — the sweep cannot kill a fresh link inside its ~5-minute connection window, and a mint that loses to a concurrent sweep reports the session gone instead of returning dead links.
- **Observability: host-only.** `sessions()` lists every named browser; `liveView(name, { mode })` mints a fresh Live View link per call. The links are bearer credentials, so they are never persisted and never handed to the model.

## Also in here

- The Live View vocabulary (modes, URL shapes, TTL, minting) moves from `connector.ts` to a shared `browser/live-view.ts`; the connector re-exports the same names.
- The "is this session dead?" judgment (HTTP 404/410) is centralized in `browser-run.ts` instead of duplicated per caller.

## Public API impact

None yet. `BrowserSessions` is unexported, and the connector's re-exports are name-for-name identical to before the extraction.

The intent is to export it and make it public in the near future — once the structure of the API has been proven out in real use and properly evaluated.

## Reviewing

- `capability.ts` first: options → lifecycle hooks (`onStart` crash recovery, `onJob` sweep) → host methods.
- 16 tests in `browser-capability.test.ts`, including real Durable Object alarm firing — timestamps are backdated instead of slept past, so the tests are deterministic — and both wiring styles (Agent subclass and plain Lifecycle Durable Object).


packages/agents/src/browser/capability.ts
```diff
@@ -0,0 +1,317 @@
...
+   * durable, so in every ordinary life the pending job is already there.
+   */
+  override async onStart(): Promise<void> {
+    if (this.lifecycle.jobs.get(BROWSER_SWEEP_JOB_ID)) {
+      // The row is durable but the physical alarm may not be: a prior life
+      // can die between the row write and `setAlarm`. Like Scheduler and
+      // Tasks, recover it explicitly (deferred until startup completes).
+      await this.lifecycle.jobs.rearm();
+      return;
+    }
```
aron-cf, line 122: Is this needed? Worth checking. I don't think a capability should have to manage rearming the jobs list. /cc @mattzcarey

packages/agents/src/browser/capability.ts
```diff
@@ -0,0 +1,317 @@
...
+  #core_(): NamedBrowserSessions {
+    this.#core ??= new NamedBrowserSessions({
+      browser: this.#options.browser,
+      store: this.#sessionStore,
+      create: this.#options.create,
+      timeoutMs: this.#options.timeoutMs,
+      sweepIdleMs: this.#sweepIdleMs
+    });
+    return this.#core;
+  }
```
aron-cf, line 284: This is an odd pattern, why not make `#core` a  getter with lazy initialization of a cached `#_core` field? And maybe call it `#sessions`?

packages/agents/src/browser/capability.ts
```diff
@@ -0,0 +1,317 @@
...
+   * a dispatch that listed an empty store retires the job by returning
+   * nothing — a session op overlapping that window must supersede the
+   * running row, not treat it as sufficient. A same-id push clears the
+   * dispatch marker, so the stale outcome defers to this newer schedule
+   * instead of deleting it.
+   */
+  async #scheduleSweep(): Promise<void> {
+    if (this.#sessionStore.list === undefined) return;
+    await this.#armSweep();
+  }
```
aron-cf, line 298: I think there's probably value in extracting this into a recurring job capability under tasks or the scheduler. So much of this file is here to ensure that cleanup function runs correctly. We already have a scheduler. It feels like the total code in this module should be:

```
onStart() {
  this.lifecycle.scheduler.register("<daily>", this.sweep);
}
```

packages/agents/src/browser/live-view.ts
```diff
@@ -0,0 +1,98 @@
...
+  } catch {
+    return rawUrl;
+  }
+}
+
+/**
+ * Project a fresh target listing into Live View links — one per tab that has
+ * a `devtoolsFrontendUrl`. Pure; the caller supplies just-listed targets.
+ */
+export function mintLiveView(
```
aron-cf, line 80:
```suggestion
export function createLiveView(
```

packages/agents/src/browser/capability.ts
```diff
@@ -0,0 +1,317 @@
...
+   */
+  override async onStart(): Promise<void> {
+    if (this.lifecycle.jobs.get(BROWSER_SWEEP_JOB_ID)) {
+      // The row is durable but the physical alarm may not be: a prior life
+      // can die between the row write and `setAlarm`. Like Scheduler and
+      // Tasks, recover it explicitly (deferred until startup completes).
+      await this.lifecycle.jobs.rearm();
+      return;
+    }
+    const entries = await this.#sessionStore.list?.(namedBrowserSessionKey(""));
```
aron-cf, line 123: I only see `#store` in the class properties, is this a typo or am I missing it...

```suggestion
    const entries = await this.#store.list?.(namedBrowserSessionKey(""));
```

aron-cf (commented):
Took a look, looking good, some inline feedback, some for the API some for lifecycle/capabilities in general.

packages/agents/src/browser/capability.ts
ben-reitz: It's referencing the lazy `sessionStore` getter down near the bottom of the file - we do lazy initialisation of the store until it's actually used.

packages/agents/src/browser/capability.ts
ben-reitz: Yep this would be better aligned with the other lazily getter that's literally above it...

Will probably call it `get #namedSessions()` rather than `get #sessions()` just to keep it distinct from the public `sessions()` method (and match the name of the `NamedBrowserSessions` class it returns).

packages/agents/src/browser/session-core.ts
```diff
@@ -206,50 +183,39 @@ export class NamedBrowserSessions {
...
-        if (outcome.winner) {
-          if (outcome.winner.closedAt !== undefined) continue; // re-tombstoned
-          return { name, restarted, ...outcome.winner };
-        }
-        return { name, restarted, ...outcome.stored };
+      if (existing === undefined) {
+        // No browser on record: create one. The commit reports whether the
+        // name was ever used before (see #createAndCommit).
+        const { session, restarted } = await this.#createAndCommit(name);
+        return { name, restarted, ...session };
```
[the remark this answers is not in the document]
ben-reitz: Not valid - tombstones never shipped so there won't be any legacy ones

packages/agents/src/browser/capability.ts
ben-reitz: cleanup job removed so this is no longer needed

packages/agents/src/browser/capability.ts
ben-reitz: We should follow up on this as the general point here still stands - lifecycle should own the rearm on startup rather than requiring capabilities to do this. I looked into it a bit and it seems like there are ways that an alarm can go missing and solving that should live in the lifecycle startup.

**It's not relevant for this PR though anymore** as i've overhauled the capability so that it's no longer got any hooks or schedules.

ben-reitz:
I've removed the sweep entirely (#2358, now merged into this PR).

@aron-cf's comments on rearming and on moving cleanup into the Scheduler made me think more about whether the sweep needed to exist at all. Browser Run's `keep_alive` already deletes a browser after 10 minutes without activity, and `resolve()` already detects a missing browser and replaces it with `restarted: true`. The sweep was kind of duplicating the platform's job a second time, and most of this file existed only to keep it running reliably.

PR #2302, opened 2026-09-18
feat(agents): add the named browser session core

ben-reitz:
## What this is

A session engine. This allows us to create Browser Run session in the host code by giving it a constant name (e.g. `"checkout"`, `"research"`) and then one call to `resolve` will do one of the following depending on the existing state of that session:
- creates a new Browser Run session with that name if one did not previously exist, or
- attaches to the existing Browser Run session by that name, or
- creates a replacement Browser Run session by that name if the previous one died or was closed

```text
resolve("checkout")
  browser still alive?    → reattach to it        (restarted: false)
  dead / closed / swept   → create a replacement  (restarted: true —
                                                   "heads up: fresh browser,
                                                    your login may be gone")
```

## Why

A Browser Run session is slow to start, and it forgets logins, cookies, and tabs when it dies. Today, keeping one browser session alive across model turns is something the model has to decide and manage. It decides when it wants survival with `cdp.startSession()` and carries raw protocol ids between turns. Evaluations showed it kept fumbling that (stale-id failures, attach ceremony). The changes in this PR set the foundation for moving that session management host-side:

#### Before
This is today, with `dynamic` mode enabled. The model owns both the **Browser Run** platform session AND the **CDP** session:

 ```js
   // ── browser_execute #1: log in, keep the browser ──────────────
   // target ceremony (every execution)
   const { targetInfos } = await cdp.send("Target.getTargets");
   const page = targetInfos.find((t) => t.type === "page");
   const { sessionId } = await cdp.attachToTarget({ targetId: page.targetId });

   await cdp.send("Page.navigate", { url: "https://shop.example/login" }, { sessionId });
   // ...fill form, submit, all with { sessionId }...

   // if the model forgets to do this next bit, then the logged-in browser is
   // deleted when this execution ends
   await cdp.startSession();
 ```

 ```js
   // ── browser_execute #2: later turn ─────────────────────────────
   await cdp.sessionInfo();                       // is my browser still there?
   const { targetInfos } = await cdp.send("Target.getTargets");   // re-discover
   const page = targetInfos.find((t) => t.type === "page");
   const { sessionId } = await cdp.attachToTarget({ targetId: page.targetId }); // re-attach
   await cdp.send("Page.navigate", { url: "https://shop.example/checkout" }, { sessionId });

   // ...and eventually, if it remembers:
   await cdp.closeSession();
 ```

#### After
Target state: both Browser Run and CDP lifecycle management gone from the model's code:

 ```js
   // ── browser_execute #1: log in ─────────────────────────────────
   await cdp.send("Page.navigate", { url: "https://shop.example/login" }, { sessionId: "active" });
   // ...fill form, submit, all with { sessionId: "active" }...
   // nothing else — persistence isn't the model's decision
 ```

 ```js
   // ── browser_execute #2: later turn, same browser, still logged in ──
   await cdp.send("Page.navigate", { url: "https://shop.example/checkout" }, { sessionId: "active" });
   // if the browser had died meanwhile, the result envelope says "sessionRestarted: true"
   //  and the model knows to log in again
 ```

### Decisions
- **Models never see platform session ids**, so they cannot lose them.
- **We tell the model if it's getting a new browser for a pre-existing session:** `restarted: true` message replaces `CDP error -32001`.
- **Consistent configuration:** `keepAliveMs`, `recording`, and `guardrails` are stored durably and reapplied to every replacement browser for a session. Keep-alive defaults to the 10 minute platform max.
- **Cleanup is durable:** `close()` tombstones the session then deletes (best-effort attempt at durability). `sweep()` closes idle browsers and prunes aged tombstones (10-minute default - alarm scheduling is in the next PR).
- **Active browsers are never deleted:** Sending CDP commands refreshes the session's idle clock, so a sweep only closes browsers that are actually idle.
- **Restart evidence is durable:** While a replacement session is created, dead sessions are tombstoned instead of deleted, so a resolver racing the recovery also reports `restarted: true` instead of mistaking it for first use.
- **One-shot is still an option** `openOneShotBrowserSession()` doesn't use the store. it's create-use-delete. Kitesurf will always use this option.

# Public API impact

Two additions to existing `agents/browser` exports, both backward compatible:

```diff
 createBrowserSession(browser, {
   keepAliveMs, includeTargets, recording,
+  guardrails: { allowedDomains: [...], allowedDomainSets: [...] }
 })

-connectBrowserSession(browser, sessionId, timeoutMs?)
+connectBrowserSession(browser, sessionId, { timeoutMs?, onClose?, onActivity? })
```

`timeoutMs` as a 3rd arg still works but is marked `@deprecated`. The new options object is the correct method going forward and internal callers have been migrated to it.

- `guardrails` are just an addition to the existing POST body to acquire a Browser Run session (recently added to their REST contract) and are fixed for the session's life, every connection included. The type names are exported (with the changeset) in the docs PR at the top of the stack.
- allow passing in `onClose`, run exactly once when the session reaches a terminal state — explicit close, peer closure, or socket error
- allow passing in `onActivity`, invoked on every CDP command sent — the activity signal the session core uses for idle tracking

The engine uses a new `browser:session:<name>` keyspace in storage, separate from the connector's `cdp:exec:` / `cdp:reuse:` entries. The two coexist for now, and `browser_execute` behaviour is untouched.

## Reviewing

- `session-core.ts` reads top to bottom: resolve → connect → close → sweep → one-shot.
- 23 new tests in `browser-session-core.test.ts`: first-create, live reattach, dead/tombstoned recreate, a concurrent dead-recovery race (both resolvers report `restarted: true`, first-commit-wins, redundant browser reclaimed), option reapplication, per-name isolation, lock discipline, activity touches (refresh, throttle, derived interval under a short sweep window, no resurrection of closed sessions), sweep behavior, close with a failed platform delete, one-shot cleanup including the attach-failure and peer-closure paths plus the dispose-once guard, Kitesurf rejection.


packages/agents/src/browser/session-core.ts
```diff
@@ -0,0 +1,498 @@
...
+  ): Promise<
+    | { stored: StoredBrowserSession; winner?: undefined }
+    | { stored?: undefined; winner: StoredBrowserSession }
+  > {
+    const info = await createBrowserSession(this.#browser, {
+      keepAliveMs:
+        this.#create.keepAliveMs ?? BROWSER_SESSION_KEEP_ALIVE_MAX_MS,
+      recording: this.#create.recording,
+      guardrails: this.#create.guardrails
+    });
```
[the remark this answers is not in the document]
ben-reitz: Pre-existing issue and extremely remote chance of this happening

packages/agents/src/browser/session-core.ts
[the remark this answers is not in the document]
ben-reitz: Lands in another PR further up the other stack

packages/agents/src/browser/browser-run.ts
```diff
@@ -128,11 +145,21 @@ export async function createBrowserSession(
     keepAliveMs?: number;
     includeTargets?: boolean;
     recording?: boolean;
+    guardrails?: BrowserSessionGuardrails;
```
aron-cf, line 166: If nothing is passed, is everything allowed or nothing?

packages/agents/src/browser/browser-run.ts
```diff
@@ -128,11 +145,21 @@ export async function createBrowserSession(
...
 ): Promise<BrowserSessionInfo> {
+  // Guardrails ride the acquire request's JSON body (the query string only
+  // carries scalar options).
+  const init: RequestInit = options?.guardrails
+    ? {
+        method: "POST",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify({ guardrails: options.guardrails })
+      }
+    : { method: "POST" };
```
aron-cf: Does this ternary exist because the API errors if passed an empty body?
```suggestion
  const { guardrails } = options; 
  const init: RequestInit = { method: "POST", headers: { "Content-Type": "application/json" }, body: JSON.stringify(guardrails ? { guardrails } : {}) };
```

packages/agents/src/browser/browser-run.ts
```diff
@@ -319,5 +379,11 @@ export async function connectBrowserSession(
   }
 
   ws.accept();
-  return new CdpSession(ws, timeoutMs, undefined, sessionId);
+  return new CdpSession(
```
aron-cf: Probably worth refactoring `CdpSession` to take an object rather than positional arguments.

packages/agents/src/browser/session-core.ts
```diff
@@ -0,0 +1,498 @@
...
+export const SESSION_TOUCH_INTERVAL_MS = 60_000;
+
+/** The session name used when a host doesn't wire one explicitly. */
+export const DEFAULT_BROWSER_SESSION_NAME = "default";
+
+const NAMED_SESSION_KEY_PREFIX = "browser:session:";
+
+/** The store key for a named browser session. */
+export function namedBrowserSessionKey(name: string): string {
+  return `${NAMED_SESSION_KEY_PREFIX}${name}`;
```
aron-cf, line 39: Does `name` need to be validated somewhere?

packages/agents/src/browser/session-core.ts
```diff
@@ -0,0 +1,498 @@
...
+  /** Applied on every create — see {@link BrowserSessionCreateOptions}. */
+  create?: BrowserSessionCreateOptions;
+  /** Default CDP command timeout for {@link NamedBrowserSessions.connect}. */
+  timeoutMs?: number;
+  /**
+   * Idle window before {@link NamedBrowserSessions.sweep} closes a session
+   * (and before it prunes a tombstone). Defaults to
+   * {@link DEFAULT_SWEEP_IDLE_MS} (10 minutes).
+   */
+  sweepIdleMs?: number;
```
aron-cf, line 71: This is weird that this exists here because the session doesn't actually handle the timeout, it's just here as metadata for the sweep method. I think it'd be cleaner to pass it into `sweep(since: number)` in the lifecycle capability.

packages/agents/src/browser/session-core.ts
```diff
@@ -0,0 +1,498 @@
...
+  /**
+   * Minimum interval between activity-driven `updatedAt` refreshes on
+   * connected sockets. Defaults to {@link SESSION_TOUCH_INTERVAL_MS} and is
+   * always capped at half of `sweepIdleMs`, so continuous CDP traffic
+   * refreshes the idle clock before a sweep deadline can pass. Overridable
+   * primarily for tests.
+   */
+  touchIntervalMs?: number;
```
aron-cf, line 92: I'm not sure I understand this one either, it looks like it exists to throttle the touch calls in the `onActivity` handler, but I don't follow how it connects to the sleep.

packages/agents/src/browser/session-core.ts
```diff
@@ -0,0 +1,498 @@
...
+ * session when the returned {@link CdpSession} closes. No store involved —
+ * one-shot sessions have no name and no durability.
+ */
+export async function openOneShotBrowserSession(
+  browser: BrowserBinding,
+  options: OneShotBrowserSessionOptions = {}
+): Promise<CdpSession> {
+  if (options.browser === "kitesurf") {
+    if (options.guardrails) {
+      throw new Error("Kitesurf does not support guardrails");
```
aron-cf: Can we use a discriminating union for the options to guard this at the type layer.

aron-cf (commented):
I think a usage example would help in the PR description. My understanding of this is it means that we can recover an existing browser session across executions?

packages/agents/src/browser/browser-run.ts
ben-reitz: Follows the [default Browser Run behaviour](https://developers.cloudflare.com/browser-run/features/guardrails/) of allowing all if nothing is passed.

packages/agents/src/browser/browser-run.ts
ben-reitz: Yep, it's to keep the pre-existing behaviour of sending no headers or body **in case** sending an empty body breaks something.

I've not verified if it actually does or not - I would assume it's fine but didn't really want to risk it, (or spend time/effort figuring it out to save a few lines of code). Happy to look into it if you feel strongly though!

packages/agents/src/browser/browser-run.ts
ben-reitz: Yep - agreed

packages/agents/src/browser/session-core.ts
ben-reitz: It's set in the host code, and only used in the session store so never reaches anything dangerous. I think the only limit it could break is the DO storage key max size which is massive and would break loudly.

we could iterate on this later if/when we start letting models or users influence names - that's when we might want to validate a more sensible length cap or check chars are printable etc.

ben-reitz:
> I think a usage example would help in the PR description. My understanding of this is it means that we can recover an existing browser session across executions?

Yep that's right. I've updated the PR description to try and make it clearer.

packages/agents/src/browser/browser-run.ts
ben-reitz: I've just verified it and confirmed it's fine with an empty body 👍🏼

packages/agents/src/browser/session-core.ts
ben-reitz: I actually moved both `sweepIdleMs` and `touchIntervalMs` to be a property of the session itself to fix a bug that occurred when the `touchIntervalMs < sweepIdleMs`.

This diagram hopefully explains it better than i can in words, but in essence, we have to keep track of when we last used the CDP connection between the agent and the browser, but we also need to throttle how often we write that message to the store (using `touchIntervalMs`). That throttle means if the sweep happens more often than the touchIntervalMs, the CDP connection could be in active use, but the browser will still get swept up if `touchIntervalMs < sweepIdleMs`  

```mermaid
flowchart LR
    subgraph W["WRITER: every open connection (connect())"]
        A[CDP message<br/>sent or received] --> B{last touch<br/>older than<br/>touchIntervalMs?}
        B -- no --> X[skip]
        B -- yes --> C[touch:<br/>updatedAt = now]
    end
    C --> S[("store<br/>browser:session:checkout<br/>updatedAt")]
    subgraph R["READER: sweep() on the Lifecycle alarm"]
        D{now − updatedAt<br/>≥ sweepIdleMs?} -- yes --> E[close browser<br/>+ tombstone]
        D -- no --> F[leave it]
    end
    S --> D
```

packages/agents/src/browser/session-core.ts
ben-reitz: https://github.com/cloudflare/agents/pull/2302#discussion_r4082365233

